PR #16, opened 2024-01-25
Db-seeding

ivanslabbert:
Create database seeding and view creating functions and documents.


laurensgroeneveld (approved):
LGTM - just fix the test/linting error. Let's add tests for the db service as a next step.

PR #19, opened 2024-01-29
Update the data model

ivanslabbert:


api/src/db/db.service.ts
```diff
@@ -166,3 +177,16 @@ export class DbService {
         });
     }
 }
+
+// eslint-disable-next-line @typescript-eslint/no-namespace
```
laurensgroeneveld, line 181: There's a reason for [this rule](https://typescript-eslint.io/rules/no-namespace/) 😉 

> TypeScript historically allowed a form of code organization called "custom modules" (module Example {}), later renamed to "namespaces" (namespace Example). Namespaces are an outdated way to organize TypeScript code. ES2015 module syntax is now preferred (import/export).

I don't think a namespace like this is needed, it's usually not how (internal) code is organized. Just put the type directly at the top of this file or move it to a separate types file. I think the first option is fine for now - other classes using the DbService don't necessarily need to import the type, as the editor/compiler will still understand which type it is when passing the argument as an object.

api/src/db/db.service.ts
```diff
@@ -102,28 +102,39 @@ export class DbService {
...
-     * @param {Array<string>} accessTags - Array with access tag ID's
-     * @param {Array<string>} subscriptionTags - Array with subscription tag ID's
+     * Get data to which a user has access to including the user document itself.
+     * @param {string} userID - User document ID.
+     * @param {Array<string>} readAcl - Array with read access acl IDs to which the user has access to. This should be an expanded list of acl IDs.
+     * @param {DbService.getDocsOptions} options - Query configuration object.
      * @returns - Promise containing the query result
      */
-    getDocs(accessTags: Array<string>, subscriptionTags: Array<string>) {
+    getDocs(
```
laurensgroeneveld, line 111: Maybe add a test for this? Then we can get a better feel for how to fetch different types of documents.

laurensgroeneveld (commented):
Got two comments on the changed code.

api/src/db/db.service.spec.ts
```diff
@@ -71,14 +76,97 @@ describe("DbService", () => {
...
+    });
+
+    it("can retrieve documents using one group selector", async () => {
+        const res: any = await service.getDocs("user-public", {
+            groups: ["group-public-content"],
+            types: ["post", "tag"],
+            from: 0,
+        });
+
+        expect(res.docs.length).toBe(4);
```
laurensgroeneveld, line 108: I suggest adding assertions for the types of document so we know what's returned.

api/src/db/db.service.spec.ts
```diff
@@ -71,14 +76,97 @@ describe("DbService", () => {
...
+
+    it("can retrieve documents of two types", async () => {
+        const res: any = await service.getDocs("user-public", {
+            groups: ["group-public-content"],
+            types: ["post", "tag"],
+            from: 0,
+        });
+
+        expect(res.docs.length).toBe(4); // result always includes the user's own user document
+    });
```
laurensgroeneveld, line 139: Same test as the one above.

api/src/db/db.service.ts
```diff
@@ -102,28 +115,63 @@ export class DbService {
...
+    /**
+     * Get data to which a user has access to including the user document itself.
+     * @param {string} userID - User document ID.
+     * @param {getDocsOptions} options - Query configuration object.
      * @returns - Promise containing the query result
      */
-    getDocs(accessTags: Array<string>, subscriptionTags: Array<string>) {
+    getDocs(userId: string, options: getDocsOptions): Promise<unknown> {
+        // Set default options
+        if (!options.from) options.from = 0;
```
laurensgroeneveld, line 138: Currently this is mandatory in the type, would probably make sense to make it optional (and add a test for this).

laurensgroeneveld (approved):
Just some comments on the tests, looks good otherwise 👍

PR #23, opened 2024-01-31
Basic socket.io implementation

ivanslabbert:


api/src/db/db.service.spec.ts
```diff
@@ -1,7 +1,7 @@
 import { Test, TestingModule } from "@nestjs/testing";
 import { DbService } from "./db.service";
 import { randomUUID } from "crypto";
-import { upsertDesignDocs, upsertSeedingDocs, destroyAllDocs } from "./db.seedingFunctions";
+// import { upsertDesignDocs, upsertSeedingDocs } from "./db.seedingFunctions";
```
laurensgroeneveld, line 4: This comment can be removed.

api/src/socketio.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+                    updateVersion: 0,
+                });
+            });
+        }
+
+        const res: any = await testSocket();
+        expect(Array.isArray(res)).toBe(true);
+
+        // CMS option is excluded, so the "group" type should not be included in result
+        expect(res.filter((t) => t.type === "group").length).toBe(0);
```
laurensgroeneveld, line 54: We probably want some tests in the future on what data is returned, but for now it's fine.

api/src/socketio.ts
```diff
@@ -0,0 +1,54 @@
...
+        if (reqData.version && typeof reqData.version === "number") from = reqData.version;
+
+        // Test query to return some data to CMS
+        this.db
+            .getDocs(socket.data.user, {
+                groups: socket.data.groups,
+                types: docTypes,
+                from: from,
+            })
+            .then((res: any) => {
```
laurensgroeneveld, line 47: We could look into typing our documents from the database, because we know that it'll have at least some fields like `_id`.

laurensgroeneveld (approved):
Nice work 🙌

PR #31, opened 2024-02-09
Accept change documents from client/cms

ivanslabbert:
Implemented Update documents received from CMS/client to API with API returning an Acknowledgement with accept or decline status.

Other issues fixed in this PR:

- Prevent updates to database on unchanged documents to prevent re-indexing (especially when the docs are design docs / views).
- Handle duplicate updating of documents correctly (this can happen when two processes are trying to update the same document simultaneously)
- Remove teardown logic deleting all documents in database from test to avoid complications on re-creating deleted documents in database interface logic testing.
- Set updatedTimeUtc in API upon Update document processing (and removed fields from test data set)


api/src/db/db.service.spec.ts
```diff
@@ -164,6 +164,59 @@ describe("DbService", () => {
...
+                    groupId: "group-private-editors",
+                    permission: ["view", "assign"],
+                },
+            ],
+        });
+
+        expect(res).toBe("passed document equal to existing database document");
+    });
+
+    it("can handle simultaneous updates to a single document", async () => {
```
ivanslabbert, line 205: This test is failing in CI but works in local testing.

Gives the following error in CI:
image

api/jest.teardown.ts
```diff
@@ -1,5 +1,13 @@
...
-    await destroyAllDocs();
+    // await destroyAllDocs();
+    // Note: Destroying documents does not delete them from the database, but marks them as deleted.
+    // This does not work well with testing where we reuse the testing database (i.e. local testing
+    // on your computer) as CouchDB / nano complains that the document is deleted, and hence cannot
+    // be created. In real life scenarios a recreating a deleted document should never occur, and
+    // if it does this would be an exception that should be logged to an error log instead of
+    // being processed.
+    // It therefore does not work to destroy our testing data set for local testing - the developer
+    // can rather delete the testing database on his computer once in a while if it gets too big.
```
laurensgroeneveld, line 12: Perhaps move this note to the documentation and remove this file?

api/src/db/db.service.spec.ts
```diff
@@ -164,6 +164,59 @@ describe("DbService", () => {
         expect(docCount).toBe(8);
     });
 
+    it("can detect that a document is exactly the same as the document in the database", async () => {
+        const res: any = await service.upsertDoc({
+            _id: "group-private-content",
```
laurensgroeneveld, line 169: Suggestion: import the actual document here from the JSON file so we know it's the same as the seed, and don't have to update it when the seed updates.

api/src/db/db.service.spec.ts
```diff
@@ -164,6 +164,59 @@ describe("DbService", () => {
...
+        for (let index = 0; index < 50; index++) {
+            pList.push(service.upsertDoc({ _id: "simultaneousTest", testVal: index }));
+        }
+
+        let res: boolean = false;
+        await Promise.all(pList);
+
+        // if we got past this point without an exception, the test was successful
+        res = true;
+        expect(res).toBe(true);
```
laurensgroeneveld, line 217: Suggestion: get the document here and assert the index is what we expect. Maybe the method silently fails and it didn't actually work.

api/src/db/db.service.ts
```diff
@@ -67,8 +114,21 @@ export class DbService {
      * Get a document by ID
      * @param id - document ID (_id field)
      */
-    getDoc(id: string) {
-        return this.db.get(id);
+    getDoc(docId: string) {
```
laurensgroeneveld, line 117: Since this is a public method, I suggest adding a test for it, especially the cases in the `catch`.

api/src/db/db.service.ts
```diff
@@ -49,16 +50,62 @@ export class DbService {
...
+                                resolve(res);
+                            })
+                            .catch((err) => {
+                                if (err.reason == "Document update conflict.") {
+                                    // This error can happen when a document is updated near-simultaneously by another process, i.e.
+                                    // after the revision has been returned to this process but before this process could write the
+                                    // change to the database. To resolve this, just try again to get the updated revision ID and update
+                                    // the document.
+
+                                    // TODO: We should probably have a retry counter here to prevent the code from retrying endlessly.
```
laurensgroeneveld, line 82: We don't have monitoring in place yet but this sounds like a good thing to monitor in the future. Just to keep it in mind.

api/src/db/db.service.ts
```diff
@@ -49,16 +50,62 @@ export class DbService {
...
+                            .insert(doc)
+                            .then((res) => {
+                                resolve(res);
+                            })
+                            .catch((err) => {
+                                if (err.reason == "Document update conflict.") {
+                                    // This error can happen when a document is updated near-simultaneously by another process, i.e.
+                                    // after the revision has been returned to this process but before this process could write the
+                                    // change to the database. To resolve this, just try again to get the updated revision ID and update
+                                    // the document.
```
laurensgroeneveld, line 80: Is it possible to write a test for this?

laurensgroeneveld (commented):
Just some comments on improving the test coverage.

api/src/db/db.service.spec.ts
laurensgroeneveld: Hmm weird since it seemingly complains about JSON.stringify and also about the socket which is not referenced in the db service 🤔

ivanslabbert: Seems this error is related to first run with an empty database. On second run it passes.

laurensgroeneveld (commented):
Not sure if it was in a reviewable state already but left some comments. The PR is getting a bit big so I would recommend postponing more functionality like validation or authentication to a different PR, to make reviewing and merging easier.

api/src/db/db.service.spec.ts
ivanslabbert: Solved. The issue was that the test which is testing simultaneous document updating ran on an empty database (while expecting the document to exist). Added a line to first create a document before doing the simultaneous update test.

ivanslabbert:
Pushed several other changes to this branch / PR, so will need a re-review.

api/src/db/db.service.ts
ivanslabbert: There is a test for this yes. It was this test that caused the failed test in CI but it is resolved now.

api/src/socketio.ts
```diff
@@ -61,4 +66,47 @@ export class Socketio {
...
+            // TODO: Permission check
+
+            // Process update document
+            this.db
+                // Update in database
+                .upsertDoc(data.doc)
+                // Send acknowledgement to client
+                .then(() => {
+                    const ack: AckDto = {
+                        _id: data._id, // Uuid of submitted UpdateDto to be acknowledged
```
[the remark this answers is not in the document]
ivanslabbert: Agree, will change it

api/src/dto.ts
```diff
@@ -0,0 +1,106 @@
...
+    acl?: Array<GroupAclEntryDto>; // Used for Group docType
+    changes: any; // Object containing changed document fields
+};
+
+/**
+ * Document structure for client / CMS submitted changes to existing documents or new documents.
+ * After successful processing, a differential "change" document (ChangeDto) is generated by the
+ * API from the UpdateDto and distributed to clients.
+ */
+export type UpdateDto = {
```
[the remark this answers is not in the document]
ivanslabbert: Perhaps ChangeReq (for Change Request) will be a better name instead of Update?

api/src/db/db.service.spec.ts
ivanslabbert: I agree that this could be better, but then I have to create a function for reading a single document from file which only would be needed for this test. The current seeding functions only includes a function to read all JSON files as documents from a directory and insert/update them in the database.

api/src/db/db.service.spec.ts
ivanslabbert: Updating documents with the wrong revision ID throws an exception. If the Promise.all() runs without exceptions the logic handles the exceptions correctly by re-trying to get the latest revision ID and updating the document again. Looking at the index value will not work as we do not know which simultaneous update is actually ending up being the final update. (The whole idea is not to have sequential updates but to simulate a real-world scenario where two users update a document at exactly the same time.)

Changed the number of simultaneous updates to 50 which should give a decent amount of real duplicate updates to test the logic (hits the exception handling around 10 times for the 50 document updates in my development environment).

api/src/db/db.service.ts
ivanslabbert: Tests added

api/src/dto.ts
```diff
@@ -0,0 +1,106 @@
...
+    permission: Array<AclPermission>;
+};
+
+/**
+ * Database structured Group object
+ */
+export type GroupDto = {
+    _id: Uuid;
+    _rev?: string;
+    type: DocType; // Should always be "group"
```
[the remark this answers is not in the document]
ivanslabbert: Thanks, fixed it on all DTO's where the type should be a specific value.

api/src/socketio.ts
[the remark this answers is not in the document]
ivanslabbert: This also concerns the UpdateDto (now renamed to ChangeReqDto)

api/src/db/db.service.spec.ts
laurensgroeneveld: Ah ok, and then we rely on our changelog to view the actual changes in the CMS in case someone ends up with an unexpected version?

api/src/db/db.service.spec.ts
```diff
@@ -65,16 +65,27 @@ describe("DbService", () => {
...
+
         const res: number = await service.getLatestDocUpdatedTime();
 
-        expect(res).toBe(3);
+        expect(res).toBeGreaterThan(Date.now() - 100);
     });
 
-    it("can update get the oldest changelogEntry document updated time", async () => {
+    it("can get the oldest changelogEntry document updated time", async () => {
+        // This is difficult to test accurately as the updated time is generated on document insertion into the
```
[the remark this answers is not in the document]
ivanslabbert: It does not seem to be possible to use jest.setSystemTime() outside of test files. We need to do this before the tests are run to set the date before any documents are inserted into the database but could not get it to work.

laurensgroeneveld (approved):
LGTM, just remember to rebase before merging.

PR #33, opened 2024-02-12
Validate Socket.io received "data" messages

ivanslabbert:
Add DTO classes used by class-transformer and class-validator and implement for received Socket.io "data" messages


api/src/permissions/permissions.service.spec.ts
```diff
@@ -84,17 +84,17 @@ describe("PermissionService", () => {
         const res = Group.getAccess(
             ["group-super-admins"],
             [
-                DocType.Audio,
+                // DocType.Audio,
```
ivanslabbert, line 87: Temporarily commented out until DTO's are added for media documents.

api/src/dto/ChangeDto.ts
```diff
@@ -0,0 +1,21 @@
...
+export class ChangeDto {
+    @IsNotEmpty()
+    _id: Uuid;
+    @IsNotEmpty()
+    type: DocType.Change;
+    @IsNotEmpty()
+    docId: Uuid;
+    @IsNotEmpty()
+    docType: DocType;
+    memberOf?: Uuid[];
```
laurensgroeneveld, line 17: How would we handle advanced validation rules? Just thinking that here we specifically want to check if the passed Uuid's are valid group id's and if the user has access to them. Or do you think we should handle that in the API code and just discard any non-valid id's?

api/src/dto/ChangeDto.ts
```diff
@@ -0,0 +1,21 @@
...
+import { GroupAclEntryDto } from "./GroupAclEntryDto";
+
+/**
+ * Database structured Post object
+ */
+export class ChangeDto {
+    @IsNotEmpty()
+    _id: Uuid;
+    @IsNotEmpty()
+    type: DocType.Change;
```
laurensgroeneveld, line 12: Does this automatically check if the type is "change"?

api/src/dto/ChangeReqAckDto.ts
```diff
@@ -0,0 +1,16 @@
...
+/**
+ * API acknowledgement to client after client submission of an ChangeReqDto document.
+ */
+export class ChangeReqAckDto {
+    @IsNotEmpty()
+    docId: Uuid; // Uuid of submitted ChangeReqDto to be acknowledged
+    @IsNotEmpty()
+    type: DocType.ChangeReqAck;
+    @IsNotEmpty()
+    ack: AckStatus;
```
laurensgroeneveld, line 13: Does this validate that it is one of the allowed enum values?

api/src/dto/ContentDto.ts
```diff
@@ -0,0 +1,36 @@
...
+    @IsString()
+    title: string;
+    @IsString()
+    summary: string;
+    @IsString()
+    author: string;
+    @IsString()
+    text: string;
+    @IsString()
+    seo: string;
```
laurensgroeneveld, line 29: Small note, I propose to leave this out for now as I don't think "seo" will actually be a string, more likely a separate class with several fields such as description and keywords. This is what the current site has:

image

I left it out in the frontend for now for that reason as well.

laurensgroeneveld (commented):
I just have some small questions on the validation library and how we handle more advanced cases.

In general I think we need some tests for this validation, or do you plan on adding those in the follow-up PR with validation messages and errors to the client? Which would be fine with me.

api/src/dto/ChangeReqAckDto.ts
ivanslabbert: Added an @Equals decorator to test for a specific value

api/src/dto/ChangeDto.ts
ivanslabbert: Adding a Equals decorator to ensure the correct value

api/src/dto/ChangeDto.ts
ivanslabbert: Can add multiple validation decorations. Access validation will be done separately in the API by making calls to the permission system

api/src/dto/ContentDto.ts
ivanslabbert: Commented out for now

api/src/dto/ChangeDto.ts
```diff
@@ -1,21 +1,39 @@
...
     docType: DocType;
+
+    @IsOptional()
+    @IsArray()
+    @IsString({ each: true })
     memberOf?: Uuid[];
+
+    @IsOptional()
+    @ValidateNested({ each: true })
+    @Type(() => GroupAclEntryDto) // This throws an exception on validation failure, so we need to catch the error on validation. The message is less user-friendly but at least the validator fails and will protect our data.
```
laurensgroeneveld, line 34: It's possible to customize the message which would make sense if it's not super clear (and especially if we want to display it in the frontend): https://github.com/typestack/class-validator?tab=readme-ov-file#validation-messages

api/src/dto/ChangeReqAckDto.ts
```diff
@@ -1,16 +1,23 @@
...
  * API acknowledgement to client after client submission of an ChangeReqDto document.
  */
 export class ChangeReqAckDto {
     @IsNotEmpty()
-    docId: Uuid; // Uuid of submitted ChangeReqDto to be acknowledged
+    @IsString()
+    reqId: Uuid; // Uuid of submitted ChangeReqDto to be acknowledged
+
     @IsNotEmpty()
+    @Equals("changeReqAck")
```
laurensgroeneveld, line 13: Did you try putting `DocType.ChangeReqAck` here? Then we don't have to use the string value directly.

api/src/dto/ChangeReqAckDto.ts
```diff
@@ -1,16 +1,23 @@
...
+import { Equals, IsEnum, IsNotEmpty, IsOptional, IsString } from "class-validator";
 
 /**
  * API acknowledgement to client after client submission of an ChangeReqDto document.
  */
 export class ChangeReqAckDto {
     @IsNotEmpty()
-    docId: Uuid; // Uuid of submitted ChangeReqDto to be acknowledged
+    @IsString()
+    reqId: Uuid; // Uuid of submitted ChangeReqDto to be acknowledged
```
laurensgroeneveld, line 10: There is also a `@IsUUID` assertion which I think fits more for these cases.

api/src/dto/ContentDto.ts
```diff
@@ -1,36 +1,60 @@
...
     @IsString()
+    @IsNotEmpty()
     title: string;
+
     @IsString()
+    @IsNotEmpty()
     summary: string;
+
     @IsString()
+    @IsOptional()
```
laurensgroeneveld, line 30: For readability (but perhaps it's a hassle to change it now) I would propose to put `IsOptional` at the top, so it's easier to spot when a field it optional.

api/src/socketio.ts
```diff
@@ -86,26 +110,45 @@ export class Socketio {
         const changeReq = plainToInstance(ChangeReqDto, data);
```
laurensgroeneveld, line 110: As a general not this file is growing quite a bit now so maybe in a follow-up PR some parts could be extracted, like the validation, which might make testing those specific parts easier as well as they don't rely on socket.io.

laurensgroeneveld (approved):
Some small comments but I'll approve because the rest looks good 👍 Would like some more test coverage for the validation of added documents but that can also be handled in another PR with the permission check for example.

api/src/dto/ChangeReqAckDto.ts
ivanslabbert: I left it as @IsString to allow using our current test data set. Might change it later though...

ivanslabbert: I left it as @IsString to allow using our current test data set. Might change it later though...

ivanslabbert: I left it as @IsString to allow using our current test data set. Might change it later though...

api/src/dto/ContentDto.ts
ivanslabbert: Agree

api/src/dto/ChangeReqAckDto.ts
ivanslabbert: Seems to also work.

api/src/dto/ChangeDto.ts
ivanslabbert: Throwing an exception makes it not show the message at all. So not very helpful here. Custom messages are however nice, and can be used when we add more detailed validations.

PR #35, opened 2024-02-15
Submit CMS changes to API

laurensgroeneveld:
Sync saved local changes with API when online

Uses hardcoded groups (probably needs to be solved in another PR when we can get the groups from the API)


api/src/dto/_baseDto.ts
```diff
@@ -17,7 +17,7 @@ export class _baseDto {
     @IsEnum(DocType)
     type: DocType;
 
-    @IsDate()
+    // TODO check if it's a valid timestamp
```
laurensgroeneveld, line 20: When I posted it with a timestamp I got an error "needs to be instance of Date"

ivanslabbert: Might need to add @IsNumber to prevent sending strings etc?

api/src/dto/ContentDto.ts
```diff
@@ -51,10 +53,11 @@ export class ContentDto extends _contentBaseDto {
     video?: Uuid;
 
     @IsDate()
-    @IsNotEmpty()
+    @IsOptional()
```
ivanslabbert, line 56: I think we can require it to be set before saving but by default set it to the document creation date (in the CMS)? The editor should be able to set it to a different date if he wants to.

Will we only include the document in Sync after the publish date? (perhaps not in MVP but later) to enable "scheduled publish".

api/src/dto/ContentDto.ts
```diff
@@ -51,10 +53,11 @@ export class ContentDto extends _contentBaseDto {
...
     publishDate: number;
 
     @IsDate()
+    // TODO required if status set to published
```
ivanslabbert, line 60: Do not think we need to have the expiry date set, unless you want the document to expire?

Notes for CMS functionality:
Would be nice if it is possible to both be able to set a date or number of days.

api/src/dto/_baseDto.ts
ivanslabbert, line 20: Set to @IsNumber for now (unless there is a better way to check if it is a Unix timestamp). Could perhaps check if the number is bigger than 1 January 2020 or so?

api/src/socketio.ts
```diff
@@ -23,6 +23,7 @@ import { LanguageDto } from "./dto/LanguageDto";
 import { PostDto } from "./dto/PostDto";
 import { TagDto } from "./dto/TagDto";
 import { UserDto } from "./dto/UserDto";
+import { _baseDto } from "./dto/_baseDto";
```
ivanslabbert, line 26: Not sure why we need to import _baseDto?

ivanslabbert:
This PR might have some conflicts with PR #36

api/src/dto/ContentDto.ts
laurensgroeneveld: My thinking was that the document might not always be published the same day it was created, and that we can instead provide a dropdown in the CMS with options:
- Set to current date when published
- Custom date.. -> datepicker

So similar to how we said we would handle expiry date.

Then it's also handled for translations created after the original translation was published, and we avoid the problem of the current site where it defaults to a wrong (older) date.

api/src/dto/ContentDto.ts
laurensgroeneveld: > Do not think we need to have the expiry date set, unless you want the document to expire?

Right, this comment belongs to publishedDate instead.

PR #36, opened 2024-02-16
Add ChangeReq permission validation

ivanslabbert:


api/src/db/db.service.ts
```diff
@@ -137,6 +138,56 @@ export class DbService {
...
     }
 
+    /**
+     * Get multiple documents by ID and type
+     * @param {Uuid[]} docIds - Document IDs to be included in search.
+     * @param {DocType[]} types - Document types to be included in search
+     * @returns - Promise containing the query result
+     */
+    getDocs(docIds: Uuid[], types: DocType[]): Promise<nano.MangoResponse<unknown>> {
+        // TODO: Write unit test
```
laurensgroeneveld, line 148: Remember to check.

api/src/permissions/AccessMap.ts
```diff
@@ -0,0 +1,38 @@
...
+ */
+export class AccessMap {
+    /**
+     * Get list of effective access (including inherited access) for the passed group IDs per docType and permission
+     * @param types - Document for which effective access should be calculated
+     * @param permission - Permission for which effective access should be calculated
+     */
+    public calculateAccess: (types: DocType[], permission: AclPermission) => Array<Uuid>;
+
+    public Map: Map<Uuid, Map<DocType, Map<AclPermission, boolean>>>;
```
laurensgroeneveld, line 14: Nitpick: Why does `Map` have a capital letter since it's just a normal public variable?

api/src/permissions/permissions.service.ts
```diff
@@ -1,31 +1,33 @@
...
-export class Group {
+export class PermissionSystem {
     private _typePermissionGroupRequestorMap = new Map<
         DocType,
         Map<AclPermission, Map<Uuid, Map<Uuid, boolean>>>
     >();
     private _groupTypePermissionMap = new Map<Uuid, Map<DocType, Map<AclPermission, boolean>>>();
     private _aclMap = new Map<Uuid, AclMapEntry>();
-    private _childGroups = new Map<Uuid, Group>();
+    private _childGroups = new Map<Uuid, PermissionSystem>();
```
laurensgroeneveld, line 30: These variables are still named "group".

api/src/permissions/validateChangeReq.ts
```diff
@@ -0,0 +1,135 @@
...
+                        if (!assignGroups.includes(groupId)) {
+                            return "No 'Assign' access to one or more tags";
+                        }
+                    }
+                }
+            }
+        }
+    }
+
+    return ""; // Empty response means the vaildation passed (i.e. no error)
```
laurensgroeneveld, line 134: Consider returning a specific object with a boolean result and message to make it more obvious when validation succeeds or fails.

api/src/socketio.ts
```diff
@@ -103,54 +79,41 @@ export class Socketio {
...
-        // Check included document existance and type validity
-        if (!changeReq.doc.type || !Object.values(DocType).includes(changeReq.doc.type)) {
-            this.emitAck(
-                socket,
-                AckStatus.Rejected,
-                data.reqId,
-                `Submitted "${changeReq.doc.type}" document validation failed:\nInvalid document type`,
-            );
+        // Reject non user-editable types
+        // TODO: add test
```
laurensgroeneveld, line 93: Just to check if this is still relevant.

laurensgroeneveld (approved):
I skimmed over it since we did a deep-dive on the code earlier together. Just some code quality comments but looks good otherwise 👍

api/src/db/db.service.ts
ivanslabbert: Updated to use nano.fetch() instead of nano.find(), and implemented document type field checking in code.
Also added a test for it.

api/src/permissions/AccessMap.ts
ivanslabbert: Fixed

api/src/permissions/permissions.service.ts
ivanslabbert: Keeping "group" naming still makes sense as the Permission System is in fact a tree structure where each node is representing a group. Added a comment in the code to explain this a bit.

api/src/socketio.ts
ivanslabbert: It is still relevant, but we need to think how we are going to do Socket.io specific testing when authentication is implemented.

An option would here would be to move the check to the permission system, where it will be easier to test.

api/src/permissions/validateChangeReq.ts
ivanslabbert: Done

api/src/socketio.ts
ivanslabbert: Document type checking moved to PermissionSystem.ts

PR #37, opened 2024-02-21
Database indexing

ivanslabbert:
- Added indexing for CouchDB mango queries, and added tests to ensure indexing warnings are not returned.
- Created a type to be used when returning any document query data from db.service.ts

### DIFF
--- a/api/src/db/db.service.ts
+++ b/api/src/db/db.service.ts
@@ -60,10 +60,10 @@ export class DbService {
                 reject("Invalid document: The passed document does not have an '_id' property");
             }
             this.getDoc(doc._id)
-                .then((e) => {
-                    let existing;
-                    if (e.docs && e.docs.length > 0) {
-                        existing = e.docs[0];
+                .then((res) => {
+                    let existing; // if no existing document, this will be undefined
+                    if (res.docs && res.docs.length > 0) {
+                        existing = res.docs[0];
                     }
 
                     let rev: string;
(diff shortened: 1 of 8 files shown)


api/src/db/db.service.ts
```diff
@@ -52,7 +60,12 @@ export class DbService {
                 reject("Invalid document: The passed document does not have an '_id' property");
             }
             this.getDoc(doc._id)
-                .then((existing: any) => {
+                .then((e) => {
+                    let existing;
+                    if (e.docs && e.docs.length > 0) {
+                        existing = e.docs[0];
+                    }
```
laurensgroeneveld, line 67: Without the knowledge that this was done to avoid rewriting code this becomes a bit confusing. It would warrant a comment explanation I feel.

api/src/db/db.service.ts
```diff
@@ -250,64 +262,136 @@ export class DbService {
...
+
+            Promise.all(pList)
+                .then((res) => {
+                    const docs = res.flatMap((r) => r.docs);
+                    const warnings = res.flatMap((r) => r.warning).filter((w) => w);
+                    if (warnings.length > 0) {
+                        resolve({ docs, warnings });
+                    } else {
+                        resolve({ docs });
+                    }
```
laurensgroeneveld, line 364: Suggestion: use a ternary to avoid the if-else condition.

```suggestion
                        resolve({
                            docs, 
                            warnings: warnings.length > 0 ? warnings : undefined,
                         });

```

api/src/db/db.service.ts
```diff
@@ -250,64 +262,136 @@ export class DbService {
...
+                },
+                use_index: "type-id-index",
+            };
+            this.db
+                .find(query)
+                .then((res) => {
+                    if (res.warning) {
+                        resolve({ docs: res.docs, warnings: [res.warning] });
+                    } else {
+                        resolve({ docs: res.docs });
```
laurensgroeneveld, line 389: Same comment.

laurensgroeneveld (approved):
Just some small comments on the code structure, looking good otherwise 👍 Nice work adding tests for the indexes.

api/src/db/db.service.ts
laurensgroeneveld: Or rewrite the code of course 😄

PR #42, opened 2024-02-26
Process change requests and store change documents to database after validation

ivanslabbert:


api/src/db/db.service.spec.ts
```diff
@@ -26,19 +26,38 @@ describe("DbService", () => {
...
 
         const testGet = (await service.getDoc(uuid)) as any;
 
         expect(testGet.docs[0]._id).toBe(uuid);
         expect(testGet.docs[0].testData).toBe("test123");
+        expect(res.changes.testData).toBe("test123");
+        expect(res.changes._id).toBe(uuid);
+    });
+
+    it("can calculate a document diff and return the diff with the upsert result", async () => {
```
laurensgroeneveld, line 46: Suggestion: add a third field that isn't changed and test that it isn't returned in the diff.

laurensgroeneveld (approved):

api/src/db/db.service.spec.ts
ivanslabbert: Agree, added.

PR #44, opened 2024-02-26
Add changedByUser field and implement memberOf and acl fields in change documents

ivanslabbert:


api/src/changeRequests/processChangeRequest.spec.ts
```diff
@@ -30,7 +30,7 @@ describe("processChangeRequest", () => {
             doc: {},
         };
 
-        await processChangeRequest(changeRequest, new AccessMap(), db).catch((err) => {
+        await processChangeRequest("", changeRequest, new AccessMap(), db).catch((err) => {
             expect(err.message).toBeTruthy();
```
laurensgroeneveld, line 34: Maybe you want to check a specific message here? Otherwise the test could be green even though there is an error for a different reason. Although I understand you're not really testing the validation here. Maybe add a spy check?

api/src/changeRequests/processChangeRequest.spec.ts
```diff
@@ -42,15 +42,25 @@ describe("processChangeRequest", () => {
                 _id: "new-language",
                 type: "language",
                 memberOf: ["group-languages"],
+                // Even though acl's are not valid for "language" documents, we include it here for unit testing purposes.
```
laurensgroeneveld, line 45: As discussed, then this field shouldn't be passed through to the database with the change request. Investigate if this is a class transformer issue, as it should convert to a class instance which presumably only has the fields you declare.

api/src/changeRequests/processChangeRequest.spec.ts
```diff
@@ -74,8 +84,8 @@ describe("processChangeRequest", () => {
                 name: "Xhoza",
             },
         };
-        await processChangeRequest(changeRequest1, accessMap, db);
-        const processResult = await processChangeRequest(changeRequest2, accessMap, db);
+        await processChangeRequest("", changeRequest1, accessMap, db);
```
laurensgroeneveld, line 87: I understand we do it for testing, but shouldn't an empty user id be invalid?

api/src/changeRequests/processChangeRequest.ts
```diff
@@ -28,6 +29,15 @@ export async function processChangeRequest(
...
         changeDoc.changes = upsertResult.changes;
+        changeDoc.changedByUser = userId;
+
+        if (upsertResult.changes.memberOf) {
+            changeDoc.memberOf = upsertResult.changes.memberOf;
+        }
+
+        if (upsertResult.changes.acl) {
+            changeDoc.acl = upsertResult.changes.acl;
+        }
```
laurensgroeneveld, line 40: I think tests are missing for this behaviour.

api/src/changeRequests/processChangeRequest.spec.ts
ivanslabbert: The user id is passed from the socket.io socket instance, and not from the user. I currently am not enforcing it as it is not a value that will be passed from the CMS / app

api/src/changeRequests/processChangeRequest.spec.ts
ivanslabbert: This seems to be quite a big flaw with the current DTO validation. Will fix with the next push

api/src/changeRequests/processChangeRequest.spec.ts
ivanslabbert: Agree. This actually uncovered a bigger issue with DTO validation, thanks!

api/src/changeRequests/processChangeRequest.spec.ts
ivanslabbert: Made the test more specific

api/src/changeRequests/processChangeRequest.ts
ivanslabbert: Added tests for this in processChangeRequest.spec.ts:
"can insert a change document into the database for a correctly submitted new document with group membership"
"can insert a change document into the database for a correctly submitted new group document"

api/src/changeRequests/processChangeRequest.ts
```diff
@@ -17,18 +18,30 @@ export async function processChangeRequest(
...
 
         return db.insertDoc(changeDoc);
+    } else {
+        return upsertResult;
     }
```
laurensgroeneveld, line 46: Typical LG comment: the `else` is not needed since you already early return inside the `if` 😅

api/src/changeRequests/validateChangeRequestAccess.ts
```diff
@@ -37,10 +37,25 @@ export async function validateChangeRequestAccess(
...
         const editGroups = accessMap.calculateAccess([DocType.Group], AclPermission.Edit);
         if (!editGroups.includes(doc._id)) {
-            return {
-                validated: false,
-                error: "No access to 'Edit' document type 'Group'",
-            };
+            // If this is a new group, the permission system does not yet know about it and it will not be in the access map.
+            // We need to check if the user has edit access to at least one of the groups in the ACL list, which will imply that
+            // the user has edit access to the new group through inheritance.
+            const aclEditGroups = accessMap.calculateAccess([DocType.Group], AclPermission.Edit);
```
laurensgroeneveld, line 43: Is it needed to calculate this again? This variable seems to be the same as `editGroups` above.

laurensgroeneveld (approved):

api/src/changeRequests/validateChangeRequestAccess.ts
ivanslabbert: Silly of me yes

PR #48, opened 2024-02-28
Send individual change requests from frontend queue

laurensgroeneveld:
Instead of batching them together which doesn't actually work as the watcher already triggers for the first `put` to localChanges, even with `bulkPut`


ivanslabbert:
This change simplifies changeRequest submission a lot, so I think it is a positive move.

ivanslabbert (approved):

PR #53, opened 2024-03-01
Subscribe to change events from database and pass changed documents to connected clients through Socket.io "data" messages

ivanslabbert:


laurensgroeneveld (commented):
Some nitpicks and questions.

api/src/socketio.ts
```diff
@@ -27,12 +27,56 @@ type ClientDataReq = {
...
+    afterInit(server: Server) {
+        // Subscribe to database changes and broadcast change to all group rooms to which the document belongs
+        this.db.on("update", (update: any) => {
+            // Only include documents with a document type property
+            if (!update.type) {
+                return;
+            }
+
+            // Broadcast CMS specific documents to CMS group rooms
+            if (this.cmsDocTypes.includes(update.type)) {
```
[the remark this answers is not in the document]
ivanslabbert: The Change documents should only be used for audit trail purposes. Changes to documents are emitted to clients as full documents and not as Change docs. This also prevents a client needing to "build" the current version of the document by applying a stack of change documents.

api/src/test/socketioTestClient.ts
```diff
@@ -0,0 +1,56 @@
...
+        }
+
+        // Wait for the server to emit the expected events
+        setTimeout(() => {
+            if (testClient) {
+                testClient.off("data", dataHandler);
+                testClient.disconnect();
+            }
+            resolve(res);
+        }, timeout);
```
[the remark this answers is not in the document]
ivanslabbert: I had tests failing at 500ms, so made the default 1s. Not sure why it is so slow though...

api/src/permissions/permissions.service.ts
```diff
@@ -58,6 +58,71 @@ export class PermissionSystem {
...
+    //             if (
+    //                 g._groupTypePermissionMap[type] &&
+    //                 g._groupTypePermissionMap[type][permission]
+    //             ) {
+    //                 resultMap.add(groupId);
+    //             }
+    //         });
+    //     });
+    //     return Array.from(resultMap);
+    // }
```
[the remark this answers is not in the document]
ivanslabbert: Code is to be used later - will fix with next PR

api/src/test/socketioTestClient.ts
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * Send a change request to the server and wait for the server to emit the expected number of 'data' events
+ * @param changeRequest
+ * @param cms - true if the test client should be a CMS client
+ * @param timeout - time to wait for server to emit the expected messages
+ * @returns
+ */
+export const socketioTestClient = (
+    cms: boolean,
+    version: number = 0,
```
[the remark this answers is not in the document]
ivanslabbert: Created a new type as we pass more values than allowed for in the ClientDataReq type

api/src/db/db.service.spec.ts
```diff
@@ -282,25 +299,31 @@ describe("DbService", () => {
...
-        const oldest = await service.getOldestChangeTime();
-        const newest = await service.getLatestDocUpdatedTime();
-        const res2: DbQueryResult = await service.getDocsPerGroup("user-public", {
-            groups: ["group-public-content"],
-            types: [DocType.Post, DocType.Tag, DocType.Group],
-            from: Math.floor(newest - (newest - oldest) / 2),
-        });
-        expect(res2.warnings).toBe(undefined);
-    });
+    // TODO: Fix this test after researching CouchDB indexing
```
laurensgroeneveld, line 302: Instead of commenting out the whole test, rather skip it by marking it as `xit` or `it.skip`. Then it's both easier to find and it serves as a reminder every time running the tests.

api/src/db/db.service.ts
```diff
@@ -461,12 +459,14 @@ export class DbService extends EventEmitter {
...
-            }
-            if (doc1.type === doc2.type) {
-                acc.type = doc1.type;
+            acc._id = doc1._id;
+            acc.type = doc1.type;
+
+            // Include parentType if it exists
+            if (doc1.parentType) {
+                acc.parentType = doc1.parentType;
             }
```
laurensgroeneveld, line 468: Do we indeed need type and parentType in the diff? Then we would need to exclude those when rendering a visual change diff in the CMS.

api/src/permissions/permissions.service.ts
```diff
@@ -42,44 +42,57 @@ export class PermissionSystem {
...
+    // static getAccessMap(groupIds: Array<Uuid>): AccessMap {
+    //     const resultMap = new AccessMap();
+    //     groupIds.forEach((id: Uuid) => {
+    //         const g = groupMap[id];
+    //         if (!g) return;
+
+    //         resultMap.map = { ...resultMap, ...g._groupTypePermissionMap };
+    //     });
+    //     return resultMap;
+    // }
```
laurensgroeneveld, line 60: Not sure if you're immediately working on this in a follow-up PR, otherwise I would suggest to stash this somewhere else and remove the commented code. It's all too easy for dead code like this to continue living in the far future as nobody knows whether we need it or not 😅 So as a general rule I feel it's safer to not commit explicitly commented code.

api/src/socketio.ts
```diff
@@ -41,38 +41,41 @@ export class Socketio implements OnGatewayInit {
...
 
-            // All other valid documents - broadcast to App and CMS group rooms
-            if (this.appDocTypes.includes(update.type)) {
-                server
-                    .to(update.memberOf.map((group) => `cms-${group}`))
-                    .to(update.memberOf.map((group) => `app-${group}`))
-                    .emit("data", [update]);
+            // Group documents
+            if (update.acl) {
+                // If the document is a "change" document, the group id is stored in the docId property
```
laurensgroeneveld, line 69: I think this comment belongs to the `if` a few lines below.

api/src/socketio.ts
```diff
@@ -41,38 +41,41 @@ export class Socketio implements OnGatewayInit {
...
+                let docType = update.type;
+                let groupId = update._id;
 
+                if (update.type == "change" && update.changes) {
+                    docType = update.changes.type;
+                    groupId = update.docId;
+                }
```
laurensgroeneveld, line 76: I'm a little confused why we're testing for change here but perhaps you can take me through this method to make me understand it better.

laurensgroeneveld (commented):
Just some small comments and questions on the changes since last review.

api/src/socketio.ts
```diff
@@ -27,12 +27,62 @@ type ClientDataReq = {
...
+
+            // We are using a socket.io room per document type per group. Change documents are broadcasted to the document-group rooms of the documents they reference.
+            // Content documents are broadcasted to their parent document-group rooms.
+
+            // Non-group documents
+            if (update.memberOf) {
+                let docType = update.type;
+                if (update.type == "change" && update.changes) {
+                    if (update.changes.type == "content") {
+                        docType = update.changes.parentType;
```
ivanslabbert, line 52: Better to rather do a DB call instead of "caching" data in the database

api/src/db/db.service.ts
ivanslabbert: Removed parentType field, but added the same logic to make sure the parentID are always included in diff's

api/src/dto/ContentDto.ts
```diff
@@ -15,6 +15,10 @@ export class ContentDto extends _contentBaseDto {
     @Expose()
     parentId: Uuid;
 
+    @IsOptional()
+    @IsString()
+    parentType?: DocType;
```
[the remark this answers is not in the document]
ivanslabbert: Need to keep parentType as it is used in database queries. Will remove if we simplify the permissions

laurensgroeneveld (approved):
LGTM :shipit:

PR #56, opened 2024-03-06
Refactor access map and time filtering in DbService and permission service

ivanslabbert:


api/src/db/db.service.ts
```diff
@@ -323,27 +324,43 @@ export class DbService extends EventEmitter {
...
+                });
+            }
+
+            if (options.to) {
+                selectors.push({
+                    updatedTimeUtc: {
+                        $lt: options.to + this.syncTolerance,
+                    },
+                });
+            }
```
laurensgroeneveld, line 356: I'm a bit confused by this. `selectors` does not seem to be used, does it work by reference? Why is it needed to have a separate variable for it?

api/src/socketio.ts
laurensgroeneveld, line 1: I'm missing tests for the changes in this file.

laurensgroeneveld (changes requested):
Missing some tests and a small question.

api/src/db/db.service.ts
ivanslabbert: Simplified this a bit to make it more readable

api/src/socketio.ts
ivanslabbert: - Added a test to check if an AccessMap is returned after clientDataReq.
- Added a skipped test as placeholder for test to check if historical data is returned after a user is assigned additional groups. This should be implemented after JWT verification (and group assignments) are implemented in the API. As all requests are hard-coded with "group-super-admins" permissions at the moment, testing this is a bit harder, and will have to be changed as soon as JWT verification is implemented.

laurensgroeneveld: > This should be implemented after JWT verification

Can you add it as a note to the authentication ticket? So we don't forget about it.

laurensgroeneveld (approved):
LGTM

PR #60, opened 2024-03-11
Add slug with duplicate detection to Content Form

ivanslabbert:


cms/src/components/content/ContentForm.vue
```diff
@@ -94,6 +96,7 @@ const save = async (validatedFormValues: typeof values, status: ContentStatus) =
         ...contentValues,
         publishDate,
         status,
+        // slug: Slug.makeUnique(slug), // Ensure slug is still unique before saving
```
laurensgroeneveld, line 99: Left a comment here.

laurensgroeneveld (approved):
Nicely done 🤩

cms/src/components/content/ContentForm.vue
ivanslabbert: Should have been enabled... All the values from the form validator seems to be optional. Not sure if it is ok to use the "!" here?

PR #61, opened 2024-03-11
Fix slug not taking the current content into account

laurensgroeneveld:
Meaning it would always make the slug unique

### DIFF
--- a/cms/src/util/slug.spec.ts
+++ b/cms/src/util/slug.spec.ts
@@ -13,7 +13,7 @@ describe("Slug", () => {
     it("should generate a slug from a title", async () => {
         const title = "Sample Title";
 
-        const generatedSlug = await Slug.generate(title);
+        const generatedSlug = await Slug.generate(title, "fake-id");
 
         expect(generatedSlug).toBe("sample-title");
     });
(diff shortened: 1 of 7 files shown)


cms/src/util/slug.spec.ts
```diff
@@ -22,8 +22,17 @@ describe("Slug", () => {
         db.docs.put({ ...mockEnglishContentDto, slug: "sample-title" } as ContentDto);
         const title = "Sample Title";
 
-        const generatedSlug = await Slug.generate(title);
+        const generatedSlug = await Slug.generate(title, "different-id");
 
         expect(generatedSlug).toMatch(/sample-title-[0-9]([0-9])*/);
     });
+
+    it("should not make the title for the same content unique", async () => {
```
ivanslabbert, line 30: Is the test title correct? Should it not read something like "should not make the SLUG for the same content [document] complete"? (Corrections in uppercase and additional clarifying text in [ ])

ivanslabbert (changes requested):

cms/src/util/slug.spec.ts
laurensgroeneveld: Renamed the variable to `documentId` as well to be a bit more generic as it could be other things than content with slugs in the future.

ivanslabbert (approved):

PR #62, opened 2024-03-12
Validate and automatically rectify content document's slug in API

ivanslabbert:


api/src/changeRequests/processChangeRequest.spec.ts
```diff
@@ -118,4 +119,58 @@ describe("processChangeRequest", () => {
...
+
+        const res = await processChangeRequest("", changeRequest, ["group-super-admins"], db);
+        const dbDoc = await db.getDoc(changeRequest.doc._id);
+
+        expect(res.ok).toBe(true);
+        expect(dbDoc.docs[0].slug).toBe("this-is-a-test-slug");
+    });
+
+    it("can validate a unique slug for a content document that exists", async () => {
+        // This test relies on the previous test to have run first
```
laurensgroeneveld, line 137: Not a huge fan of this as you never know if the order of tests is guaranteed - maybe Jest will change it in a future version. Instead I would suggest to specifically insert a different item into the database at the start of this test and then use that to verify it.

api/src/changeRequests/processChangeRequest.spec.ts
```diff
@@ -118,4 +119,58 @@ describe("processChangeRequest", () => {
...
+
+        const res = await processChangeRequest("", changeRequest, ["group-super-admins"], db);
+        const dbDoc = await db.getDoc(changeRequest.doc._id);
+
+        expect(res.ok).toBe(true);
+        expect(dbDoc.docs[0].slug).toBe("this-is-a-test-slug");
+    });
+
+    it("can rectify a non-unique slug by appending a random number to the end of the slug", async () => {
+        // This test relies on the previous test to have run first
```
laurensgroeneveld, line 151: Same comment.

laurensgroeneveld (approved):
Just a comment on the robustness of the tests, looks good otherwise.

PR #64, opened 2024-03-12
Edit slug

ivanslabbert:


cms/src/components/content/ContentForm.vue
```diff
@@ -171,9 +170,25 @@ const hasTag = computed(() => {
...
+const updateSlugFromTitle = async (title: string) => {
+    // Check if the slug is still the default value
+    if (
+        previousTitle == "" ||
+        values.slug?.replace(/-[0-9]*$/g, "") == Slug.generateNonUnique(previousTitle)
+    ) {
+        await updateSlug(title);
+    }
+    previousTitle = title;
 };
```
ivanslabbert, line 192: Not sure to what degree we need to test the slug functionality?

laurensgroeneveld: Would be nice to have a test that it doesn't update the slug if you update the title and the slug has been modified.

ivanslabbert: Busy writing it now

laurensgroeneveld (changes requested):
Two UX things I noticed:

- If you empty the title input, then start typing, the slug will reset to the slugified title, even if I edited it
- If I customize the slug, change the title doesn't change the slug. However if I change the title so the outputted slug is the same, the slug resumes updating with the title, which feels unexpected.

And a thing that's not due to your PR, pressing Enter on any input field in the form will close the whole card 🙈 Especially on this slug one I would not expect that to happen. We can solve that separately, likely a focus issue.

cms/src/components/content/ContentForm.vue
```diff
@@ -187,6 +202,14 @@ const addTag = (tag: Tag) => {
 const removeTag = (tag: Tag) => {
     selectedTags.value = selectedTags.value.filter((t) => t._id != tag._id);
 };
+
+const startEditingSlug = () => {
+    isEditingSlug.value = true;
+    nextTick(() => {
+        console.log(slugInput.value);
+        slugInput.value?.focus();
+    });
```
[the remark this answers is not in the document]
ivanslabbert: Without the nextTick it is not able to focus the text input.

cms/src/components/content/ContentForm.spec.ts
```diff
@@ -256,6 +256,41 @@ describe("ContentForm", () => {
...
+
+    it("saves the new slug when the text input loses focus and hides the input", async () => {
+        const wrapper = mount(ContentForm, {
+            props: {
+                post: mockPost,
+                content: mockEnglishContent,
+            },
+        });
+        const input = wrapper.find("input[name='slug']");
+        const slug = wrapper.find("span[name='slugSpan']");
```
laurensgroeneveld, line 282: As discussed, this makes the test very dependent on the HTML layout, so even small changes like changing the `span` to a `div` will make the test failing (and I think they should only fail if you actually change the functionality to a certain degree). It does make the tests a little nicer to read so I understand if you want to keep it. I would just make two changes in any case:

- remove the `span` (the `name` is already unique enough)
- Change `name` to `data-test`, which I've used in other tests for that purpose. `name` is not valid on a `span`, but `data-*` elements are. This comment holds for the every test except the `input`s where `name` makes sense.

ivanslabbert:
> 2. As discussed, I feel like the slug should not be editable when the post is published, because then links that have been shared on socials etc. are suddenly not working anymore (a likely future feature we need is redirects, but that's a whole other thing).

This is implemented now

cms/src/components/content/ContentForm.spec.ts
```diff
@@ -248,12 +248,120 @@ describe("ContentForm", () => {
             },
         });
 
+        const slug = wrapper.find("[name='slugSpan']");
+
         expect(wrapper.text()).toContain(mockEnglishContent.slug);
```
laurensgroeneveld, line 253: Can also use `slug` here.

cms/src/components/content/ContentForm.spec.ts
```diff
@@ -248,12 +248,120 @@ describe("ContentForm", () => {
             },
         });
 
+        const slug = wrapper.find("[name='slugSpan']");
```
laurensgroeneveld, line 251: Still uses `name` instead of `data-test`.

cms/src/components/content/ContentForm.spec.ts
```diff
@@ -248,12 +248,120 @@ describe("ContentForm", () => {
...
+
         expect(wrapper.text()).toContain(mockEnglishContent.slug);
 
         await wrapper.find("input[name='title']").setValue("Updated Title");
 
         await flushPromises();
-        expect(wrapper.text()).toContain("Slug:updated-title");
+        waitForExpect(() => {
```
laurensgroeneveld, line 258: Missing `await`, so the test is always succeeding.

cms/src/components/content/ContentForm.spec.ts
```diff
@@ -248,12 +248,120 @@ describe("ContentForm", () => {
...
+
+        await wrapper.find("button[data-test='editSlugButton']").trigger("click");
+        await flushPromises();
+        await input.setValue("new-slug");
+        await input.trigger("blur");
+        await flushPromises();
+        await title.setValue("");
+        await flushPromises();
+        await title.setValue("New title");
+        await flushPromises();
```
laurensgroeneveld, line 339: I _think_ you don't need to have quite so many `flushPromises` 😅

PR #70, opened 2024-03-15
Authenticate API socket.io connection

ivanslabbert:
- add JWT based authentication for Socket.io connections in API
- include the JWT as auth token with the client Socket.io session
- a few small fixes


api/src/configuration.ts
```diff
@@ -12,16 +12,25 @@ export type SyncConfig = {
...
-        connectionString: process.env.DB_CONNECTION_STRING,
-        database: process.env.DB_DATABASE,
-        maxSockets: parseInt(process.env.DB_MAX_SOCKETS, 10) || 512,
-    } as DatabaseConfig,
-    sync: {
-        tolerance: parseInt(process.env.PORT, 10) || 1000,
-    } as SyncConfig,
-});
+export type Configuration = {
+    permissionMap: any;
```
laurensgroeneveld, line 16: Would be nice to have some typing here, not strictly necessary though.

cms/src/stores/socketConnection.spec.ts
```diff
@@ -9,18 +9,27 @@ import { flushPromises } from "@vue/test-utils";
...
-        emit: vi.fn(),
-    };
-});
-
-vi.mock("socket.io-client", () => {
-    return {
-        io: vi.fn().mockImplementation(() => socketMocks),
-    };
-});
+// const socketMocks = vi.hoisted(() => {
```
laurensgroeneveld, line 12: Left some commented code here, it can be deleted I think.

api/src/jwt/jwtPermissionMap.ts
```diff
@@ -0,0 +1,74 @@
...
+        if (map.jwt && map.jwt.userId) {
+            Object.keys(map.jwt.userId).forEach((key) => {
+                map.jwt.userId[key] = eval(map.jwt.userId[key]);
+            });
+        }
+
+        return map as PermissionMap;
+    } catch (err) {
+        // TODO: Add error logging provider
+        console.log(err.message);
```
laurensgroeneveld, line 38: Is this correct? I see it in some places that we do a generic try-catch, in this case we're not returning a default. Can the code that uses this method continue then, or should this actually throw? Maybe a test is needed for this?

api/src/jwt/jwtPermissionMap.spec.ts
```diff
@@ -0,0 +1,24 @@
+import configuration, { Configuration } from "../configuration";
+import { getJwtPermission, parsePermissionMap } from "./jwtPermissionMap";
+
+describe("jwtPremissionMap", () => {
+    let config: Configuration;
+    beforeAll(() => {
+        // This test relies on the .env.test file configured as per .env.test.example
```
laurensgroeneveld, line 7: Perhaps we should instead mock the resulting object here, so the test is not too brittle?

laurensgroeneveld (commented):
Just a few notes from my phone, I'll run the code on my laptop as well when I'm back but it looks good in general 👍

api/src/configuration.ts
ivanslabbert: We are typing the validated permission map. This property is holding the raw string value stored in the .env file

api/src/jwt/jwtPermissionMap.ts
ivanslabbert: Fixed in following push

laurensgroeneveld (approved):
Nice work!

PR #82, opened 2024-03-19
Clear revoked documents from CMS

ivanslabbert:
- Remove revoked documents from the CMS IndexedDB upon receipt of an access map

### DIFF
--- a/cms/src/db/repositories/baseRepository.ts
+++ b/cms/src/db/repositories/baseRepository.ts
@@ -37,7 +37,7 @@ export class BaseRepository {
                 (group) =>
                     (group.type === DocType.Group ||
                         (group.type === DocType.Change && group.docType === DocType.Group)) &&
-                    group.acl &&
+                    group.acl! &&
                     !group.acl.some(
                         (acl) =>
                             acl.type === DocType.Group &&
--- a/cms/src/db/repositories/baseRepository.ts
+++ b/cms/src/db/repositories/baseRepository.ts
@@ -51,7 +51,8 @@ export class BaseRepository {
         return db.docs.filter(
             (doc) =>
                 doc.memberOf !== undefined &&
-                (doc.type === docType || (doc.type == DocType.Change && doc.docType == docType)) &&
+                (doc.type === docType ||
+                    (doc.type === DocType.Change && doc.docType === docType)) &&
                 !doc.memberOf.some((groupId) => groupIds.includes(groupId)),
         );
     }
--- a/cms/src/db/repositories/baseRepository.ts
+++ b/cms/src/db/repositories/baseRepository.ts
@@ -33,28 +33,42 @@ export class BaseRepository {
     whereNotMemberOf(groupIds: Array<Uuid>, docType: DocType) {
         // Query groups and group changeDocs
         if (docType === DocType.Group) {
-            return db.docs.filter(
-                (group) =>
-                    (group.type === DocType.Group ||
-                        (group.type === DocType.Change && group.docType === DocType.Group)) &&
-                    group.acl! &&
-                    !group.acl.some(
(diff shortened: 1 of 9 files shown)


cms/src/stores/socketConnection.ts
```diff
@@ -39,7 +49,67 @@ export const useSocketConnectionStore = defineStore("socketConnection", () => {
...
+                .forEach((_permission) => {
+                    if (docTypePermissions[_permission as AclPermission]) {
+                        if (!groups[docType as DocType]) groups[docType as DocType] = [];
+                        groups[docType as DocType]?.push(groupId as Uuid);
+                    }
+                });
+        });
+    });
+    return groups;
+}
```
ivanslabbert, line 115: This function can probably move to a new file when permission checking is implemented in the CMS

cms/src/types/dto.ts
```diff
@@ -77,3 +86,19 @@ export type ChangeReqAckDto = {
...
+    Assign = "assign",
+    Edit = "edit",
+    Translate = "translate",
+    Publish = "publish",
+    Delete = "delete",
+}
+
+// export type AccessMap = Map<Uuid, Map<DocType, Map<AclPermission, boolean>>>;
+export type AccessMap = {
+    [a: Uuid]: { [b in DocType]?: { [c in AclPermission]?: boolean | undefined } };
```
laurensgroeneveld, line 101: Nitpick: for types commonly `T` is used as a placeholder, followed by `U`, `V` etc.

Also, there's a commented line with an older version above this one.

cms/src/db/repositories/baseRepository.ts
```diff
@@ -19,6 +19,44 @@ export class BaseRepository {
...
+     * Return a list of documents and change documents of specified DocType that are NOT members of the given groupIds
+     */
+    whereNotMemberOf(groupIds: Array<Uuid>, docType: DocType) {
+        // Query groups and group changeDocs
+        if (docType === DocType.Group) {
+            return db.docs.filter(
+                (group) =>
+                    (group.type === DocType.Group ||
+                        (group.type === DocType.Change && group.docType === DocType.Group)) &&
+                    group.acl! &&
```
laurensgroeneveld, line 40: Is this `!` needed here? Shouldn't matter too much but if it's not it feels a bit cleaner.

cms/src/db/repositories/baseRepository.ts
```diff
@@ -19,6 +19,44 @@ export class BaseRepository {
...
+            );
+        }
+
+        // Query other documents
+        return db.docs.filter(
+            (doc) =>
+                doc.memberOf !== undefined &&
+                (doc.type === docType ||
+                    (doc.type === DocType.Change && doc.docType === docType)) &&
+                !doc.memberOf.some((groupId) => groupIds.includes(groupId)),
```
laurensgroeneveld, line 56: Suggestion:  These filter clauses are a bit hard to read. It might help to split them up in local variables with clearer names.

laurensgroeneveld (approved):
Just a few comments on the code readability. Nice work on the test cases 👍

cms/src/db/repositories/baseRepository.ts
ivanslabbert: TypeScript checks are failing if I remove the "!". Guess it is one of the joys of using TypeScript...

cms/src/types/dto.ts
ivanslabbert: Fixed, thanks

cms/src/db/repositories/baseRepository.ts
ivanslabbert: Rewritten to be a bit more readable.

PR #88, opened 2024-04-05
Group posts by category on homepage

ivanslabbert:


app/src/components/tags/HorizontalScrollableTagViewer.vue
```diff
@@ -0,0 +1,95 @@
...
+        showRightSpin.value = false;
+        return;
+    }
+
+    const scrollRight =
+        scrollElement.value.scrollWidth -
+        scrollElement.value.clientWidth -
+        scrollElement.value.scrollLeft;
+    const scrollLeft = scrollElement.value.scrollLeft;
+    console.log(scrollRight);
```
laurensgroeneveld, line 51: console.log left in.

app/src/components/tags/HorizontalScrollableTagViewer.vue
```diff
@@ -0,0 +1,95 @@
...
+    if (scrollRight > 20) showRightSpin.value = true;
+    if (scrollRight < 5) showRightSpin.value = false;
+};
+
+onMounted(() => {
+    // wait for child controls in the scroll element to be rendered
+    setTimeout(() => {
+        setSpinBtnVisibility();
+    }, 100);
+});
```
laurensgroeneveld, line 63: Instead, you could also add a one-time `watch` on `scrollElement`, which could be a bit more reliable.

app/src/components/tags/HorizontalScrollableTagViewer.vue
laurensgroeneveld, line 1: Can you add tests for this file? At least a simple test that it renders the tag title and the first post would be nice (scroller would be nice to test but not sure how easy that is).

laurensgroeneveld (commented):
Just some minor comments, it works nicely. Noticed one thing, when I add a new post that makes the row go over the "scroll limit", the new post appears immediately (because of the realtime socket.io connection), but the arrow doesn't, it only appears after a page refresh. Bit of a niche usecase, but maybe you can easily solve it by having a watcher on `tag` and rechecking button visibility every time it changes?

laurensgroeneveld (approved):

PR #102, opened 2024-04-11
Add sort and filter options to tag queries

ivanslabbert:


app/src/stores/post.ts
```diff
@@ -37,27 +59,53 @@ export const usePostStore = defineStore("post", () => {
...
-                    const sorted = res?.sort((a, b) => {
+        return (tagId: Uuid, queryOptions?: postQueryOptions) => {
+            // query
+            let res = posts.value?.filter((p) => {
+                return p.tags.some((t) => t._id == tagId);
+            });
+
+            // optional sorting
+            if (
+                res &&
```
laurensgroeneveld, line 70: Maybe just have an early return so you don't have to check if `res` exists for every `if`. Also this method now returns either an array or undefined, not sure if that's intended or if an empty array would make sense as a default.

app/src/stores/tag.ts
```diff
@@ -32,19 +61,105 @@ export const useTagStore = defineStore("category", () => {
...
 });
+
+function tagQuery(
+    tags: Readonly<Ref<Tag[] | undefined>>,
+    tagType: TagType,
+    queryOptions?: tagQueryOptions,
+    pinned?: "pinned" | "unpinned" | "any",
+) {
+    const postStore = usePostStore();
+    if (postStore) {
```
laurensgroeneveld, line 85: This if is not necessary, `usePostStore` will always return either an existing store or create it. It was needed before because the store was created at the top of the file, which is not the right place.

app/src/stores/tag.ts
```diff
@@ -32,19 +61,105 @@ export const useTagStore = defineStore("category", () => {
...
+        // query with optional filtering
+        let res = tags.value?.filter((t) => {
+            const posts = postStore.postsByTag(t._id, {});
+
+            let filter1 = true;
```
laurensgroeneveld, line 90: For readability it could be nice to give these `filter` variables more readable names.

laurensgroeneveld (approved):
Few small comments but looks good otherwise 👍

app/src/stores/tag.ts
ivanslabbert: Done intentionally as naming filters could be confusing (especially as they are set to true to be bypassed if not selected)

PR #105, opened 2024-04-12
Style videojs player for more suitable mobile experience

ivanslabbert:


app/src/components/posts/VideoPlayer.vue
```diff
@@ -12,6 +15,7 @@ const props = defineProps<Props>();
 const playerElement = ref<HTMLVideoElement>();
 
 onMounted(() => {
+    // TODO: Restrict startup bandwidth for live streaming
```
laurensgroeneveld, line 18: Should we rather create a ticket for this?

app/src/components/posts/VideoPlayer.vue
```diff
@@ -23,20 +27,115 @@ onMounted(() => {
...
+            remainingTimeDisplay: false,
+        },
     };
 
     const player = videojs(playerElement.value!, options);
 
     player.poster(props.contentParent.image);
     player.src({ type: "application/x-mpegURL", src: props.contentParent.content[0].video });
+
+    // @ts-ignore - videojs-mobile-ui
```
laurensgroeneveld, line 42: It can help to leave a little more readable comment here, like

```suggestion
    // @ts-ignore 2024-04-12 Typing of videojs-mobile-ui don't work
```

app/src/components/posts/VideoPlayer.vue
```diff
@@ -1,6 +1,9 @@
 <script setup lang="ts">
 import type { Post, Tag } from "@/types";
 import videojs from "video.js";
+import "videojs-mobile-ui";
+// TODO: fix typings
+import type mobileUi from "videojs-mobile-ui";
```
laurensgroeneveld, line 6: It could very well be that this doesn't work because the version of the typings is 0.8 while the package is 1.1, it seems the typings haven't been updated in a long time. We could either just remove it for now, or maybe commit a new change to the [types repo](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/videojs-mobile-ui).

laurensgroeneveld (approved):
Looks very nice 👍 As discussed I think displaying the maximum length of the video is very important, but it can be solved in a follow-up PR.

One small UI thing I noticed is that the live text is not so nicely aligned:
image

PR #112, opened 2024-04-15
Add audio-only mode to the videojs player

ivanslabbert:


app/src/components/posts/VideoPlayer.vue
```diff
@@ -51,127 +78,74 @@ onMounted(() => {
...
-            controls
-            preload="auto"
-            data-setup="{}"
-        ></video>
+    <div class="relative mb-2 rounded-lg bg-zinc-100 shadow-md dark:bg-zinc-800">
+        <div class="relative">
+            <img
+                v-if="audioMode"
+                :src="props.contentParent.image"
+                class="mb-2 aspect-video w-full rounded-lg object-cover"
```
laurensgroeneveld, line 117: Tiny UI improvement: I think it would look a little better if the bottom of the image isn't rounded for the audio player:
image
vs

image

Really like the styling of the player otherwise!

app/src/components/posts/VideoPlayerVideoMode.css
```diff
@@ -0,0 +1,110 @@
...
+    width: 100% !important;
+}
+
+.audio-mode-toggle-video-init {
+    @apply !absolute bottom-2 right-2;
+}
+
+.audio-mode-toggle-video {
+    @apply !absolute bottom-12 right-2;
+}
```
laurensgroeneveld, line 110: I would put all of these in the same file, otherwise it can be hard to find both the audio and video toggle classes in the future. Maybe put these in the `style` tag in the Vue component?

app/src/components/posts/VideoPlayer.vue
```diff
@@ -51,127 +78,74 @@ onMounted(() => {
...
+            />
+        </div>
+
+        <div class="" :class="{ 'video-mode': !audioMode, 'audio-mode': audioMode }">
```
laurensgroeneveld, line 121: Nitpick: empty `class` can be removed.

laurensgroeneveld (approved):
Very nice work, looks and works great 👌

PR #142, opened 2024-05-01
Added duplicate button on group

belljohan3:
@ivanslabbert i have added a duplicate button on groups.

### DIFF
--- a/cms/src/components/groups/GroupEditor.vue
+++ b/cms/src/components/groups/GroupEditor.vue
@@ -65,7 +65,7 @@ type Props = {
 const props = defineProps<Props>();
 
 const groupStore = useGroupStore();
-const { group: getGroup, updateGroup } = groupStore;
+const { group: getGroup, createGroup, updateGroup } = groupStore;
 const { groups } = storeToRefs(groupStore);
 const { addNotification } = useNotificationStore();
 const { isConnected } = storeToRefs(useSocketConnectionStore());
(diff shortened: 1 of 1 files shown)


cms/src/components/groups/GroupEditor.vue
```diff
@@ -263,6 +263,19 @@ const discardChanges = () => {
...
 };
 
+const duplicateGroup = async () => {
+    const duplicatedGroup = { ...toRaw(props.group) };
+    duplicatedGroup._id = "";
+    duplicatedGroup.name = `Copy of ${duplicatedGroup.name}`;
+    await createGroup(duplicatedGroup);
+
+    addNotification({
+        title: `Group "${duplicatedGroup.name}" duplicated successfully`,
```
laurensgroeneveld, line 273: I think this should either say the old name (`Group "Editors Private" duplicated successfully`) or say "created" instead of "duplicated" (`Group "Copy of Editors Private" created successfully`)

cms/src/components/groups/GroupEditor.vue
```diff
@@ -389,7 +402,13 @@ const saveChanges = async () => {
...
                         Offline changes
                     </LBadge>
-
+                    <LButton
+                        v-if="groups && groups.length > 0"
+                        variant="secondary"
+                        @click="duplicateGroup"
+                    >
+                        Duplicate
+                    </LButton>
```
laurensgroeneveld, line 411: This button feels a bit prominent right now, whereas I think it will not be the most important action on the page. It also clashes with the "save changes" button if there are changes:

image


I propose two changes:
- Only show the button when the group accordion is opened
- Instead of a text button, make it just an icon button, and make it tertiary, so it doesn't stand out as much.

laurensgroeneveld (changes requested):
Two small suggestions.

laurensgroeneveld:
Unrelated to your PR, but as we noticed when saving with a duplicated group it sometimes displays the group twice. Can you add this to the `saveChanges` method  just after `updateGroup` where we reset the other properties as well (line 340 for me):

```
    addedGroups.value = [];
```

cms/src/components/groups/GroupEditor.vue
```diff
@@ -403,12 +409,12 @@ const saveChanges = async () => {
                         Offline changes
                     </LBadge>
                     <LButton
-                        v-if="groups && groups.length > 0"
-                        variant="secondary"
+                        v-if="groups && groups.length > 0 && open"
+                        variant="tertiary"
+                        size="sm"
+                        :icon="DocumentDuplicateIcon"
```
laurensgroeneveld, line 415: Tiny suggestion: Add `title="Duplicate"` to this so that when you move your mouse over it it shows what the button will do.

laurensgroeneveld (approved):
Nice work 👍

PR #147, opened 2024-05-03
ACL duplicate button

belljohan3:


ivanslabbert:
Visual check: Functions as required.
Button placement: not horizontally centre in div (visible when hovering). Overall it needs to move a bit more right (same distance from right border as from top)? (It feels a bit off where it is now)

image

ivanslabbert:
@laurensgroeneveld comments?

ivanslabbert (approved):
Thanks Johan!

laurensgroeneveld:
> Visual check: Functions as required.
> Button placement: not horizontally centre in div (visible when hovering). Overall it needs to move a bit more right (same distance from right border as from top)? (It feels a bit off where it is now) - Edit: Might be fixed using the new "muted" button state as per https://github.com/bccsa/luminary/pull/144.

Just to come back to this: This is because there is currently an empty slot passed like this `<LButton [props]></LButton>`. It will be fixed by making the tag selfclosing, like `<LButton [props] />`

PR #149, opened 2024-05-03
Group copy ID button

belljohan3:
This is a temporary feature

### DIFF
--- a/cms/src/components/groups/GroupEditor.vue
+++ b/cms/src/components/groups/GroupEditor.vue
@@ -281,6 +281,23 @@ const duplicateGroup = async () => {
     });
 };
 
+const copyGroupId = (group: Group) => {
+    if (!group?._id) {
+        console.error("Group ID is missing or invalid.");
+        return;
+    }
+
+    const groupId = group._id;
+    navigator.clipboard
+        .writeText(groupId)
+        .then(() => {
+            console.log("Group ID copied to clipboard:", groupId);
+        })
+        .catch((error) => {
+            console.error("Failed to copy Group ID:", error);
+        });
+};
+
 const saveChanges = async () => {
     const updatedGroup = {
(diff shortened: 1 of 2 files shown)


cms/src/components/groups/GroupEditor.vue
```diff
@@ -281,6 +281,22 @@ const duplicateGroup = async () => {
     });
 };
 
+const copyGroupId = (group: Group) => {
+    if (!group?._id) {
```
laurensgroeneveld, line 285: I don't think this if is needed, because an id should always be present on the group.

cms/src/components/groups/GroupEditor.vue
```diff
@@ -281,6 +281,22 @@ const duplicateGroup = async () => {
...
+        console.error("Group ID is missing or invalid.");
+        return;
+    }
+
+    const groupId = group._id;
+    navigator.clipboard.writeText(groupId);
+
+    addNotification({
+        title: `Group ID Copied`,
+        description: "The ID's group has been copied to the clipboard",
```
laurensgroeneveld, line 295: "The ID of the group has been copied to the clipboard".

laurensgroeneveld (commented):
Looks good code wise just two small comments 👍 Can't test it from my phone so I'll leave that to Ivan.

cms/src/components/groups/GroupEditor.vue
belljohan3: Yes ! i have to remove it now.

cms/src/components/groups/GroupEditor.vue
belljohan3: OKay ! Thanks for the suggestion i am editing now.

ivanslabbert (approved):

PR #154, opened 2024-05-08
Add S3 storage provider

ivanslabbert:


api/README.md
```diff
@@ -2,6 +2,35 @@
...
+
+### S3 storage (MinIO)
+
+For development purposes, MinIO can be installed as a docker for S3 compatible storage:
+
+```shell
+docker run -d \
+   -p 9000:9000 \
+   -p 9001:9001 \
+   --name your-container-name \
```
laurensgroeneveld, line 28: Perhaps add a more proper name here so a new user doesn't have to edit this but can just copy and run it?

```suggestion
   --name luminary-storage \
```

api/README.md
```diff
@@ -2,6 +2,35 @@
...
+   -e "MINIO_ROOT_USER=rootuser" \
+   -e "MINIO_ROOT_PASSWORD=password" \
+   quay.io/minio/minio server /data --console-address ":9001"
+```
+
```
laurensgroeneveld, line 33: Are these all the instructions I need to run it? The `.env.example` gives `secretkey` as the default secret key but I'm not sure if that actually connects (since we don't use it in the code currently).

laurensgroeneveld (approved):
Looking good. Just two small notes on running MinIO locally but I guess we'll also run into that when we start using S3 in the code.

api/README.md
ivanslabbert: The example starts a MinIO instance which the developer can log in to and configure a secret key combination. But guess it is a good idea to have it "copy-paste"-able so will test and add a secret key combination to the readme

PR #156, opened 2024-05-08
Add a Group selection dropdown list

belljohan3:


cms/src/components/content/EditContentForm.vue
```diff
@@ -214,7 +232,12 @@ const hasNoPublishErrors = computed(() => {
         return hasParentImage.value;
     }
 
-    return hasOneContentField.value && hasParentImage.value && hasTag.value;
+    return (
+        hasOneContentField.value && hasParentImage.value && hasTag.value && hasNoDraftErrors.value
+    );
+});
+const hasNoDraftErrors = computed(() => {
+    return hasGroup.value;
```
laurensgroeneveld, line 240: I would suggest renaming `hasNoDraftErrors` to `canSaveAsDraft`, and then using that as well to disable the "save as draft" button (like `canPublish` does for publishing).

cms/src/components/content/GroupSelector.vue
```diff
@@ -0,0 +1,147 @@
...
+import {
+    Combobox,
+    ComboboxButton,
+    ComboboxInput,
+    ComboboxLabel,
+    ComboboxOption,
+    ComboboxOptions,
+} from "@headlessui/vue";
+import type { Group } from "@/types"; // Assuming you have a Group type defined
+import LTag from "./LTag.vue"; // Assuming you have a component for rendering groups
```
laurensgroeneveld, line 13: Not sure where these comments come from? Copilot? 😄

belljohan3: 😄😄

ivanslabbert:
@belljohan3 could you rebase this branch to main so that it does not fall to far behind? We'll get to implement it when done with #178

PR #157, opened 2024-05-10
API image processing

ivanslabbert:


api/src/changeRequests/validateChangeRequest.spec.ts
```diff
@@ -156,4 +158,50 @@ describe("validateChangeRequest", () => {
...
+                        preset: "default",
+                    },
+                ],
+            },
+        };
+
+        const result = await validateChangeRequest(changeRequest, ["group-super-admins"], db);
+
+        expect(result.validated).toBe(false);
+        expect(result.error).toBeDefined();
```
laurensgroeneveld, line 205: This test is called "validates a valid document" but the actual expect says that the validated is `false`.

laurensgroeneveld (approved):
Just one small comment so I'll approve. Looking forward to seeing this in action from the CMS 👍

api/src/changeRequests/validateChangeRequest.spec.ts
ivanslabbert: Thanks! You helped to uncover a bigger issue with the image validator which this test should have picked up...

PR #162, opened 2024-05-17
adding Expiry date on post page

belljohan3:


cms/src/components/content/EditContentForm.vue
```diff
@@ -350,6 +389,76 @@ const checkIfDirty = () => {
...
+                            class="ml-1"
+                            @click="() => calculateExpirationDate('1 Year')"
+                        >
+                            1Y
+                        </LButton>
+                        <LButton
+                            type="button"
+                            variant="secondary"
+                            :icon="BackspaceIcon"
+                            class="ml-1 pt-2"
```
laurensgroeneveld, line 458: The backspace button is not aligned with the other buttons:

image

Probably something to do with the `pt-2` here that is not set on the other items.

cms/src/components/content/EditContentForm.vue
```diff
@@ -350,6 +389,76 @@ const checkIfDirty = () => {
...
+                            @click="() => calculateExpirationDate('')"
```
laurensgroeneveld, line 459: Clicking this button doesn't actually reset the value, it just sets it to the current date (it also triggers the "unknown duration" warning from the `calculateExpirationDate` function). Instead I would expect it to totally clear the expiry date, which probably needs a different method like `clearExpirationDate`.

cms/src/components/content/EditContentForm.vue
```diff
@@ -350,6 +389,76 @@ const checkIfDirty = () => {
                     >
                         Only used for display, does not automatically publish at this date
                     </LInput>
+                    <LInput
```
laurensgroeneveld, line 392: Clearing the input using the browser "clear" button leads to an empty error in the interface:

image

I ran into the same problem with the publish date, with the problem being the browser. To fix it, copy the `transform` method in the `yup` declaration:
```ts
.transform((value, _, context) => {
                // Check to see if the previous transform already parsed the date
                if (context.isType(value)) {
                    return value;
                }

                // Default validation failed, clear the field
                // This happens when the 'clear' button in the browser datepicker is used,
                // which sets the value to an empty string
                return undefined;
            })
```

cms/src/components/content/EditContentForm.vue
```diff
@@ -150,11 +152,47 @@ onBeforeMount(() => {
...
+
+    setValues({ expiryDate: expirationDate.toISO()?.split(".")[0] as unknown as Date });
+};
+
 const save = async (validatedFormValues: typeof values, status: ContentStatus) => {
     // Make sure we don't accidentally add the 'parent' object to the content
     const contentValues = { ...validatedFormValues };
     delete contentValues["parent"];
     let publishDate;
+    let expiryDate;
```
laurensgroeneveld, line 195: You're declaring `expiryDate` here, but do nothing with it, so it doesn't actually save the expiry date 😉 (try saving with an expiry date and then reload the page). You need to add the same `if` as for the publish date (the `else` is not needed I think).

```ts
if (contentValues.expiryDate) {
        publishDate = DateTime.fromJSDate(contentValues.expiryDate);
}
```
(this is needed because we handle dates generally with the Luxon library (which is where `DateTime` comes from), but veeValidate/yup deal with JavaScript dates)

laurensgroeneveld (commented):
Having separate buttons is an elegant solution 👍 Just ran into some issues, see my comments.

cms/src/components/content/EditContentForm.vue
belljohan3: I have tried to set that class into the others items but it didn't work. I added the: inset-x-100 and bottom-0.2

cms/src/components/content/EditContentForm.vue
belljohan3: Yes, that was planned in my next commit. But thanks

belljohan3:
Hi @ivanslabbert. I will have a look at it tonight.

ivanslabbert:
> On narrower screens / browser windows the button wrapping looks a bit less nice: image
> 
> Perhaps give the buttons a fixed width, and check that the wrapping does not add an extra margin?
> 
> But while we are at it... Would it be nicer with something like this? image

New issue for this one: #165

ivanslabbert (approved):

PR #163, opened 2024-05-17
Feat/image-uploader

ivanslabbert:


ivanslabbert:
Still a draft, but feel free to look at it in the meantime @belljohan3

belljohan3:
I will have a look at it as soon as possible.

cms/src/types/cms.ts
```diff
@@ -64,9 +72,9 @@ export type Audio = Media & {
     downloads: MediaDownload[];
 };
 
-export type Image = Media & {
-    mediaType: MediaType.Image;
-};
+// export type Image = Media & {
```
belljohan3, line 75: Maybe we must delete comments to make it clean ?

api/.env.example
```diff
@@ -27,7 +27,11 @@ PERMISSION_MAP='{
 S3_ENDPOINT="your.s3.endpoint"
 S3_PORT=9000
 S3_USE_SSL=true
-S3_ACCESS_KEY="accesskey"
-S3_SECRET_KEY="secretkey"
-S3_IMG_BUCKET="bucketname"
+S3_ACCESS_KEY="minio"
```
belljohan3, line 30: I hope this is just for testing purpose otherwise it might be a faillure security access

ivanslabbert: Just placeholder info. But trying to make it work by default for testing and dev purposes

cms/src/components/images/ImageBrowser.vue
```diff
@@ -0,0 +1,71 @@
...
+import LCard from "../common/LCard.vue";
+import ImageEditor from "./ImageEditor.vue";
+import { ref } from "vue";
+
+// TODO: Implement select event
+
+const baseUrl: string = import.meta.env.VITE_CLIENT_IMAGES_URL;
+// TODO: Replace with a generic fallback image included in project source code
+const fallbackImg: string =
+    "https://cdn.activechristianity.org/image/upload/c_thumb,w_600,g_center/v1666007697/en_ke/f7bcwvet6axzmih97pv8.jpg";
```
laurensgroeneveld, line 16: Would be nice to solve this todo before we merge this 😄

cms/src/types/cms.ts
```diff
@@ -1,11 +1,12 @@
...
+import type { AclPermission, BaseDocumentDto, ImageFileDto, ImageUploadDto, Uuid } from "./dto";
 
 export enum DocType {
     Change = "change",
     Content = "content",
     Group = "group",
     Language = "language",
+    Image = "image",
     Media = "media",
     MediaDownload = "mediaDownload",
```
laurensgroeneveld, line 11: Why is `Image` a docType in it's own right now and not a part of `Media`/`MediaType`?

docs/socket.io-messages.md
```diff
@@ -1,10 +1,11 @@
...
+| clientDataReq    | Client data request                                                                                   | client / cms -> API   | Object with client details such as current update version |
+| data             | Document transfer                                                                                     | API -> client / cms   | Object with Array of document objects and sync version    |
+| changeRequest    | Document changes                                                                                      | client / cms -> API   | Array of change requests                                  |
+| changeRequestAck | Acknowledgement of change request                                                                     | API -> client / cms   | Acknowledgement of individual change request with status  |
+| accessMap        | Map with user specific access (depreciated)                                                           | API -> client / cms   | Object                                                    |
```
laurensgroeneveld, line 10: It seems the code related to the `accessMap` event is removed, so does it serve a purpose to keep it here, even with the "deprecated" note?

cms/src/components/images/ImageBrowser.vue
```diff
@@ -0,0 +1,71 @@
...
+    const id = db.uuid();
+    await db.upsert<ImageDto>({
+        _id: id,
+        type: DocType.Image,
+        name: "New Image",
+        description: "",
+        files: [],
+        memberOf: ["group-private-content"],
+        updatedTimeUtc: Date.now(),
+    });
```
laurensgroeneveld, line 30: What do you think about abstracting this in a repository class? Or do you think that's a needless extra layer? I'm just thinking a bit about that possibly in the future there is more logic related to for example the groups, and then it might be easier to have that in a separate class.

cms/src/components/images/ImageBrowser.vue
laurensgroeneveld, line 1: Missing a test for this file, it has some logic with the create action that would be nice to add unit tests for.

laurensgroeneveld (commented):
Some preliminary comments, didn't review the ImageEditor parts yet or test the UI, will get to that later today.

cms/src/components/images/ImageBrowser.vue
ivanslabbert: Yes, agree

cms/src/types/cms.ts
ivanslabbert: After thinking more about it I realised that image handling is quite different from other media types, so therefore I chose to keep it separate.

docs/socket.io-messages.md
ivanslabbert: I still kept it for backwards compatibility (as the app and CMS are still on seperate duplications of the code handling the access map). As soon as it is implemented throughout it can be removed.

cms/src/components/images/ImageBrowser.vue
ivanslabbert: I am trying to use generic classes (using T extends BaseDocumentDto) which should be reusable between different document types. The base toRef function is also exposed, so if there are some queries that will not be reused, the db query function can be passed directly.

The query function calls should mostly be one-liners (need to send the initial state along, that is why it looks a little big here), so I think it should not be necessary for seperate repository files. Will see how it goes when getting to groups etc. though...

cms/src/components/images/ImageBrowser.vue
ivanslabbert: Yes, as mentioned I still needed to add some tests

cms/src/components/images/ImageEditor.vue
laurensgroeneveld, line 1: When trying out the component in the browser, I get these console errors:

image

I didn't have MinIO running at the time, perhaps that is reason? But it seems a purely frontend error.

cms/src/components/images/ImageEditor.vue
```diff
@@ -0,0 +1,186 @@
...
+                name="fileName"
+                label="Image name"
+                class="mb-2 flex-1"
+                @change="save"
+                data-test="image-name"
+            />
+            <div class="mb-2 flex items-end gap-4">
+                <LSelect
+                    name="preset"
+                    label="Upload preset"
```
laurensgroeneveld, line 130: I find the "upload preset" field not super clear - I guess it does some specific optimizations related to the type? But not sure if editors will understand it, for example the difference between "photo", "picture" and "icon" (I also don't understand that 😅).

laurensgroeneveld (commented):
I can't test the component functionally at the moment - not entirely sure if I'm missing something in my local setup or if there is indeed an error somewhere 😅

cms/src/components/images/ImageEditor.vue
ivanslabbert: The preset is something that comes from the Sharp image processing library. Perhaps we should rename the "Upload preset" to something like "What type of image are you uploading?" and rename the options to:
- Photo
- Drawing
- Drawing with text
- Other

Photo will be the most common, and therefore is the default. (Options like "icon" can be removed as it is not really applicable here)

cms/src/components/images/ImageEditor.vue
ivanslabbert: Strange. Working on my side (tested starting with a clean database as well)

cms/src/components/images/ImageEditor.vue
laurensgroeneveld: I think that will help 👍 Still a bit torn whether it should be as prominent as it is. It can be tricky but maybe hide it behind an "advanced settings" or something. We can safely merge it but we should think a bit about this before onboarding people that we don't overload the UI with things that most of the time are not relevant (thus returning to the current AC admin situation).

ivanslabbert: Happy to remove the dropdown and leave it on default "photo". Effect will be that some images will be stored in less-than-ideal compression, but that will be very few as I see it at the moment.

laurensgroeneveld: > Happy to remove the dropdown and leave it on default "photo". Effect will be that some images will be stored in less-than-ideal compression, but that will be very few as I see it at the moment.

I think that would be nice 👍 If we need it we can bring it back in the future but think a bit more about the ergonomics for editors/less technical people, like having them explicitly select before uploading "is this a normal image or image with text".

laurensgroeneveld (approved):
This looks pretty good to me, just some smaller comments. As we discussed yesterday there are some improvements to be made for when we start using the image editor/browser in the post editor, but I think this can be merged already so we can continue to build on it (and not to make this PR linger more).

cms/src/components/images/ImageEditor.vue
```diff
@@ -0,0 +1,174 @@
...
+};
+
+const showFilePicker = () => {
+    // @ts-ignore
+    uploadInput.value!.showPicker();
+};
+
+// Read files into Buffer
+const upload = () => {
+    // @ts-ignore
```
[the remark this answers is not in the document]
ivanslabbert: Seems like the file upload input is not fully implemented in TypeScript for some or other reason. But will add some comments to the ts-ignore lines

cms/src/components/images/ImageBrowser.vue
```diff
@@ -0,0 +1,70 @@
...
+import { DocType, type ImageDto, type Uuid } from "@/types";
+import LImage from "./LImage.vue";
+import LButton from "../button/LButton.vue";
+import { PlusIcon } from "@heroicons/vue/20/solid";
+import LCard from "../common/LCard.vue";
+import ImageEditor from "./ImageEditor.vue";
+import { ref } from "vue";
+import fallbackImg from "../../assets/fallback-image-cms.webp";
+
+// TODO: Implement select event
```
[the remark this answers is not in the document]
ivanslabbert: Note to self here. Need to add a "selection mode" to select existing images from the content edit page

cms/src/db/baseDatabase.ts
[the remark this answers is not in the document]
ivanslabbert: Will add IndexedDB specific tests with #178

cms/src/components/images/ImageEditor.vue
```diff
@@ -0,0 +1,174 @@
...
+        const fileData = e.target!.result as ArrayBuffer;
+
+        if (fileData.byteLength > globalConfigStore.maxUploadFileSize) {
+            addNotification({
+                title: `Invalid image file size`,
+                description: `Image file size is larger than the maximum allowed size of ${maxUploadFileSize.value}MB`,
+                state: "error",
+            });
+            return;
+        }
```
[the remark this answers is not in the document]
ivanslabbert: Added new issue for this:
https://github.com/bccsa/luminary/issues/179

PR #166, opened 2024-05-22
enhancement of expiry shorcut buttons

belljohan3:


ivanslabbert:
@belljohan3 can you mark the selected button (first selected) so that the user will know his selection is active?

cms/src/components/content/EditContentForm.vue
```diff
@@ -760,3 +809,10 @@ const checkIfDirty = () => {
         <ConfirmBeforeLeavingModal :isDirty="isDirty" />
     </form>
 </template>
+
+<!-- <style scoped>
+.active {
+    background-color: #000000; /* or any other color to indicate selection */
+    color: white;
+}
+</style> -->
```
ivanslabbert, line 818: Do we need to keep this?

ivanslabbert (approved):
Thanks Johan, nice work. Just check my comment, and merge when ready.

cms/src/components/content/EditContentForm.vue
belljohan3: No ! I will delete it

PR #167, opened 2024-05-22
features for GroupSelection on post page

johan-bell:
These are the modifications codes of Group selection on post page coming from my fork !


laurensgroeneveld:
So should we close #156? Should I review this?

belljohan3:
Hi @laurensgroeneveld ! I think we should wait a bit. It is possible that we missed à modification I did in my fork and that is not present here. While investigate on it,  we will see. That's my point of view

ivanslabbert:
@belljohan3 could you rebase this branch to main so that it does not fall to far behind? We'll get to implement it when done with #178

PR #173, opened 2024-05-27
Filtering posts to show on the app

belljohan3:
Show posts that are not expired if the expired date is defined and posts that have the publish date greater than the actual date

### DIFF
--- a/app/src/stores/post.spec.ts
+++ b/app/src/stores/post.spec.ts
@@ -2,6 +2,7 @@ import { describe, it, expect, vi, afterEach, beforeEach } from "vitest";
 import { setActivePinia, createPinia } from "pinia";
 import { usePostStore } from "@/stores/post";
 import { mockCategory, mockPost } from "@/tests/mockData";
+import { DateTime } from "luxon";
 
 describe("Post store", () => {
     beforeEach(() => {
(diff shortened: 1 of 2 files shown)


app/src/stores/post.ts
```diff
@@ -42,7 +43,22 @@ export const usePostStore = defineStore("post", () => {
...
                 resolve(
                     posts.filter((post) => {
-                        return post.content[0]?.status == ContentStatus.Published;
+                        const isPublished = post.content[0]?.status == ContentStatus.Published;
+                        const publishDate = post.content[0]?.publishDate
+                            ? post.content[0]?.publishDate
+                            : null;
+                        const expiryDate = post.content[0]?.expiryDate
+                            ? post.content[0]?.expiryDate
+                            : null;
```
laurensgroeneveld, line 52: Please don't use `null`: we use `undefined` consistently throughout the app. JavaScript is a bit problematic in that both `null` and `undefined` are a thing, in most languages it's just `null`. By using only one we do not run into any cases where we check something like `date == null` but if it's `undefined` then that will return `false` which is unexpected.

app/src/stores/post.ts
```diff
@@ -42,7 +43,22 @@ export const usePostStore = defineStore("post", () => {
             return new Promise((resolve) => {
                 resolve(
                     posts.filter((post) => {
-                        return post.content[0]?.status == ContentStatus.Published;
+                        const isPublished = post.content[0]?.status == ContentStatus.Published;
```
laurensgroeneveld, line 46: I think there should still be an early return here, `if (!isPublished) return false` => because we only want to display published posts anyways so then we already know that this one should not be visible.

app/src/stores/post.ts
```diff
@@ -42,7 +43,22 @@ export const usePostStore = defineStore("post", () => {
...
+                        const expiryDate = post.content[0]?.expiryDate
+                            ? post.content[0]?.expiryDate
+                            : null;
+                        const isNotExpired =
+                            !post.content[0]?.expiryDate ||
+                            post.content[0]?.expiryDate > DateTime.now();
+
+                        const isAvailable =
+                            (publishDate ? publishDate <= DateTime.now() : true) &&
+                            (expiryDate ? expiryDate > DateTime.now() : true);
```
laurensgroeneveld, line 59: I notice two things here:
- `isAvailable` also checks for expiry date, which is double with `isNotExpired` above, so doesn't seem needed to me.
- This can be answered by Ivan, but not sure if it makes sense to check publish date at all? Or is that how we'll handle future published, by setting `status` to `published`, and having a publish date in the future?

laurensgroeneveld (commented):
Just a few things I noticed.

app/src/stores/post.ts
belljohan3: Hi @laurensgroeneveld ! Thanks for the comment.  So instead of `null` i have to `indefined` right ?

app/src/stores/post.ts
belljohan3: Yes @laurensgroeneveld ! He said that we can schedule posts in the CMS for the future. For that, i think we need to look at the both `publishDate` and `expiryDate`

app/src/stores/post.ts
belljohan3: Here is the return: `return isPublished && isNotExpired && isAvailable;`

laurensgroeneveld: Yes but I mean check for `isPublished` already at the top of the method, that is called an "early return", because you're not running any of the other code (potentially expensive operation). We can already exclude a lot of content by first checking if it's published at all (currently non-published posts get sent to the client as well, we might change that in the future).

ivanslabbert:
@belljohan3 could you add a few tests?
1. Check that posts with publish date < now are visible if no expiry date is set
2. Check that posts with publish date < now and expiry date > now are visible
3. Check that posts with publish date > now and expiry date > now are not visible
4. Check that posts with publish date < now and expiry date < now are not visible

belljohan3:
Hi @ivanslabbert ! Yes i can add it.

belljohan3:
Hello @laurensgroeneveld , @ivanslabbert 
I am facing to type-check issue problem.
It saids that publishDate can be also undefined
Do you have an idea on how to fix it ?

belljohan3:
Hi @laurensgroeneveld @ivanslabbert 
I am facing to type-check issue problem.
It saids that publishDate can be also undefined
Do you have an idea on how to fix it ?

app/src/stores/post.spec.ts
```diff
@@ -123,4 +124,150 @@ describe("Post store", () => {
...
+            _id: "2",
+            content: [{ ...mockPost.content[0], title: "post2", publishDate: now + 1000 }],
+        } as any;
+
+        postStore.posts = [post1, post2];
+
+        const visiblePosts = postStore.posts.filter((post) => {
+            const { publishDate, expiryDate } = post.content[0];
+            return publishDate < now && !expiryDate;
+        });
```
laurensgroeneveld, line 147: I don't think that `filter` should be here, otherwise you're not testing the logic in the store.

(so the code becomes obsolete, but if you would want to work around the `publishDate is possibly undefined` error you could first check that is not, like `return publishDate && publishDate < now && !expiryDate`)

belljohan3: OK ! What Logic should i test in store ? PostByTag ?

app/src/stores/post.spec.ts
```diff
@@ -228,25 +203,10 @@ describe("Post store", () => {
...
+        postStore.posts = [post1];
 
-        const visiblePosts = postStore.posts.filter((post) => {
-            const { publishDate, expiryDate } = post.content[0];
-            return publishDate > now && (!expiryDate || expiryDate > now);
-        });
+        const visiblePosts = postStore.postsByTag(mockCategory._id);
 
         expect(visiblePosts.length).toBe(1);
         expect(visiblePosts[0].content[0].publishDate).toBeGreaterThan(now);
```
ivanslabbert, line 212: The result of this test should be 0 (as the filter should have excluded the post). So this test is not testing a valid condition.

app/src/stores/post.spec.ts
```diff
@@ -263,30 +223,15 @@ describe("Post store", () => {
...
+        postStore.posts = [post1];
 
-        const visiblePosts = postStore.posts.filter((post) => {
-            const { publishDate, expiryDate } = post.content[0];
-            return publishDate < now && (!expiryDate || expiryDate < now);
-        });
+        const visiblePosts = postStore.postsByTag(mockCategory._id);
 
         expect(visiblePosts.length).toBe(1);
         expect(visiblePosts[0].content[0].publishDate).toBeLessThan(now);
```
ivanslabbert, line 237: The result of this test should be 0 (as the filter should have excluded the post). So this test is not testing a valid condition.

ivanslabbert:
The tests (as per above comments) are not testing the correct condition. But as discussed with @belljohan3 the filters do not work correctly in the testing environment but works correctly in runtime.

I suggest we fix the tests (expect result to be 0), but set them to skipped and fix it once we have fixed the underlying issue (part of planned refactoring).

ivanslabbert (changes requested):

ivanslabbert (approved):

PR #175, opened 2024-05-31
delete expired content from indexDB

belljohan3:


ivanslabbert:
Seems not to work yet, but probably as the API is resending the changes. The API needs to filter out expired content from its response to an App client (but keep it for CMS clients)

belljohan3:
Hi @ivanslabbert. As the PR #173 is already merged, I think now it can help with the process. I will have a look at it.

PR #178, opened 2024-06-01
Refactor client data model

ivanslabbert:
TODO:
- [x] Add tests as mentioned in #163 (IndexedDB specific tests)
- [x] Add tests for ContentOverview.vue
- [x] Convert Audioplaylists and Topics to the same way as Categories
- [x] Remove permission checks from generic routes (is not possible to know which doctype to check)
- [x] Update Create Post / Tag page
- [x] Implement sorting on content overview - move to separate issue
- [x] Cleanup old files
- [x] Implement local changes check
- [x] Extract shared code to separate folder / NPM library
- [x] Setup CI-CD for shared folder / NPM library - separate issue
- [x] Update CMS to use luminary-shared package for DB and shared types
- [x] Update app to use luminary-shared package for DB and shared types
- [x] Remove unused dependencies (moved to shared lib)
- [x] Include shared library in docker builds
- [x] Add query caching for heavy queries - move to separate issue
- [x] Filter on only published tags in tagsWhereTagType where the publish date is in the past and expiry date in the future
- [x] Shared: Write tests for permissions.ts (hasAnyPermission, getAccessibleGroups)
- [x] Shared: Write tests for database deleteRevoked function
- [x] Move all datetime transform operations to shared lib (remove luxon from cms and app) to ensure uniform implementation of dates and times - create separate issue

API:
- [x] Update seeding docs (ContentDto) with fields from post/tag parents

TagSelector.vue
- [x] Exclude tags in list to which the editor does not have assign access
- [x] Fix skipped test

ContentOverview.spec.ts
- [x] Fix tests
- [x] Debug post showing published content but not published in EditContent.vue

EditContent.vue
- [x] Add empty state (partially done)
- [x] Add loading state (not priority) - (added but commented)
- [x] Enable and fix tests for above in EditContent.spec.ts
- [x] Add guard with prompt when leaving page with unsaved changes
- [x] Validation not working when publish date is cleared and published toggle set
- [x] CMS bug: EditContent shows placeholder on first load for Posts/Tags with content documents

EditContentBasic.spec.ts
- [x] Add tests to test all the shortcut buttons for the expiry date
- [x] Add test to test Publish / Draft toggle switch

EditContentParent.spec.ts
- [x] Add test to test the tag pinned toggle
- [x] Add test to test the image input
- [x] Add test to see if Category, Tag and Audio playlist selected tags are displayed (Not completed yet)
- [x] Add test to see if added tags are added to underlying Parent document (PostDto or TagDto)

EditContentParentValidation.spec.ts
- [x] Check if save button is hidden when not dirty
- [x] Check if save button is showing when dirty
- [x] Check if save button is hidden on validation error
- [x] Add test for each parent validation (check if validation text is displayed)

EditContentValidation.spec.ts
- [x] Add test for each content validation
- [x] Check that correct language name is displayed

EditContentText.spec.ts
- [x] Check if the text from the mock content is displayed in the text editor

EditContentVideo.spec.ts
- [x] Check that the video URL is displayed in the text input
- [x] Check that the video URL can be updated from the text input

LanguageSelector.spec.ts
- [x] Add tests as per LanguageSelector.spec.ts
- [x] Remove database mocks (should not be needed as mock data can be passed directly to the component props)

TagSelector.vue
- [x] Add tests as per TagSelector.spec.ts

ContentRow.vue
- [x] fix TODO: Move this to ContentTable and pass ref to ContentRow - separate issue


cms/src/components/content/EditContentPreview.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+            },
+        });
+
+        // Find and update the title input field
+        const livePreview = wrapper.find('div[data-test="livePreview"]');
+
+        // Check if the content's title was updated
+        expect(content.value.status).toBe("published");
+        expect(livePreview.exists()).toBe(true);
+    });
```
ivanslabbert, line 37: It will be better to inject the publish status instead of relying on the mock data

belljohan3: Hi Ivan ! Which publishStatus do you mean ?

cms/src/components/content/EditContentText.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+        const content = ref<ContentDto>(mockEnglishContentDto);
+        const wrapper = mount(EditContentText, {
+            props: {
+                disabled: false,
+                content: content.value,
+            },
+        });
+
+        expect(wrapper.html()).not.toContain(content.value.text);
+    });
```
ivanslabbert, line 33: This test is not testing what the description says. mockEnglishContentDto has it's "text" property set, so the text area should show. The mock data should be modified so that the text field is removed, and then check if the div containing the text editor is not rendered.

cms/src/components/content/EditContentText.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+            props: {
+                disabled: true,
+                content: content.value,
+            },
+        });
+
+        const textContent = wrapper.find('div[data-test="textContent"]');
+        expect(textContent.exists()).toBe(true);
+
+        expect(textContent).toBeTruthy();
```
ivanslabbert, line 47: The disabled prop is used for preventing editing, not showing the div containing the editor.

The disabled prop should be tested with a separated test (or two)

cms/src/components/content/EditContentVideo.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+                disabled: true,
+                content: content.value,
+            },
+        });
+
+        const videoContent = wrapper.find('div[data-test="videoContent"]');
+        expect(videoContent.exists()).toBe(true);
+
+        expect(videoContent).toBeTruthy();
+    });
```
ivanslabbert, line 51: The disabled prop is used for preventing editing, not showing the div containing the editor.
The disabled prop should be tested with a separated test (or two)

The mock data does not contain a video URL, so this test is not testing what we expect it to test.

cms/src/components/content/EditContentVideo.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+            props: {
+                disabled: false,
+                content: content.value,
+            },
+        });
+
+        const videoContent = wrapper.find('div[data-test="videoContent"]');
+        expect(videoContent.exists()).toBe(true);
+
+        expect(wrapper.html()).not.toContain(videoContent);
```
ivanslabbert, line 35: We cannot do a comparison of wrapper html (text) and videoContent (object). This test will always pass.

cms/src/components/content/EditContentPreview.spec.ts
belljohan3: This one is solved now.

cms/src/components/content/ContentOverview.spec.ts
```diff
@@ -0,0 +1,176 @@
...
+                    return ref([mockEnglishContentDto]);
+                }),
+                isLocalChange: vi.fn(() => {
+                    return false;
+                }),
+                toDateTime: vi.fn((val) => {
+                    return DateTime.fromMillis(val);
+                }),
+            },
+        }));
```
ivanslabbert, line 44: Shall we remove the database mocks and rather test using mock date added to indexeddb? (Same as in EditContent.spec.ts)

docs/setup-vue-app.md
```diff
@@ -31,10 +31,33 @@ cp .env.example .env
...
+3. Build the shared library
+
+```sh
+cd ../shared
+npm run build
+```
+
+_Remember that this needs to be done every time changes are made to the shared library._
+
+4. Then start a live-reloading server with:
```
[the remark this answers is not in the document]
ivanslabbert: You first need to build the shared library:
```
cd shared
npm run build
```
Then install it along with other packages in the app and cms
```
cd app
npm ci --install-links
```

Might be the docs are not so clear - will have a look.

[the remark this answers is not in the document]
laurensgroeneveld: Ah yeah the docs had it the other way around 😅

PR #221, opened 2024-07-13
Add notification to Settings Page when local cache is cleared

belljohan3:


ivanslabbert (changes requested):
Hi @belljohan3 

I tested the PR, and it works correctly. But when the API is offline the button is disabled, while in the CMS we are keeping the button enabled and display an error notification when the button is pressed (see screenshot from CMS below). Could you change the functionality in the app to match that of the CMS?

image

belljohan3:
Hi @ivanslabbert. Yes I will have a look at it as soon as possible.

ivanslabbert (approved):

PR #227, opened 2024-07-15
APP: Profile menu language selection

belljohan3:
### What i have added in this PR.

- Icons on Profile Menu
- LanguageModal 
- Tests for LanguageModal 

Note that we need to read the language id from the LocalStorage and implment it the HomePage by taking into account the reactiveness.


app/src/components/navigation/TopBar.spec.ts
```diff
@@ -1,32 +1,41 @@
...
         (auth0 as any).useAuth0 = vi.fn().mockReturnValue({
             isAuthenticated: ref(false),
         });
 
         const wrapper = mount(TopBar);
 
-        expect(wrapper.html()).toContain("Log in");
-        expect(wrapper.findComponent(ProfileMenu).exists()).toBe(false);
+        expect(wrapper.html()).toContain("Profile");
     });
```
ivanslabbert, line 36: Description of this test is not valid anymore. Test could be combined with the next test, as the profile menu should now always show?

app/src/pages/HomePage.spec.ts
```diff
@@ -57,7 +57,7 @@ describe("HomePage", () => {
         vi.clearAllMocks();
     });
 
-    it("displays a message when there are no contents", async () => {
+    it.skip("displays a message when there are no contents", async () => {
```
ivanslabbert, line 60: Fix skipped tests

ivanslabbert (changes requested):
@belljohan3 left some comments. Also saw there are tests failing.

belljohan3:
The tests that fails actually i something that we need to look at it. I have tried it before but it is not working.

app/src/components/content/LanguageModal.vue
laurensgroeneveld, line 1: For modals in the CMS we use the [Headless UI modal](https://headlessui.com/v1/vue/dialog). It doesn't have any styling, so you can copy the styling you have, but it has several benefits:

1. Clicking outside the modal closes it
2. Pressing the Escape button closes it
3. Using the Tab key to cycle through the buttons (common for people that don't or can't use a mouse) will only cycle through the buttons of the modal.

So I think we should also use HeadlessUI here to ensure we have good accessibility.

app/src/components/content/LanguageModal.vue
```diff
@@ -0,0 +1,56 @@
...
+
+<template>
+    <div
+        v-if="isVisible"
+        class="fixed inset-0 z-50 flex items-center justify-center bg-gray-800 bg-opacity-75"
+    >
+        <div class="w-11/12 rounded-lg bg-white p-6 shadow-lg sm:w-96 dark:bg-zinc-900">
+            <h2 class="mb-4 text-lg font-semibold">Select Language</h2>
+            <ul class="divide-y divide-gray-200">
+                <li
```
laurensgroeneveld, line 30: Currently while pressing Tab is doesn't focus these buttons, which is what I would expect. This is most easiest solved by using an actual `button` element. Then the `ul` can be changed to a `div` for example.

app/src/components/content/LanguageModal.vue
```diff
@@ -0,0 +1,56 @@
...
+                        v-if="appLanguageIdAsRef === language._id"
+                        class="ml-auto h-6 w-6 text-yellow-500"
+                        aria-hidden="true"
+                    />
+                </li>
+            </ul>
+            <LButton
+                variant="primary"
+                size="lg"
+                :to="{ name: 'login' }"
```
laurensgroeneveld, line 47: This line is probably accidentally copied, it doesn't do anything (because the button is not a link), it can be removed.

app/src/components/navigation/ProfileMenu.spec.ts
```diff
@@ -14,35 +15,34 @@ vi.mock("vue-router", () => ({
...
     it("shows the user's name", async () => {
         (auth0 as any).useAuth0 = vi.fn().mockReturnValue({
+            isAuthenticated: ref(true),
             user: {
                 name: "Test Person",
             },
         });
 
         const wrapper = mount(ProfileMenu);
+        console.log(wrapper.text());
```
laurensgroeneveld, line 31: Left in a console.log.

app/src/components/navigation/ProfileMenu.vue
```diff
@@ -39,8 +78,15 @@ const userNavigation: {
...
                     class="ml-4 text-sm font-semibold leading-6 text-zinc-900 dark:text-zinc-50"
                     aria-hidden="true"
+                    v-if="isAuthenticated"
                     >{{ user?.name }}</span
                 >
+                <span
+                    class="ml-4 text-sm font-semibold leading-6 text-zinc-900 dark:text-zinc-50"
+                    aria-hidden="true"
+                    v-else
+                    >Profile</span
```
laurensgroeneveld, line 88: Just a small thing and can be easily changed later, but perhaps "Profile" is a bit strange if you're not logged in. Maybe it should just be `Menu`? Can quickly check at next standup.

app/src/components/navigation/ProfileMenu.vue
```diff
@@ -1,25 +1,64 @@
...
+    return [
+        {
+            name: "Settings",
+            icon: Cog6ToothIcon,
+            action: () => router.push({ name: "settings" }),
         },
-    },
-];
+        {
+            name: "Language",
```
laurensgroeneveld, line 29: I think a test for that it shows the modal when clicking the language button would be nice.

app/src/components/content/LanguageModal.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+        expect(wrapper.find("h2").exists()).toBe(false);
+    });
+
+    it.skip("emits close event on language click and stores the selected language", async () => {
+        const wrapper = mount(LanguageModal, {
+            props: { isVisible: true },
+        });
+
+        // The click event is not being triggered for some or other reason. Skipping test for now.
+        await wrapper.findAll("li").at(1)?.trigger("click");
```
laurensgroeneveld, line 52: I think this might be fixed when changing the `li` to a `button`, so be sure to test that (otherwise I can take a look, there's an issue for it on the board I think).

laurensgroeneveld (dismissed):
Works great 👍 Just a few comments on usability/accessibility and a few minor issues.

app/src/components/content/LanguageModal.vue
laurensgroeneveld: @belljohan3 This was resolved but I don't see any changes related to it, is that correct?

laurensgroeneveld:
I left a comment but see it was added to the commit, so just leaving a link to it here for visibility: https://github.com/bccsa/luminary/commit/8a8c662ac9e604fb33a6e507227ae9dd6147b4db#r144422881

belljohan3:
Hey @laurensgroeneveld !
Not sure why you don't see changes but i have pushed. 
Check this commit.
https://github.com/bccsa/luminary/pull/227/commits/91e11fad435757417572e05a1168ff4bc86ca3d2

ivanslabbert (changes requested):
The language modal background colour is different from the theme selector. Perhaps change the language modal to the same colour as the theme selector?

app/src/pages/HomePage.spec.ts
```diff
@@ -1,89 +1,145 @@
...
+                isAuthenticated: ref(false),
+            });
+            const wrapper = mount(HomePage);
 
-        expect(wrapper.text()).toContain(
-            "You don't have access to any content. If you believe this is an error, send your contact person a message.",
-        );
-    });
+            expect(wrapper.text()).toContain(
+                "There is currently no content available. If you have an account, first  log in.",
```
ivanslabbert, line 49: Change "If you have an account, first log in"
to
"Please log in if you have an account."

Sorry, made comment in test file. But please update in Vue component as well?

ivanslabbert (changes requested):

app/src/pages/HomePage.spec.ts
belljohan3: Ok

app/src/components/navigation/ProfileMenu.vue
```diff
@@ -61,21 +107,35 @@ const userNavigation: {
...
+                            aria-hidden="true"
+                        />
+                        <div class="flex flex-col gap-y-0">
+                            {{ item.name }}
+                            <span
+                                class="-mt-0 text-[12px] text-zinc-500 dark:text-white"
+                                v-if="item.language"
+                                >{{ item.language.name }}</span
+                            >
+                        </div>
```
ivanslabbert, line 132: image
The distance between the Settings and Theme icons are closer than between the other icons. This is due to the language being inserted underneath the "Language" menu item, but it makes the "Settings" menu item feel a bit too tight compared to the rest.

The menu item height could perhaps be made a bit bigger, so that it can fit the language name without expanding, and ensure that the spacing between the icons are visually equal?

app/src/components/navigation/ProfileMenu.vue
```diff
@@ -61,21 +107,35 @@ const userNavigation: {
             leave-to-class="transform opacity-0 scale-95"
         >
             <MenuItems
-                class="absolute right-0 z-10 mt-2.5 w-32 origin-top-right rounded-md bg-white py-2 shadow-lg ring-1 ring-zinc-900/5 focus:outline-none"
+                class="absolute right-0 z-10 mt-2.5 w-32 origin-top-right rounded-md bg-white py-2 shadow-lg ring-1 ring-zinc-900/5 focus:outline-none dark:bg-zinc-800"
             >
```
ivanslabbert, line 111: image

In dark mode, the background colour of the profile menu is the same as the page background. Perhaps it should be one shade lighter?

app/src/components/content/LanguageModal.vue
ivanslabbert: The Theme Selection modal also should be changed to HeadlessUI. Think we can just do it here in the same go, as the theme selection feature is already merged to main.

ivanslabbert (changes requested):

belljohan3:
NOTE: i have skipped test in LanguageModal and ThemeSelector as the modal is not rendering properly because we use HeadlessUI/Modal.

ivanslabbert (approved):
Approving but created a new issue to review the skipped tests
#245

PR #228, opened 2024-07-15
198 cms show language of unsaved changes

belljohan3:
- Add in validation area.
 - Unsaved changes to the Post / Tag (parent) document to show in General
 - Unsaved changes to ContentDto's (child documents) to show per ContentDto language
 - Rename "Validation errors" to "Validation"
 - Remove current "Unsaved Changes" label.
 - Use a yellow "info" icon for unsaved changes


ivanslabbert (changes requested):
@belljohan3 
When there are no validation errors, the unsaved changes does not show:
image

Would it look better if the unsaved changes looks the same as validation errors, but with a yellow "info" icon?
image

ivanslabbert:
Would also be nice to have some tests to check that the unsaved changes validation is showing as expected.

belljohan3:
image

This is what i did on validations

ivanslabbert (changes requested):
When making a change that triggers a validation error, the "unsaved changes" message also appears under the General section
image

ivanslabbert (changes requested):
Hi @belljohan3 
Can you move this text to be next to the save button?
telegram-cloud-photo-size-4-5836830808776753206-x

cms/src/components/content/EditContentParentValidation.vue
```diff
@@ -139,6 +157,7 @@ watch(
                         :languages="props.languages"
```
ivanslabbert, line 157: Add a sort function to languages prop like something like this?
`:languages="languages.sort(sortByName)"`

It is also not necessary to write "props.xxx" in the template section.

cms/src/components/content/EditContentParentValidation.vue
```diff
@@ -139,6 +157,7 @@ watch(
...
                         :key="content._id"
                         @isValid="(val) => setOverallValidation(content._id, val)"
+                        :contentPrev="props.contentPrev?.find((c) => c._id == content._id)"
```
ivanslabbert, line 160: Not needed to write `props.xxx`?

belljohan3:
Hey @ivanslabbert ! 
There is something that we should have a look at. 
Actually, the languages are sorted on EditContent for categories but not in the post. I tried what you said but it didn't work. 

I think we need to look a bit at it. Normally it should work everywhere as it is a Generic component.

ivanslabbert (changes requested):
I did a quick check, and it looks nice.
image

But when clicking on the badges next to the language labels, it routes with an incorrect router link (it appends the language code to the content document ID in the URL)
telegram-cloud-photo-size-4-5845722061094372585-x

Perhaps have a look on how the language switching was done in the language switcher?

ivanslabbert (approved):
Please just fix this? https://github.com/bccsa/luminary/pull/228/files#r1688018433

PR #234, opened 2024-07-19
add theme selection on app

belljohan3:


ivanslabbert (changes requested):
make the modal colour a bit lighter than the background?

ivanslabbert (approved):
The ThemeSelector files should perhaps be moved to the /app/src/components/navigation folder? It is currently in the "content" components folder, but the theme does not directly have anything to do with content.

PR #249, opened 2024-07-23
Hide the Publish date of pinned Categories

belljohan3:


app/src/components/tags/HorizontalScrollableTagViewer.vue
```diff
@@ -119,6 +122,7 @@ useResizeObserver(scrollContent, setSpinBtnVisibility);
                         v-for="content in taggedDocs"
                         :key="content._id"
                         :content="content"
+                        :show-publish-date="props.showPublishDate"
```
laurensgroeneveld, line 125: `props` is not necessary here, as all props are available by their name in the Vue template.

```suggestion
                        :show-publish-date="showPublishDate"
```

app/src/components/tags/HorizontalScrollableTagViewerCollection.vue
```diff
@@ -7,6 +7,7 @@ type Props = {
     tagType: TagType;
     tagQueryOptions: queryOptions;
     contentQueryOptions: queryOptions;
+    showPublishDate?: boolean;
```
laurensgroeneveld, line 10: I don't think this prop is needed as it's value is never used.

laurensgroeneveld (commented):
Just two small comments, but looks good otherwise, nice work on adding a test 👍

ivanslabbert (approved):
Approved, but please fix this before merging?
https://github.com/bccsa/luminary/pull/249/files#r1689564228

PR #250, opened 2024-07-23
Fix style for expiring date quick buttons

belljohan3:


cms/src/components/content/EditContentBasic.vue
laurensgroeneveld, line 1: There's a linter warning here now, just be sure to remove this import:

```
'ChevronLeftIcon' is defined but never used
```

belljohan3: This is now done.

cms/src/components/content/EditContentBasic.vue
```diff
@@ -307,10 +308,10 @@ const publishStatus = computed({
                     <LButton
                         type="button"
                         name="6"
-                        variant="custom"
+                        variant="secondary"
                         class="flex-1"
                         size="lg"
```
laurensgroeneveld, line 313: Please remove this one, as it's the only one on size `lg` and it causes the other ones to increase in size as well. The base size is fine I think.

```suggestion
```

laurensgroeneveld (approved):
Good fix and big improvement by using the secondary button I think 👍 We can look into other solutions like a button group in the future.

PR #251, opened 2024-07-23
add test to see if the route is generated correctly

belljohan3:


app/src/components/content/ContentTile.spec.ts
```diff
@@ -37,4 +47,32 @@ describe("ContentTile", () => {
...
+            global: {
+                plugins: [router],
+            },
+        });
+
+        const push = vi.spyOn(wrapper.vm.$router, "push");
+
+        await wrapper.trigger("click");
+
+        await router.push(`/${mockEnglishContentDto.slug}`);
```
ivanslabbert, line 72: It feels like we are simulating what the component itself should do here. The component should have called the route, and we must check if the router received the call, but now the test is also pushing the route...

ivanslabbert (changes requested):

app/src/components/content/ContentTile.spec.ts
laurensgroeneveld: To expand on this, it shouldn't be needed to create a router and routes using `createRouter`, as that gives very limited value in our tests, they're not testing the actual code. Instead, you can only mock the router `push`, and then test that. Something like (copied from `ProfileMenu.spec.ts`):
```ts
const routePushMock = vi.hoisted(() => vi.fn());
vi.mock("vue-router", () => ({
    useRouter: vi.fn().mockImplementation(() => ({
        push: routePushMock,
    })),
}));
```

and then the assertion would be something like

```ts
expect(routePushMock).toHaveBeenCalledWith({route: 'post'})
```

app/src/components/content/ContentTile.spec.ts
```diff
@@ -37,4 +50,30 @@ describe("ContentTile", () => {
...
+        });
+
+        await wrapper.trigger("click");
+        await router.isReady();
+
+        expect(routePushMock).toHaveBeenCalledWith({
+            name: "post",
+            params: { slug: mockEnglishContentDto.slug },
+        });
+    });
```
laurensgroeneveld, line 78: Nice improvement by testing the actual push from the component. Creating your own router is not needed here (and should indeed almost never be necessary, please think twice if you ever feel like you want to reach for it as it likely means the test is not testing the right thing), so you can remove the following parts:

```suggestion
    it("navigates to the correct route on click", async () => {
        const wrapper = mount(ContentTile, {
            props: {
                content: mockEnglishContentDto,
            },
        });

        await wrapper.trigger("click");

        expect(routePushMock).toHaveBeenCalledWith({
            name: "post",
            params: { slug: mockEnglishContentDto.slug },
        });
    });
```

With these changes the test is still green, as it will now just use the actual route definition of our app, and because we have mocked the `route.push` in the component we can test whether it is called with the right arguments.

laurensgroeneveld (dismissed):

ivanslabbert (approved):
Approved, but please remove unused imports in ContentTile.spec.ts before merging.

ivanslabbert (approved):

PR #252, opened 2024-07-23
Add Online/offline indicator on app

belljohan3:


app/src/components/OnlineIndicator.vue
```diff
@@ -0,0 +1,39 @@
...
+            leave-to-class="transform opacity-0 scale-95"
+        >
+            <PopoverPanel
+                class="absolute z-10 w-screen max-w-40 -translate-x-1/2 lg:left-0 lg:max-w-72 lg:translate-x-0"
+            >
+                <div
+                    class="mt-2 rounded-md bg-white px-4 py-3 text-xs shadow-lg ring-1 ring-inset ring-zinc-900/5"
+                >
+                    {{ isConnected ? "You are online." : "You are offline." }}
+                </div>
```
ivanslabbert, line 35: Missing dark mode here
image

I am however wondering whether the notification should perhaps be a banner instead of an icon that uses space in the top bar.
telegram-cloud-photo-size-4-5839471904426149424-x

This could perhaps be a reusable banner component that we can use to communicate different important information to the user.

@laurensgroeneveld what would you suggest?

app/src/components/navigation/TopBar.vue
```diff
@@ -2,36 +2,46 @@
...
-                    class="flex items-center gap-2 text-xl hover:text-zinc-600 dark:hover:text-zinc-200"
-                >
+                <div class="flex items-center text-xl hover:text-zinc-600 dark:hover:text-zinc-200">
                     <PlayCircleIcon class="h-8 w-8 text-yellow-500" />
-                    <div class="hidden items-center gap-1 min-[380px]:flex">
+                    <div class="hidden items-center md:flex">
+                        <span class="font-semibold">BCC Africa</span>
+                        <span>App</span>
+                    </div>
+                    <div class="flex items-center gap-1 md:hidden">
```
ivanslabbert, line 29: In the main branch the app name is hidden if a device has a small screen:
image

This is however not working anymore in this branch:
image

The spacing between the logo and app name text also seems to have changed. It looks like this on the main branch:
image

ivanslabbert (changes requested):

app/src/components/navigation/TopBar.vue
```diff
@@ -2,36 +2,46 @@
...
                         <span class="font-semibold">BCC Africa</span>
```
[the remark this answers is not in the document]
laurensgroeneveld: I think we did that to easily make a "logo" with parts in semibold font.

app/src/components/OnlineIndicator.vue
laurensgroeneveld: I think for the user the "online" status is not so interesting - usually they will be online, especially as we don't have a lot of offline content anyways (the videos can't be watched offline). In the CMS such an indicator is much more important I feel, as editors need to be aware if their content gets saved right away.

> This could perhaps be a reusable banner component that we can use to communicate different important information to the user.

This I feel would be a better solution. Just display a warning banner if the app is offline, and otherwise we don't show anything.

app/src/components/BarnerComponent.spec.ts
ivanslabbert, line 1: Check the name of this file. Barner should be Banner.

We could perhaps rename the component to NotificationBanner? Will that make more sense?

app/src/components/navigation/TopBar.vue
```diff
@@ -1,7 +1,8 @@
 <script setup lang="ts">
-import { PlayCircleIcon, ChevronLeftIcon } from "@heroicons/vue/24/solid";
+import { PlayCircleIcon, ChevronLeftIcon, SignalSlashIcon } from "@heroicons/vue/24/solid";
 import ProfileMenu from "./ProfileMenu.vue";
 import { useRoute, RouterLink } from "vue-router";
+import BarnerComponent from "../BannerComponent.vue";
```
ivanslabbert, line 5: Check naming...

ivanslabbert (changes requested):

app/src/components/BarnerComponent.spec.ts
laurensgroeneveld: Or `LBanner`? We use the `L` prefix for more generically named components. Including `Notification` in the name might be confusing with the popup notifications we have, which are named `LNotification`.

ivanslabbert:
telegram-cloud-photo-size-4-5854710349608109068-x
Please move the shadow to be below the top bar, and not below the banner. It now looks like the banner is part of the top bar, but it should look as if it is in the content section.

ivanslabbert:
image
The banner is too big for desktop mode. Also align the content to the left, and close button to the right without excessive margins.

app/src/components/notifications/NotificationManager.vue
```diff
@@ -1,15 +1,22 @@
+q
```
laurensgroeneveld, line 1: Accidental typo at the top of this file.

app/src/components/notifications/NotificationManager.vue
```diff
@@ -21,11 +28,24 @@ const { notifications } = storeToRefs(useNotificationStore());
...
+        v-if="type == 'banner'"
+    >
+        <div class="w-full">
+            <LNotification
+                v-for="notification in notifications.filter((c) => c.type == 'banner')"
+                :key="notification.id"
+                :notification
+            />
+        </div>
+    </div>
```
laurensgroeneveld, line 50: I think this should rather belong in it's own component. The thought behind the `NotificationManager` before was that you render it once. Now it can be a bit confusing, because it matters where you render it. A `NotificationManager` with `type=banner` should not be rendered at the end of the `body` tag, where a `toast` should. I think it would be clearer to split it in something like a `NotificationToastManager` and `NotificationBannerManager`.

app/src/components/notifications/NotificationManager.vue
```diff
@@ -21,11 +28,24 @@ const { notifications } = storeToRefs(useNotificationStore());
...
         </div>
     </div>
+    <div
+        aria-live="assertive"
+        class="inset-x-0 flex items-end sm:items-start"
+        v-if="type == 'banner'"
+    >
+        <div class="w-full">
+            <LNotification
+                v-for="notification in notifications.filter((c) => c.type == 'banner')"
```
laurensgroeneveld, line 45: Nitpick: Especially if we split up the `NotificationManager` in two components (see my other comment), I would perhaps opt to make this a getter in the store, so you can do something like this:

```ts
const { banners } = storeToRefs(useNotificationStore());

v-for="banner in banners"
```

app/src/App.vue
```diff
@@ -47,16 +49,42 @@ onBeforeMount(async () => {
...
+setTimeout(() => {
+    watch(
+        isConnected,
+        () => {
+            if (!isConnected.value) {
+                useNotificationStore().addNotification({
+                    id: "offlineBanner",
+                    title: "You are offline",
+                    description:
+                        "Do not worry - you can still use the app and browse through the offline content.",
```
laurensgroeneveld, line 63: Suggestion to make it a bit clearer what type of content could be available:

```suggestion
                        "You can still use the app and browse through offline content, but some content (like videos) might not be available.",
```

app/src/stores/notification.ts
```diff
@@ -25,10 +36,18 @@ export const useNotificationStore = defineStore("notification", () => {
             });
         }, 100);
 
-        setTimeout(() => {
-            notifications.value = notifications.value.filter((n) => n.id != notificationId);
-        }, 2000);
+        if (notification.type == "toast") {
+            setTimeout(() => {
+                removeNotification(notificationId!);
+            }, 4000);
```
laurensgroeneveld, line 42: 4 seconds feels a bit long for me, just want to check if it was a conscious decision to double it?

app/src/components/notifications/LNotification.vue
```diff
@@ -1,35 +1,77 @@
...
+            icon.value = ExclamationCircleIcon;
+            break;
+        case "warning":
+            icon.value = ExclamationCircleIcon;
+            break;
+    }
+}
+
+const colour = ref<string>("bg-gray-100");
+const textColor = ref<string>("text-gray-400");
```
laurensgroeneveld, line 39: Nitpick: Here we have `colour` (British English) and `Color` (American English). Let's use only one spelling, everywhere else it's American, at least in the case of `color`.

app/src/components/notifications/LNotification.vue
```diff
@@ -41,8 +83,11 @@ const show = ref(true);
                 <div class="ml-4 flex flex-shrink-0">
                     <button
                         type="button"
-                        @click="show = false"
+                        @click="
+                            notification.id ? removeNotification(notification.id) : (show = false)
```
laurensgroeneveld, line 87: A notification always has an id, so this will always remove the notification. The reason why this only toggles `show` was that the `NotificationManager` is responsible for deleting the actual notification from the store after a while anyway, which I think could be fine to keep, what do you think?

app/src/components/notifications/LNotification.vue
```diff
@@ -51,4 +96,27 @@ const show = ref(true);
...
+                <component :is="icon" class="h-5 min-h-5 w-5 min-w-5" />
+                <div class="flex flex-col md:inline-block md:align-middle">
+                    <span class="text-md md:text-sm">{{ notification.title }}</span>
+                    <span v-if="notification.description" class="text-xs md:ml-3">{{
+                        notification.description
+                    }}</span>
+                </div>
+            </div>
+            <XMarkIcon
+                @click="notification.id ? removeNotification(notification.id) : (show = false)"
```
laurensgroeneveld, line 117: Similar to the comment above, this will always call `removeNotification`. I am wondering if that is right here as well. The banner is still valid, but the user is hiding it. Especially in the case of the offline banner, should the user be able to click it away? It serves as a nice warning to them why they can't play videos for example, which could 
be confusing otherwise. And most people are conditioned to click away banners without reading them.

So I would propose to remove the close button for the banner, or at least make it optional.

app/src/components/notifications/LNotification.vue
```diff
@@ -1,35 +1,77 @@
...
+            break;
+        case "error":
+            icon.value = ExclamationCircleIcon;
+            break;
+        case "info":
+            icon.value = ExclamationCircleIcon;
+            break;
+        case "warning":
+            icon.value = ExclamationCircleIcon;
+            break;
```
laurensgroeneveld, line 34: This could also be written as:

```suggestion
        case "error":
        case "info":
        case "warning":
            icon.value = ExclamationCircleIcon;
            break;
```

app/src/components/notifications/LNotification.vue
```diff
@@ -51,4 +96,27 @@ const show = ref(true);
...
+                        notification.description
+                    }}</span>
+                </div>
+            </div>
+            <XMarkIcon
+                @click="notification.id ? removeNotification(notification.id) : (show = false)"
+                class="h-6 min-h-6 w-6 min-w-6 cursor-pointer underline md:h-5 md:min-h-5 md:w-5 md:min-w-5"
+            />
+        </div>
+    </div>
```
laurensgroeneveld, line 121: Again here I think this belongs in its own component. The only code that's shared here is the `icon`, and that could be extracted to a different place - perhaps the store?

I suggest to split this in `LToast` and `LBanner` components, or just `NotificationToast` and `NotificationBanner`.

app/src/components/notifications/LNotification.vue
laurensgroeneveld: Also I think it's confusing to have `colour` which is the background color for the banner, and `textColor` which is only used for the toast. Part of the reason why I propose to split this component into two, see my other comment.

app/src/App.vue
```diff
@@ -47,16 +49,42 @@ onBeforeMount(async () => {
...
+            }
+        },
+        { immediate: true },
+    );
+}, 1000);
 </script>
 
 <template>
     <TopBar class="sticky top-0 z-40" />
+    <NotificationManager type="banner" class="sticky" />
```
laurensgroeneveld, line 81: This generates a warning in the console:

```
[Vue warn]: Extraneous non-props attributes (class) were passed to component but could not be automatically inherited because component renders fragment or text root nodes. 
```

Basically, you shouldn't pass the class here, but put it on the `div` in the component itself. An extra reason to split up the NotificationManager in two.

app/src/App.vue
```diff
@@ -47,16 +49,42 @@ onBeforeMount(async () => {
         Sentry.captureException(err);
     }
 });
+
+// Wait 1 second to allow the socket connection to be established before checking the connection status
+setTimeout(() => {
```
laurensgroeneveld, line 54: I always dislike artificial timeouts like this. While it's a working solution, and I'm not necessarily advocating a different approach right now, it does trigger the thought in me if we can do something different. To me there's a few downsides with this approach:
- Someone might close the page or navigate away. This is the `App` component which will stay alive throughout the app, but in other components that means the `setTimeout` keeps floating around (that's why you generally want to do a [clearTimeout](https://developer.mozilla.org/en-US/docs/Web/API/setTimeout#return_value) in a Vue component's `onBeforeDestroy`).
- Someone's connection might be slower, meaning that in this case they are perhaps not offline at all

One thing I'm thinking of is that we should track if the socket has tried to connect, and base the display of this banner on that (`hasTriedToConnect && !isConnected`). In my console when I disable the API I see a lot of `Failed to load resource: net::ERR_CONNECTION_REFUSED`, so the browser does know if you did at least one request or not.

Just paging @ivanslabbert to get his thoughts on this.

app/src/App.vue
```diff
@@ -47,16 +49,42 @@ onBeforeMount(async () => {
...
+    watch(
+        isConnected,
+        () => {
+            if (!isConnected.value) {
+                useNotificationStore().addNotification({
+                    id: "offlineBanner",
```
laurensgroeneveld, line 60: While this is a fine solution, another solution could be to return the `id` from `addNotification` and save that in a variable. Then it's a little easier to reference any notification instead of having to come up with unique names.

app/src/App.vue
```diff
@@ -47,16 +49,42 @@ onBeforeMount(async () => {
...
     }
 });
+
+// Wait 1 second to allow the socket connection to be established before checking the connection status
+setTimeout(() => {
+    watch(
+        isConnected,
+        () => {
+            if (!isConnected.value) {
+                useNotificationStore().addNotification({
```
laurensgroeneveld, line 59: I'm missing a test for `App` for this, I think that would be good to have.

app/src/components/notifications/LNotification.spec.ts
```diff
@@ -14,11 +32,19 @@ describe("LNotification", () => {
...
+                    title: "Important News",
+                    description: "Read this.",
+                    state: "success",
+                    type: "toast",
+                },
+            },
         });
 
-        await wrapper.find("button").trigger("click");
+        await wrapper.find("button[data-test='toast']").trigger("click");
```
laurensgroeneveld, line 45: Nitpick: make this button name a bit more unique, like `toast-close-button`.

laurensgroeneveld (changes requested):
It looks good visually, I have quite a few comments mostly related to splitting up the components a bit to make them clearer.

app/src/stores/notification.ts
laurensgroeneveld: @belljohan3 If it was a conscious decision to do this it's fine, but please place a comment instead of just resolving, otherwise I don't know what the status is here 😅

app/src/App.spec.ts
```diff
@@ -30,4 +39,46 @@ describe("App", () => {
...
+
+        mount(App, {
+            shallow: true,
+        });
+
+        // Wait for the setTimeout and watch to trigger
+        await waitForExpect(() => {
+            expect(removeNotification).toHaveBeenCalledWith("offlineBanner");
+        }, 1000);
+    });
```
laurensgroeneveld, line 83: Thanks for adding tests! Just a small note, there's no need to mock the store here I think. Generally stores work really well in unit tests. You can just directly use it and do assertions on it.

```ts
        const notificationStore = useNotificationStore();
        expect(notificationStore.addNotification).toHaveBeenCalledWith(
            expect.objectContaining({ state: "error" }),
        );
```

I copied this from a test in the CMS, `SettingsPage.spec.ts`. Also notice the `expect.objectContaining` here by the way, this is a very handy method that you can use to only check part of an object, instead of having to check for all fields.

laurensgroeneveld (commented):
There's a few open comments from my previous review, please check them and either change the relevant code or leave a comment.

ivanslabbert (approved):

PR #253, opened 2024-07-24
CMS: Assign groups to posts / tags

ivanslabbert:
#156


ivanslabbert (commented):
Added task to fix skipped test where click event does not seem to trigger in GroupSelector.spec.ts
#254

ivanslabbert (commented):
Also see #167 which is replaced by this PR

cms/src/components/content/EditContentParent.vue
```diff
@@ -56,16 +57,16 @@ const canEdit = computed(() => {
...
         <LInput
             name="parent.image"
-            label="Default image"
+            label="Image"
             :icon="LinkIcon"
             placeholder="https://..."
             :disabled="!canEdit"
             v-model="parent.image"
-        >
-            This image can be overridden in a translation
```
laurensgroeneveld, line 65: Just for my understanding, this was just removed for now because we don't have images yet on the content, but when we have that this will come back, right? Makes sense to remove it now, it's confusing when the other functionality is not there.

cms/src/components/groups/GroupSelector.spec.ts
```diff
@@ -0,0 +1,173 @@
+import "fake-indexeddb/auto";
+import { describe, it, expect, beforeEach, afterEach, vi, beforeAll, afterAll } from "vitest";
+import { mount } from "@vue/test-utils";
+import {
+    mockGroupDtoPrivateContent,
+    mockGroupDtoPublicContent,
+    mockGroupDtoPublicEditors,
+    mockGroupDtoPublicUsers,
+    mockGroupDtoSuperAdmins,
```
laurensgroeneveld, line 9: Some linting warnings in this file. Chec https://github.com/bccsa/luminary/pull/253/files

cms/src/components/groups/GroupSelector.vue
laurensgroeneveld, line 1: Since this is very similar to the TagSelector we should probably extract a generic Combobox component. But not necessarily in this PR, can be a future task (perhaps when we need it a third time 😛)

cms/src/components/groups/GroupSelector.vue
```diff
@@ -0,0 +1,158 @@
...
+
+type Props = {
+    disabled?: boolean;
+};
+withDefaults(defineProps<Props>(), {
+    disabled: false,
+});
+const groups = defineModel<Uuid[]>("groups");
+
+const availableGroups = db.whereTypeAsRef<GroupDto[]>(DocType.Group, []);
```
laurensgroeneveld, line 30: It's nice seeing this new database layer in action, feels pretty clean 👍

cms/src/components/groups/GroupSelector.spec.ts
```diff
@@ -0,0 +1,173 @@
...
+import GroupSelector from "./GroupSelector.vue";
+import { Combobox } from "@headlessui/vue";
+import LTag from "../content/LTag.vue";
+import { accessMap, AclPermission, db } from "luminary-shared";
+import waitForExpect from "wait-for-expect";
+import { access } from "fs";
+
+describe("GroupSelector", () => {
+    // Need to set the access map before starting the tests. When moving this to beforeAll, it fails for some or other reason.
+    accessMap.value = superAdminAccessMap;
```
laurensgroeneveld, line 22: Probably because you're modifying the result of an import. Not necessarily wrong in tests, but I suspect the change is only locally scoped in the `beforeAll`, and therefore not propagating to any other methods, while doing it here changes it for the whole scope, including the test methods.

cms/src/components/groups/GroupSelector.spec.ts
```diff
@@ -0,0 +1,173 @@
...
+            mockGroupDtoPublicContent,
+            mockGroupDtoPublicEditors,
+            mockGroupDtoPublicUsers,
+            mockGroupDtoPrivateContent,
+        ]);
+    });
+
+    beforeEach(async () => {
+        // setActivePinia(createTestingPinia());
+    });
```
laurensgroeneveld, line 35: Unused code.

cms/src/components/groups/GroupSelector.spec.ts
```diff
@@ -0,0 +1,173 @@
...
+        await wrapper.find("input").setValue("edit");
+
+        await waitForExpect(() => {
+            expect(wrapper.text()).toContain("Public Editors");
+            expect(wrapper.text()).not.toContain("Public Content");
+            expect(wrapper.text()).not.toContain("Super Admins");
+        });
+    });
+
+    it.skip("Updates the passed array when selecting a group", async () => {
```
laurensgroeneveld, line 114: Please add a note with a link to the relevant issue why this test is skipped so we don't lose track of it.

laurensgroeneveld (approved):
Looks good 👍 Just left a few of my thoughts.

cms/src/components/content/EditContentParent.vue
ivanslabbert: Correct

cms/src/components/groups/GroupSelector.spec.ts
ivanslabbert: Issue #254 added with reference to this PR

Issue #255, opened 2024-07-24
Shared: Reactive permission updates


ivanslabbert:
Change shared permissions verifyAccess and hasAnyPermissions functions to reactive (computed) functions (or add reactive versions of the functions).

Should be implemented in the CMS to prevent users from making changes when permissions has been revoked, or to live-add new permissions

PR #260, opened 2024-07-26
Group editor permissions validation

ivanslabbert:
TODO:

- [ ] Create CMS tests on disabled edit capabilities - separate task (#262)
- [x] Visually indicate which groups a user can edit or not
- [x] Add a description explaining that ACL's are giving access to the users who are members of the assigned groups to perform the permission on documents who are members of the given group.
- [x] Prevent users from saving a group to which he/she will not have edit access
- [x] Check if the way group document permissions are applied makes sense (applied to child docs or to the group itself). Applying it to the group itself feels like the more correct way, but check in the APi how it is currently implemented.
- [x] Remove "Create" permission and allow group membership to be set on content docs if the CMS user has "Edit" access to the document type. The "Assign" permission should then be sufficient to control who can create groups, as a user needs to have "Assign" access to be able to save a group with an ACL.
- [ ] Apply downstream permissions to parent groups. - separate issue (#261)
- [x] Update the GroupSelector component to show groups to which the user has Edit access for the given DocType
- [x] EditContent: Disable saving to published content if user does not have publish access (validation section?) - separate issue
- [x] Add note on disabled groups that the user will know that he does not have permissions to edit the group
- [x] Make deselected permission icon one shade lighter in disabled groups
- [x] When creating a new group, no-edit-access verification is not working. -> This is working, but as the accessmap is not updating reactively yet, a page reload is needed to enable editing of the newly created group. (See #255)


shared/src/tests/mockdata.ts
```diff
@@ -207,22 +207,19 @@ export const fullAccessToAllContentMap = {
     "group-private-content": {
         post: {
             view: true,
-            create: true,
```
laurensgroeneveld, line 210: Remember to also remove this from all the seeding docs in the API (search for `"create"`)

cms/src/components/groups/EditGroup.vue
```diff
@@ -310,6 +394,13 @@ const saveChanges = async () => {
                 leave-to-class="transform scale-95 opacity-0"
             >
                 <DisclosurePanel class="space-y-6 px-6 pb-10 pt-2">
+                    <p v-if="!disabled">
+                        Configure permissions to this group ({{ editable.name }}) and it's members:
+                    </p>
+                    <p v-else>
+                        No access to edit permissions to this group ({{ editable.name }}) and it's
+                        members.
+                    </p>
```
laurensgroeneveld, line 403: Typo: it's => its

cms/src/components/groups/EditGroup.vue
```diff
@@ -280,17 +346,35 @@ const saveChanges = async () => {
...
                             Save changes
                         </LButton>
                     </div>
 
                     <LBadge v-if="isDirty && !open">Unsaved changes</LBadge>
+                    <LBadge v-if="isEmpty" variant="warning">
+                        The group does not have any access configured</LBadge
+                    >
+                    <LBadge v-if="!hasEditPermission && !isEmpty" variant="warning">
+                        The group will not be editable after saving</LBadge
```
laurensgroeneveld, line 364: This reads to me as if I can still save it. Instead of "will", "would" might fit better. And maybe put the actual error (group can't be saved) in front. Should it maybe be an error badge?

```suggestion
                        Can't save: The group would not be editable</LBadge
```

cms/src/pages/internal/ComponentSandbox.vue
laurensgroeneveld, line 1: Unused variable linter warning in this file.

laurensgroeneveld (approved):
Some small comments, but looks good overall.

PR #267, opened 2024-07-30
Upstream permission inheritance

ivanslabbert:
Expand permissions of upstream groups with permissions applied to lower level groups


api/src/permissions/permissions.service.spec.ts
```diff
@@ -171,7 +172,152 @@ describe("PermissionService", () => {
...
+            expect(
+                PermissionSystem.verifyAccess(
+                    ["group-test-low"],
+                    DocType.Language,
+                    AclPermission.Edit,
+                    ["group-test-top"],
+                ),
+            ).toBe(true);
+
+            // Remove the test groups for further tests
```
laurensgroeneveld, line 312: This feels a bit error-prone, is there no way to reset the state / create a new permission system instance for every test?

laurensgroeneveld (approved):
It's a lot of code but based on the tests it seems to do what it needs to do. Nicely added some documentation as well 👍

PR #272, opened 2024-08-01
add automatic language content switching

belljohan3:


ivanslabbert (changes requested):

app/src/pages/SingleContent.vue
```diff
@@ -22,9 +23,30 @@ const content = db.getBySlugAsRef<ContentDto>(props.slug);
...
+        } else {
+            if (appLanguageAsRef.value?._id != content.value.language) {
+                const contentDocs = await db.whereParent(content.value.parentId);
+                const preferred = contentDocs.find(
+                    (c) => c.language == appLanguageAsRef.value?._id,
+                );
+                if (preferred) {
+                    await router.push({ name: "content", params: { slug: preferred.slug } });
+                    content.value = preferred;
+                } else {
```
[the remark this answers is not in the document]
belljohan3: The ```else``` here is required. Otherwise it will show the notification even if the content is not found.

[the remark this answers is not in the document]
laurensgroeneveld: What Ivan means is that you can place an empty `return;` statement inside the `if`. Then the `else` is not needed. I also think that is a nicer way of doing it.

app/src/pages/SingleContent.vue
```diff
@@ -22,9 +23,28 @@ const content = db.getBySlugAsRef<ContentDto>(props.slug);
...
 watch(
-    content,
+    [content, appLanguageAsRef],
     async () => {
-        if (!content.value) return;
+        if (!content.value) {
+            return;
+        }
+
+        if (appLanguageAsRef.value?._id != content.value.language) {
```
laurensgroeneveld, line 32: I don't think this should be in the same watcher as the one for `content`. If I am in an English article that doesn't have a Swahili translation, and change my language to Swahili, I get the notification. But when I reload the page, I get the notification twice, because this watcher triggers everytime either `content` or `appLanguageAsRef` changes. So I think at the very least this should be in it's own watcher.

laurensgroeneveld: Although this issue is also something else: the notification will now always show, even if I'm not switching language. That could be the intended behaviour, just want to double check that we do this deliberately.

app/src/pages/SingleContent.vue
```diff
@@ -22,9 +23,28 @@ const content = db.getBySlugAsRef<ContentDto>(props.slug);
...
+            const contentDocs = await db.whereParent(content.value.parentId);
+            const preferred = contentDocs.find((c) => c.language == appLanguageAsRef.value?._id);
+
+            if (preferred) {
+                await router.push({ name: "content", params: { slug: preferred.slug } });
```
laurensgroeneveld, line 37: Instead of `router.push`, `router.replace` might be a good option as well. It will replace the current URL, so when you click back in the browser, you don't go back to the "wrong" language, but instead to the page before that, which might be the homepage.

app/src/pages/SingleContent.vue
```diff
@@ -22,9 +23,28 @@ const content = db.getBySlugAsRef<ContentDto>(props.slug);
...
+                content.value = preferred;
```
laurensgroeneveld, line 38: This is needed because routing to the same route will not remount the component but keep using the same instance. For clarity, I would suggest to place this one line up, above the `router` call.

app/src/pages/SingleContent.vue
```diff
@@ -22,9 +23,28 @@ const content = db.getBySlugAsRef<ContentDto>(props.slug);
...
+            } else {
+                useNotificationStore().addNotification({
```
laurensgroeneveld, line 40: Can we get a test for this?

laurensgroeneveld (changes requested):
Works nicely, just have a few minor comments 👍

belljohan3:
I am struggling to write test for automatic language content switching. @laurensgroeneveld & @ivanslabbert

ivanslabbert:
> I am struggling to write test for automatic language content switching. @laurensgroeneveld & @ivanslabbert

Perhaps write a test where you open an article in French (with the preferred language set to English), and wait to see if the English article title is in wrapper.text()?

app/src/pages/SingleContent.spec.ts
```diff
@@ -126,4 +150,31 @@ describe("SinglePost", () => {
...
+            },
+        });
+
+        // Wait for the initial content to load
+        await waitForExpect(() => {
+            expect(wrapper.html()).toContain(mockEnglishContentDto.summary);
+        });
+
+        // Simulate language change to French
+        mockAppLanguageAsRef.value = mockLanguageDtoFra;
```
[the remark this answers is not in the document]
laurensgroeneveld: Ivan says, check how the language switcher changes it.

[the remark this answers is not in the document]
laurensgroeneveld: In `LanguageModel` it sets `appLanguageIdAsRef` which is not readonly, as opposed to `appLanguageAsRef`.

app/src/pages/SingleContent.vue
```diff
@@ -114,7 +144,7 @@ onMounted(async () => {
 
         <div
             class="mt-6 border-t border-zinc-200 pt-6 dark:border-zinc-500"
-            v-if="content.tags.length > 0"
+            v-if="content.tags.length != 0"
```
laurensgroeneveld, line 147: Why did you change this from `>` to `!=`? It seems clearer to me if it's `>` - I would never expect it to be less than zero.

app/src/pages/SingleContent.vue
```diff
@@ -34,6 +38,32 @@ watch(
...
+        if (appLanguageAsRef.value?._id != content.value.language) {
+            const contentDocs = await db.whereParent(content.value.parentId);
+            const preferred = contentDocs.find((c) => c.language == appLanguageAsRef.value?._id);
+
+            if (preferred) {
+                content.value = preferred;
+                await router.replace({ name: "content", params: { slug: preferred.slug } });
+                return;
+            }
+            useNotificationStore().addNotification({
```
laurensgroeneveld, line 57: I would still like a test for this notification. Did you look at that? Just test switching to a language for which there is no translation and then test if `addNotification` was not called:

```ts
 expect(notificationStore.addNotification).not.toHaveBeenCalled()
```

app/src/pages/SingleContent.vue
belljohan3: @laurensgroeneveld I change like this because, in the test is not showing the tags when there is/are.

laurensgroeneveld: @belljohan3 I think then we should investigate why that happens. Logically thinking:

- There are no tags: the length is 0
- There are tags: the length is greater than 0

So checking for if it's greater than 0 should always be sufficient, because the length can't be less than 0.

The only reason why I can imagine it's causing problems in the test is that `content.tags` itself is not an array but `undefined`. That means the `.length` will cause an error to occur. But then we should rather fix the test and not change the production code. 

Making the tests work is not a goal of itself: the tests are there to check that the code is working, so the code should generally not be changed to make the tests pass.

app/src/pages/SingleContent.vue
laurensgroeneveld: @belljohan3 I see I didn't mention it in this comment but a test that the notification _was_ called is also good I think.

laurensgroeneveld: This is still missing, we only test that when the language does change the notification also is displayed, please also add a test that when the language doesn't change because it isn't available the notification is displayed.

laurensgroeneveld (approved):

PR #275, opened 2024-08-04
Fix "No content message flashes briefly when reloading home page"

DirkDevelopmentFreelance:
Add code to add a delay when mounting the Homepage to the DOM, so the Data has enough time to load

### DIFF
--- a/app/src/pages/HomePage.vue
+++ b/app/src/pages/HomePage.vue
@@ -12,7 +12,8 @@ console.log(isAuthenticated.value);
 
 const hasPosts = db.someByTypeAsRef(DocType.Post);
 
-const waitForDataRenderIsOver = waitForDataRender(4.5);
+const noContentMessageDelay = ref(false);
+setTimeout(() => { noContentMessageDelay.value = true }, 1000);
 </script>
 
 <template>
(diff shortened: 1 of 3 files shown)


app/src/pages/HomePage.vue
```diff
@@ -5,10 +5,14 @@ import IgnorePagePadding from "@/components/IgnorePagePadding.vue";
...
 import { DocType, TagType, db } from "luminary-shared";
 import { appLanguageIdAsRef } from "@/globalConfig";
+import { waitForDataRender } from "@/util/waitForDataRender";
 
 const { isAuthenticated } = useAuth0();
+console.log(isAuthenticated.value);
 
 const hasPosts = db.someByTypeAsRef(DocType.Post);
+
+const waitForDataRenderIsOver = waitForDataRender(4.5);
```
ivanslabbert, line 15: Some comments:
The name is not really describing what we are waiting for? (It takes a moment for the data to be loaded from the database before homepage components can render the data.) Perhaps a better name would be```noContentMessageDelay```?

On slow devices it can take some time for the query to load, but 4.5 seconds is perhaps a bit much. You also want to show something before the user thinks the app is broken. Perhaps 1 second could work?

Implementing the delay as a separate (imported) function looks nice and clean, but as this is such a small thing, could it rather have been done with e.g. a ref and a timer setting the value? Perhaps something like this?
```suggestion
const noContentMessageDelay = ref(false);
setTimeout(() => { noContentMessageDelay.value = true }, 1000);
```

DirkDevelopmentFreelance:
Tested updated code, there is a slight bug where the timeout works and it is delayed, but there is no delay in rendering the content. Working on a fix with updated code as reference.

PR #276, opened 2024-08-05
Show error when opening the expired or schedule doc

belljohan3:


ivanslabbert (dismissed):

app/src/pages/SingleContent.vue
```diff
@@ -21,10 +18,20 @@ const props = defineProps<Props>();
 const content = db.getBySlugAsRef<ContentDto>(props.slug);
 const tagsContent = ref<ContentDto[]>([]);
 
+const isExpiredOrScheduled = computed(() => {
+    if (!content.value) return false;
+    return content.value.publishDate! > Date.now() || content.value.expiryDate! < Date.now();
```
laurensgroeneveld, line 23: I was triggered by the use of `!` here - generally it's good to be super careful when using it in production code (I care a lot less about it in test code). Are you certain there is always a `publishDate` AND `expiryDate` here? Both are optional. I can guess that a publish date is usually available for posts in the app, but expiry dates will probably not be present for the vast majority of posts. And even a publish date is not guaranteed, perhaps we add some way to view draft posts in the future?

So I think we need to do extra checks here (`content.value.publishDate && content.value.publishDate > Date...`) and remove the `!`

app/src/pages/NotFoundPage.vue
```diff
@@ -6,11 +6,13 @@ import { RouterLink } from "vue-router";
     <div class="grid min-h-full place-items-center px-6 py-24 sm:py-32 lg:px-8">
         <div class="text-center">
             <p class="text-base font-semibold text-yellow-600">404</p>
-            <h1 class="mt-4 text-3xl font-bold tracking-tight text-zinc-900 sm:text-5xl">
-                Page not found
+            <h1
+                class="mt-4 text-3xl font-bold tracking-tight text-zinc-900 dark:text-white sm:text-5xl"
```
laurensgroeneveld, line 10: Nice that you fixed dark mode for this page 👍

laurensgroeneveld (changes requested):
Just have one important note on the use of `!` in the TS code, otherwise looks good and works fine 👍

app/src/pages/SingleContent.vue
```diff
@@ -21,10 +18,25 @@ const props = defineProps<Props>();
...
 watch(
     content,
     async () => {
         if (!content.value) return;
+
+        document.title = isExpiredOrScheduled.value
+            ? `Page not found - ${appName}`
+            : `${content.value.title} - ${appName}`;
+
+        if (isExpiredOrScheduled.value) return;
```
laurensgroeneveld, line 38: I think this would be clearer with an early return:

if (isExpired)
document.title = Page not found
return;

and then the rest of the code continues

app/src/pages/SingleContent.spec.ts
```diff
@@ -7,8 +7,25 @@ import SingleContent from "./SingleContent.vue";
...
-
-vi.mock("vue-router");
+import NotFoundPage from "./NotFoundPage.vue";
+
+const routePushMock = vi.hoisted(() => vi.fn());
+// vi.mock("vue-router", () => ({
+//     useRouter: vi.fn().mockImplementation(() => ({
+//         push: routePushMock,
+//     })),
+// }));
```
laurensgroeneveld, line 17: Some commented code here.

laurensgroeneveld (approved):
Looks good, just make sure to remove the commented code in the test.

PR #279, opened 2024-08-06
Add toggle to link dates on content documents

DirkDevelopmentFreelance:
Added linkDates as an optional data field, and a toggle in the CMS that updates the value of the linkDates value
@belljohan3 Just make sure it doesn't conflict with your changes and cause rendering issues, will make this PR available for review if everything checks out


DirkDevelopmentFreelance:
Issues with processChangeRequest in API, tests failed

PR #246, opened 2024-07-23
CMS: Edit page visual update

belljohan3:
Notes from today's standup meeting:

- Draft / Published:
>- Should not be toggle of make it a text toggle (text left and right of toggle). Add colour to make it stand out a bit more.(Done)

- State indications (also add new status badges to overview - separate issue) - (Done)
>- Published 
>- Draft
>- Expired
>- Scheduled

- Validation section
>- Change "validation error" text to "There are some errors that prevent saving" - (Done)
>- Permanently show all languages (used by post / tag) in validation section - (Partially done)
>- Add status badge for each language next to the language name. Status badge to have colour, icon and status text - (Done)
>- See if the badge component can be improved to have an icon, colour and text - (Done)
>- Make language header row clickable, and show "selected" tick-mark icon. - (Done)
>- Keep the language dropdown for now - will remove later

Make the publish/draft toogle a list of radio
move the tick-mark button out the language card 
Set the Published badge to orange color when  unsaved changes, publish with an excalamation icon

PR #280, opened 2024-08-06
add link dates toggle

DirkDevelopmentFreelance:
add link dates toggle and add "linkDates" as an optional field for PostDTo.  Discuss API tests that fail?


api/src/dto/PostDto.ts
```diff
@@ -16,4 +16,8 @@ export class PostDto extends _contentBaseDto {
     @IsString({ each: true })
     @Expose()
     tags: Uuid[];
+
+    @IsBoolean()
+    @Expose()
+    linkDates: boolean;
```
ivanslabbert, line 22:
```suggestion
@IsOptional()
    linkDates?: boolean;
```

ivanslabbert (changes requested):

cms/src/components/content/EditContentParent.spec.ts
```diff
@@ -27,10 +27,28 @@ describe("EditContentParent.vue", () => {
...
+    it("test the link date toggle", async () => {
+        const parent = ref<PostDto>({ ...mockData.mockPostDto, linkDates: true });
+        const wrapper = mount(EditContentParent, {
+            props: {
+                docType: DocType.Post,
+                modelValue: parent.value,
+            },
+        });
+
+        // // Check if the LToggle component is rendered
```
laurensgroeneveld, line 46: Nitpick: I see you copied the comment from above, but in my opinion it's superfluous and we can do without it as the code makes it clear enough what's happening.

cms/src/components/content/EditContentParent.spec.ts
```diff
@@ -27,10 +27,28 @@ describe("EditContentParent.vue", () => {
...
-        const toggle = wrapper.findComponent({ name: "LToggle" });
+        const toggle = wrapper.find('[data-testId="pinned-LToggle"]');
         expect(toggle.exists()).toBe(true);
 
-        expect(toggle.props("modelValue")).toBe(true);
+        const props = toggle.getCurrentComponent()?.props;
+        expect(props!["modelValue"]).toBe(true);
+    });
+
+    it("test the link date toggle", async () => {
```
laurensgroeneveld, line 37: I see the other tests in this file have similar names, but I think they can be written a bit more clearly. Generally I would recommend writing it in a sentence form (it already starts with "it", so write whatever comes after that: "it does X when I click Y" for example). And when this test fails, a clear description of what this test does helps to view at a glance what might be wrong. My suggestion for this specific case would be something like "it renders the link dates toggle with the correct initial state". 

Do you agree that makes it clearer? In that case I suggest to also rewrite the other test names in this file.

On a separate note, we could perhaps add a test or assertion in this same test that we can toggle it, and that it updates the right value on the parent.

laurensgroeneveld (commented):
I just have two small comments, but also a more general question. This code looks fine, but I don't have a good insight in what problem is solved here. I'm guessing it will link publish date & expiry date across translations?

On its own, this will not do anything, which in general I'm not a huge fan of: in other places we've removed UI elements for features that we plan to do but are not there yet, as it can lead to confusion.

I'm also wondering whether just a "link dates" toggle is clear enough for less technical users. Should there be some visual indication on the dates that they are linked? Should the toggle be on by default? Note that we're also touching the date fields in #246 so it might not be a good idea to do that in this particular PR.

ivanslabbert:
After discussion on 2024/08/12, we decided that there are complications around a linked-date functionality, especially w.r.t. translation permissions (should a user be able to change the dates of content to which he does not have translate permissions).

We are putting this on ice for now.

PR #283, opened 2024-08-09
282 cms group editor additional tests

belljohan3:


cms/src/components/groups/EditAclByGroup.spec.ts
```diff
@@ -149,4 +149,45 @@ describe("EditAclByGroup.vue", () => {
...
+                    mockGroupDtoPublicEditors,
+                    mockGroupDtoPublicUsers,
+                    mockGroupDtoSuperAdmins,
+                ],
+                disabled: true,
+            },
+        });
+
+        expect(wrapper.text()).toContain(mockGroupDtoPublicContent._id);
+        expect(wrapper.html()).not.toContain('div[data-test="duplicateAcl"]');
```
ivanslabbert, line 170: Perhaps add a comment stating that you are checking that the group is not editable by checking that the duplicate button is not visible

cms/src/components/groups/EditAclByGroup.spec.ts
```diff
@@ -149,4 +149,45 @@ describe("EditAclByGroup.vue", () => {
...
+                    mockGroupDtoPublicContent,
+                    mockGroupDtoPublicEditors,
+                    mockGroupDtoPublicUsers,
+                    mockGroupDtoSuperAdmins,
+                ],
+                disabled: false,
+            },
+        });
+
+        expect(wrapper.html()).not.toContain('div[data-test="duplicateAcl"]');
```
ivanslabbert, line 191: Perhaps add a comment stating that you are checking that the group is not editable by checking that the duplicate button is not visible

cms/src/components/groups/EditAclByGroup.vue
```diff
@@ -51,7 +51,7 @@ const duplicateGroup = (targetGroup: GroupDto) => {
                 <div class="flex items-center justify-between">
                     <div></div>
                     <div class="py-1">
-                        {{ assignedGroup.name }}
+                        {{ disabled ? assignedGroup._id : assignedGroup.name }}
```
ivanslabbert, line 54: I do not think this is correct. Groups may be disabled for editing while the group doc is available (meaning we have the group name)

ivanslabbert (changes requested):
Left a few comments

ivanslabbert (approved):

Issue #288, opened 2024-08-12
CMS: Add filtering/searching/sorting to content overview


ivanslabbert:
For UI inspiration, look at [Polaris](https://polaris.shopify.com/patterns/resource-index-layout), although their specific implementation is probably to intricate for our needs.

Some highlights:
- Separate sort button
- Filters can be individually removed
- Search can be combined with filters
- All filters can be removed with one button

![Image](https://github.com/user-attachments/assets/bfed6a2f-074d-49cc-953b-68a2d781ca4d)

ivanslabbert:
![Image](https://github.com/user-attachments/assets/92c173c3-a596-4aab-aca5-e13730534763)

Issue #287, opened 2024-08-12
CMS: Add sorting to content overview


ivanslabbert:

laurensgroeneveld:
Combined with #288

PR #290, opened 2024-08-12
Searchable, sortable and filterable content overview query

ivanslabbert:
- [x] Add tests for query
- [ ] Implement in UI in separate PR (see #287 #288 #289 )


cms/src/components/content/query.ts
```diff
@@ -0,0 +1,131 @@
...
+    if (!options.pageIndex) options.pageIndex = 0;
+    if (!options.translationStatus) options.translationStatus = "all";
+    if (!options.publishStatus) options.publishStatus = "all";
+
+    const translated = (await db.docs // This may slow down the query if there are many documents, but it is necessary to be able to include and filter on untranslated documents
+        .where({ type: DocType.Content, language: options.languageId })
+        .toArray()) as ContentDto[];
+    const untranslatedByParentId: Uuid[] = [];
+
+    let res = db.docs.orderBy(options.orderBy);
```
laurensgroeneveld, line 30: Shouldn't this have a `where` for only Content docs? Now we're loading everything.

I also guess this is where pagination will come into play? As this can load a very large dataset.

cms/src/components/content/query.ts
```diff
@@ -0,0 +1,131 @@
...
+    if (options.orderDirection == "desc") res = res.reverse();
+
+    return res
+        .filter((doc) => {
+            const contentDoc = doc as ContentDto;
+
+            // Filter documents by type
+            if (!contentDoc.parentId) return false;
+            if (contentDoc.type != DocType.Content) return false;
```
laurensgroeneveld, line 39: Related to my previous comment, why not include this in the query?

cms/src/components/content/ContentOverview.vue
```diff
@@ -32,13 +29,28 @@ const tagType = Object.entries(TagType).some((t) => t[1] == props.tagType)
...
+    orderBy: "updatedTimeUtc",
+    orderDirection: "desc",
+    pageSize: 20,
+    pageIndex: 0,
+    tags: [],
+    search: "",
+    publishStatus: "all",
+});
+
+const queryKey = computed(() => JSON.stringify(queryOptions.value));
```
laurensgroeneveld, line 52: Can't we move this to the `ContentTable` as well? It seems that the only reason why the query options are declared here is that we have the language selector in this component, and now it feels divided between this component and the table. We could pass the language along as a prop instead. (in any case I think the language selection should happen in the user profile, not here on the page).

laurensgroeneveld (changes requested):
Looks really good 🔥 Just have a few smaller comments, and one more important one on the naming of the new `tagType` on the content.

api/src/changeRequests/processChangeRequest.ts
```diff
@@ -46,6 +46,10 @@ export async function processChangeRequest(
             doc.memberOf = parentDoc.memberOf;
             doc.tags = parentDoc.tags;
             doc.image = parentDoc.image;
+
+            if (parentDoc.type == DocType.Tag) {
+                doc.tagType = (parentDoc as TagDto).tagType;
```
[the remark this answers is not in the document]
ivanslabbert: I kept the name the same as the parent document as these are replicated fields from the parent (together with other fields - see [add github permalink to "processChangeRequest.ts line 40]

If we should change the naming for better readability / clarity, we should do it for all the replicated fields.

cms/src/components/content/query.ts
ivanslabbert: It is not possible to use a "where" query together with "orderBy". If we would want to sort the data at a later stage (with sortBy), it would mean that we would need to load all the filtered data into memory, defeating the purpose of pagination.
I believe with pagination the performance will not be too bad, and the "filter" logic is filtering out non-ContentDto documents as early as possible to avoid unnecessary processing.

cms/src/components/content/ContentOverview.vue
ivanslabbert: My thought was that the query and filtering controls can be added to `ContentOverview.vue`, and not to `ContentTable.vue`. That keeps `ContentTable` a bit cleaner, and sorting / filtering is applied through the `queryOptions`.

But this is debateable, so welcome to challenge my thinking (again).

api/src/changeRequests/processChangeRequest.ts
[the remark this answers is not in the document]
laurensgroeneveld: Ah, right. I noticed `parentType` and `parentId`, which are both prefixed. It makes a little bit more sense for replicated fields.

> If we should change the naming for better readability / clarity, we should do it for all the replicated fields.

Maybe that would make sense? I feel it could be confusing, especially for cases where we have similar names, like now we replicate `image` to the content, which could clash in naming with a localised image.

cms/src/components/content/ContentOverview.vue
laurensgroeneveld: Hmm still a bit torn on this one. Intuitively I feel like filtering and sorting is an integral part of the table, and should thus be part of that component too (that is how Polaris does it as well). But if we can foresee that we need that info outside of the table then this approach makes sense. The way it currently is it just feels like it can all be packed into one component.

ivanslabbert: I am a bit afraid creating one "super component", but perhaps we can create a separate filter component and include it in the table component to keep it a bit separated.

cms/src/components/content/query.ts
ivanslabbert: The where includes ```{ type: DocType.Content, language: options.languageId }```, but I suppose I might have added it after your comment.

laurensgroeneveld: That is the other query for `translated` it seems?

api/src/changeRequests/processChangeRequest.ts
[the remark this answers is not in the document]
ivanslabbert: Updated replicated fields to be prefixed with "parent", with the exception of the memberOf field as this is needed to query data from the database for given group ID's for most document types.

cms/src/components/content/ContentOverview.vue
ivanslabbert: Tried to move it, but it makes refreshing after changing query variables difficult, so reverted back to having the query options in `ContentOverview.vue`

laurensgroeneveld (approved):

PR #291, opened 2024-08-13
add deleteExpired documents function/make cms-flag a global variable

DirkDevelopmentFreelance:
### NOTES:
 **_### TODO -->_**
- [Revert seedingDocs to original states.] 
- Fix CMS tests failing. Maybe something to do with database imports being incorrect?


shared/src/db/database.ts
```diff
@@ -83,6 +85,8 @@ class database extends Dexie {
             },
             { immediate: true },
         );
+
+        this.deleteExpired();
```
laurensgroeneveld, line 89: As discussed, calling this in the constructor is fine, but I think it should be an interval, so that if someone has the website open for a while their expired documents get deleted as well.

Suggestion: Make `isCms` a property of the `database` class and set it in the constructor of `socket`: `db.isCms = ...`.

Then use that to check in your interval whether it's CMS or not.

shared/src/db/database.ts
```diff
@@ -626,6 +630,51 @@ class database extends Dexie {
...
+        if (cmsFlag.cms) {
+            console.log("CMS is on");
+            return;
+        } else {
+            console.log("CMS is not on...deleting documents");
+            const now = DateTime.now().toMillis();
+
+            this.docs
+                .where("type")
+                .anyOf(DocType.Content, DocType.Post)
```
laurensgroeneveld, line 645: Only content can expire, posts don't have an expiry date (we'll have to see how we handle that, maybe also delete  posts without content?)

shared/src/db/database.ts
```diff
@@ -626,6 +630,51 @@ class database extends Dexie {
...
+                            const contentDoc = doc as ContentDto;
+                            return contentDoc.expiryDate && contentDoc.expiryDate <= now;
+                        } else if (doc.type === DocType.Post) {
+                            const postDoc = doc as PostDto;
+                            if (postDoc.metadata && postDoc.metadata.length > 0) {
+                                for (const metadata of postDoc.metadata) {
+                                    if (metadata.expiryDate && metadata.expiryDate <= now) {
+                                        return true; // Expired document found
+                                    }
+                                }
```
laurensgroeneveld, line 659: I _think_ there is no `metadata` field on the post, because the expiry dates are per content, not globally for the whole post.

shared/src/db/database.ts
laurensgroeneveld: That does mean `isCms` should be `true` by default I think because otherwise in the first run it will delete documents if you're in the CMS - so perhaps this way is also not the best, but we can look into refactoring it if it works.

shared/src/socket/socketio.ts
```diff
@@ -27,16 +41,17 @@ class Socketio {
     private socket: Socket;
     private retryTimeout: number = 0;
     private localChanges = db.getLocalChangesAsRef();
-    private isCms: boolean;
+    private isCMS = false;
```
laurensgroeneveld, line 44: For consistency, variables should be all lowercase, except the first letter of each distinct word (camelCase). So `isCms` is the "correct" way of spelling "is CMS".

laurensgroeneveld (commented):
Some comments that we discussed during the standup.

shared/src/config.ts
DirkDevelopmentFreelance, line 1: I tried without using a class singleton, and other solutions as well, but _**all**_ of them (including this solution) have the issue where the cms flag given in the "CMS" project or app, is not the one that the database gets, the database sticks with the default value in every scenario, I tried making it a vue-ref as well. The only solution that worked for this PR so far regarding the cms-flag being properly updated inside the database file was just by using a **timer** before checking the cms-flag in the `deleteExpired` function.

My personal opinion is that a timer used here won't cause trouble elsewhere or in the future, as it's run **only** when the database is **constructed**, and _only_ when this function is **called inside the constructor.** Which is (I believe) before the user will even be able to interact with the app, so it will not really be noticeable to the user.

shared/src/config.ts
```diff
@@ -0,0 +1,29 @@
...
+    public getConfig() {
+        return this.config;
+    }
+
+    public getCmsFlag() {
+        return this.config.cms;
+    }
+}
+
+export const config: Configuration = new Configuration({ cms: true });
```
laurensgroeneveld, line 25: What I was saying in the standup, how I expected this to work is this class is only instantiated the first time this variable is imported. Evidently this is not the case as `config` is definitely imported before `db`, but the latter already exists. So a good learning opportunity to dive a bit into how JS/TS does these things.

DirkDevelopmentFreelance: Sequence
From this image, you can see that the area where the config is made in cms, is run after the database. So I agree, Something worth going below the water for. Going to research this and how NodeJS handles modules and export variables, before continuing on this PR.

api/src/db/seedingDocs/content-post1-eng.json
DirkDevelopmentFreelance, line 1: I modified the API's default seeded database to test if expired documents will be deleted when not in Cms mode. It seems that way, just a bit confusing as the logs that I made locally says those documents are deleted,but they still appear in the IndexedDB and no posts are being shown on the App only "newest content"

DirkDevelopmentFreelance:
### CURRENT STATUS -- CLEANUP
I am currently **just cleaning up** the PR and making it ready for review. Deleting expired Documents feature and test **has been added**. _Refactored luminary-shared_.

PR #302, opened 2024-08-24
option to hide the publish date 

belljohan3:


ivanslabbert (changes requested):
Good work @belljohan3 

Some comments:

- The field must also be added to ContentDto, and copied from the incoming PostDto/TagDto in the API
- I think we need to add something to set the default value of `publishDateVisible` to false when creating a new Tag. On Posts the default of true is good. Should be ok to only do this in the CMS.
- Add a test to check if the publish date is shown when `publishDateVisible` is `undefined` (if you have not done so already)
- Add a Add a test to check if the publish date is hidden when `publishDateVisible` is `false` (if you have not done so already)

belljohan3:
As discussed, it might be better to run an upgrade in the database schema with functionnality added to #290.

ivanslabbert:
@belljohan3 we should please go through the conflicts together to avoid reverting recent changes that was merged to main?

ivanslabbert:
@belljohan3 I am getting an error in the CMS console when loading an existing Post / Tag document:
image

cms/src/components/content/EditContent.vue
```diff
@@ -125,6 +128,7 @@ const createTranslation = (language: LanguageDto) => {
         title: `Translation for ${language.name}`,
         slug: "",
         parentTags: [],
+        publishDateVisible: true,
```
ivanslabbert, line 131: No need to set the `publishDateVisible` field on the ContentDto documents as this is done in the API.

cms/src/components/content/EditContentParentValidation.vue
```diff
@@ -69,6 +69,7 @@ const createTranslation = (language: LanguageDto) => {
         title: `Translation for ${language.name}`,
         slug: "",
         parentTags: [],
+        publishDateVisible: true,
```
ivanslabbert, line 72: No need to set the `publishDateVisible` field on the `ContentDto` documents as this is done in the API.

cms/src/tests/mockdata.ts
```diff
@@ -21,6 +21,7 @@ export const mockCategoryDto: TagDto = {
     pinned: false,
     image: "test-image.jpg",
     tags: [],
+    publishDateVisible: true,
```
ivanslabbert, line 24:
```suggestion
    publishDateVisible: false,
```

cms/src/tests/mockdata.ts
```diff
@@ -32,6 +33,7 @@ export const mockTopicDto: TagDto = {
...
```
ivanslabbert, line 36:
```suggestion
    publishDateVisible: false,
```

cms/src/tests/mockdata.ts
```diff
@@ -125,6 +131,7 @@ export const mockCategoryContentDto: ContentDto = {
     text: '{"type":"doc","content":[{"type":"paragraph","content":[{"type":"text","text":"A category"}]}]}',
     publishDate: 1704114000000,
     parentImage: "test-image.jpg",
+    publishDateVisible: true,
```
ivanslabbert, line 134:
```suggestion
    publishDateVisible: false,
```

cms/src/tests/mockdata.ts
```diff
@@ -143,6 +150,7 @@ export const mockTopicContentDto: ContentDto = {
...
```
ivanslabbert, line 153:
```suggestion
    publishDateVisible: false,
```

api/src/dto/ContentDto.ts
```diff
@@ -98,4 +107,9 @@ export class ContentDto extends _contentBaseDto {
     @IsString()
     @Expose()
     parentTagType?: TagType;
+
+    @IsOptional() // Optional as it is set upon change request processing
+    @IsBoolean()
+    @Expose()
+    publishDateVisible?: boolean;
```
ivanslabbert, line 114: This is perhaps debatable, but as this field is replicated from the parent post/tag document, it should perhaps be renamed to `parentPublishDateVisible` to keep it in line with other replicated fields?

belljohan3:
Yes I am aware of that. It is coming from EditContentParent component. It seems like we need a computed function that set a the new value to the publishdateVisible when the parent changed

ivanslabbert (changes requested):
As an upgrade script has been added updating existing documents to include `publishDateVisible` field, we can remove code in the CMS and app handling cases where `publishDateVisible` is undefined. The field can also be made mandatory (currently it is optional).

ivanslabbert:
> Yes I am aware of that. It is coming from EditContentParent component. It seems like we need a computed function that set a the new value to the publishdateVisible when the parent changed

Or use a v-if on the toggle component?

belljohan3:
> > Yes I am aware of that. It is coming from EditContentParent component. It seems like we need a computed function that set a the new value to the publishdateVisible when the parent changed
> 
> Or use a v-if on the toggle component?

I think as the field is now mandatory, the warning is not showing anymore.

app/src/components/content/ContentTile.spec.ts
```diff
@@ -60,6 +60,16 @@ describe("ContentTile", () => {
         expect(wrapper.text()).not.toContain("Jan 1, 2024");
     });
 
+    it("hides the publishDate is false", async () => {
```
ivanslabbert, line 63:
```suggestion
    it("hides the publishDate of publishDateVisible is false", async () => {
```

app/src/components/content/ContentTile.vue
```diff
@@ -25,7 +25,10 @@ const openContent = () => {
     >
         <img :src="content.parentImage" class="aspect-video rounded-lg object-cover shadow-md" />
         <h3 class="mt-2 text-sm text-zinc-800 dark:text-zinc-50">{{ content.title }}</h3>
-        <div v-if="showPublishDate" class="mt-0.5 text-xs text-zinc-500 dark:text-zinc-200">
+        <div
+            v-if="props.showPublishDate && content.parentPublishDateVisible"
```
ivanslabbert, line 29: Not needed to include `props.xxx` in the html as Vue does this automatically

ivanslabbert (approved):
Approved, but please have a look at two minor things I commented on?

PR #303, opened 2024-08-25
Handle rejected JWT tokens by redirecting to Auth0 login

ivanslabbert:


app/src/App.vue
```diff
@@ -9,24 +9,43 @@ import { getSocket } from "luminary-shared";
...
+
+    localStorage.removeItem("auth0AuthFailedRetryCount");
+    localStorage.removeItem("usedAuth0Connection");
+    await logout({ logoutParams: { returnTo: window.location.origin } });
+};
+
+// Clear the auth0AuthFailedRetryCount if the user logs in successfully (if the app is not redirecting to the login page, we assume the user either logged out or the login was successful)
+setTimeout(() => {
+    localStorage.removeItem("auth0AuthFailedRetryCount");
+}, 10000);
```
laurensgroeneveld, line 40: Doesn't `loginWithRedirect` always redirect to the login page? If you're already logged in you will immediately be redirected back without actually seeing the page, but it will route you via Auth0 I think. I feel like the clearing of this rather belongs in the `onBeforeMount` where we have the `await waitUntilAuth0IsLoaded()`, after which we can be sure Auth0 is initialized. But perhaps you've already thought of all of this and there is a reason this is here. As you know I always itch when I see a `setTimeout` 😛

laurensgroeneveld (approved):
I have a comment but otherwise this looks fine, and I can still log in and out locally so it seems to work 😅

app/src/App.vue
ivanslabbert: No, I tested, and it does not always go to the login page. If you still have a valid session, it redirects you back to the app with a working token.

Do not like the timeout either, but I cannot think of another way to determine if the user is logged into the app successfully to clear the retry counter.

PR #310, opened 2024-08-27
Implement image editor

ivanslabbert:


ivanslabbert:
This PR is closed in favour of a simpler approach to image handling, where images are directly attached to PostDto and TagDto documents instead of using separate documents for images. The same storage structure will still be used (S3), but file data will be directly attached to the PostDto / TagDto documents and replicated to their child ContentDto documents.

PR #311, opened 2024-08-28
Add filtering to content overview

DirkDevelopmentFreelance:


cms/src/components/content/ContentOverview.vue
```diff
@@ -86,44 +96,116 @@ const titleType = tagTypeString ? tagTypeString : props.docType;
...
+        if (filter.filter == "translation")
+            queryOptions.value.translationStatus = filter.by as
+                | "all"
+                | "translated"
+                | "untranslated"
+                | undefined;
+        console.log(filter);
+    });
+});
+
```
DirkDevelopmentFreelance, line 197: ### Filter implementation
I am adding filters this way, but it seems that it's not really working

cms/src/components/content/ContentOverview.vue
```diff
@@ -179,19 +184,42 @@ const handleFilterDelete = async (event: Event) => {
...
+            queryOptions.value.publishStatus = filter.by as
+                | "all"
+                | "published"
+                | "scheduled"
+                | "expired"
+                | "draft"
+                | undefined;
+        }
     });
 });
```
DirkDevelopmentFreelance, line 224: @ivanslabbert  @laurensgroeneveld  This is how the filtering is currently working. It works to a degree but I feel there can be a better solution than doing it this way. As this method is buggy.

laurensgroeneveld:
Feedback from the standup:

- Look into splitting up the ticket into separate parts (search / sort / filtering), to make it easier to review and fix bugs
   - This would make it far easier to merge the individual parts. If needed, you can base one branch on the other. Or just keep say search in review now, and open a PR for sorting when search is done and merged.
- Sorting: the reset button could be confusing. Sorting doesn't need to reset
   - Remove the global reset and add individual reset buttons for search and filtering
 - Sorting: Add last updated sort (the default) so people can always go back to that
 - Search: Keep the search term in the search bar
 - Search: Add a small delay before the search actually goes through (calls the query). For example, 200ms after the user stops typing
 - Search: Add reset button (could also be nice-to-have for a follow-up PR as you can manually clear the input)
 - Filters: typing the filters is probably too advanced. Instead, have some simple dropdowns where the user can select the options.
 - Filters: clear all filters button (could also be in a follow-up PR)
 
Future considerations (not for these PRs):
- Loading state so it's clear something is happening. Nice-to-have
- Fix that sorting also filters the list, in the case of filters like expiry date

cms/src/components/content/ContentOverview.vue
```diff
@@ -77,12 +78,83 @@ if (!Object.entries(TagType).some((t) => t[1] == tagTypeString)) tagTypeString =
...
+    },
+    {
+        value: "draft",
+        label: "Drafted Documents",
+    },
+    {
+        value: "all",
+        label: "All Documents",
+    },
+];
```
ivanslabbert, line 122: In my opinion we can leave out "documents". The Document concept is something we use internally to describe a piece of JSON, but for the users it might be less intuitive.

ivanslabbert (changes requested):
The filters work as expected - good work!

image

1. I would expect the filter button to invert colours showing that it is active
2. I do not think we need the "filter items" heading?
3. The colour scheme of the filter section feels a bit out of place. Perhaps it should have the same look and feel as the table (shadow, white background)?
4. The labels for the dropdown boxes could perhaps be inline, and smaller text? Then the whole thing would feel more compact.
5. We are still missing the tag filter, but that can perhaps be done in a separate PR.

laurensgroeneveld (changes requested):
Some small comments, but it works great 👍

cms/src/components/content/ContentOverview.vue
```diff
@@ -77,12 +78,87 @@ if (!Object.entries(TagType).some((t) => t[1] == tagTypeString)) tagTypeString =
...
 <template>
     <BasePage :title="`${capitaliseFirstLetter(titleType)} overview`">
         <template #actions>
             <div class="flex gap-4">
+                <button
+                    @click="handleShowFilterOptions"
+                    class="rounded-md border-[1px] p-2 px-3 shadow-sm"
+                    :style="{
+                        backgroundColor: showFilterOptions ? '#18181b' : 'white',
+                        color: showFilterOptions ? 'white' : '#18181b',
```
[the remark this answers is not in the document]
DirkDevelopmentFreelance: Yeah maybe you could help me with this. As this was the only method I found that worked for this part of the code. Could be nice to learn what a better way could be

cms/src/components/content/ContentOverview.vue
```diff
@@ -77,12 +78,87 @@ if (!Object.entries(TagType).some((t) => t[1] == tagTypeString)) tagTypeString =
...
+});
+
+watch(filterByStatus, () => {
+    queryOptions.value.publishStatus = filterByStatus.value as
+        | "all"
+        | "published"
+        | "scheduled"
+        | "expired"
+        | "draft"
+        | undefined;
```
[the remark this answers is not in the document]
DirkDevelopmentFreelance: On my side I kept getting a error like `cant assign "type a, b" to "type a"`

cms/src/components/content/ContentOverview.vue
```diff
@@ -108,6 +184,31 @@ router.currentRoute.value.meta.title = `${capitaliseFirstLetter(titleType)} over
                 </LButton>
             </div>
         </template>
+        <div
```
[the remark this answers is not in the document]
DirkDevelopmentFreelance: This will be done in the follow-up visual PR

cms/src/components/content/ContentOverview.vue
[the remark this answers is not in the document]
DirkDevelopmentFreelance: I actually think when we merge it, inverting the colors will look inconsistent with the sort button, so that part of the visual should be added when merged if we decide on it.

cms/src/components/content/ContentOverview.spec.ts
```diff
@@ -126,6 +126,32 @@ describe("ContentOverview.vue", () => {
         });
     });
 
+    it("should display filter options and inputs", async () => {
```
laurensgroeneveld, line 129: I'm missing some tests here for the functionality.

cms/src/components/content/ContentOverview.spec.ts
```diff
@@ -126,6 +126,32 @@ describe("ContentOverview.vue", () => {
...
+            },
+        });
+
+        await waitForExpect(async () => {
+            const showFilterOptionsBtn = wrapper.find('[data-test="show-filter-options-btn"]');
+            expect(showFilterOptionsBtn.exists()).toBe(true);
+            await showFilterOptionsBtn.trigger("click");
+
+            const filterOptions = wrapper.find('[data-test="filter-options"]');
+            expect(filterOptions.exists()).toBe(true);
```
laurensgroeneveld, line 145: Similar to the other PR, I don't think testing that the wrapping div exists is totally necessary.

cms/src/components/content/ContentTable.vue
```diff
@@ -13,11 +13,12 @@ const props = defineProps<Props>();
 const languages = db.whereTypeAsRef<LanguageDto[]>(DocType.Language, []);
 
 const contentDocs = contentOverviewQueryAsRef(props.queryOptions);
+
 </script>
 
 <template>
     <LCard padding="none">
-        <div class="overflow-x-auto rounded-md">
+        <div class="overflow-x-auto rounded-b-md">
```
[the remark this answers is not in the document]
laurensgroeneveld: image

This is still relevant. Not sure why the top rounding is removed? It makes the table look a bit ugly.

laurensgroeneveld (changes requested):
Looking good, just missing some tests for the functionality.

cms/src/components/content/ContentOverview.spec.ts
DirkDevelopmentFreelance: Discussed on Discord

cms/src/components/content/ContentOverview.spec.ts
DirkDevelopmentFreelance: Added tests

laurensgroeneveld (approved):
Nice 👍 As discussed we'll do a visual pass in another ticket, but it looks & works great already!

PR #312, opened 2024-08-28
refactoring notification system to support link

belljohan3:


app/src/components/notifications/NotificationBanner.vue
```diff
@@ -51,7 +51,35 @@ switch (props.notification.state) {
 
 <template>
     <div v-if="show" class="inset-x-0 top-0 z-50 text-zinc-900" :class="color">
-        <div class="flex items-center justify-between px-6 py-1">
+        <RouterLink
```
laurensgroeneveld, line 54: Copying all of the markup just to render a `RouterLink` or not will be a bit cumbersome to maintain I think. Instead, I recommend to use a different solution to dynamically render it. This can probably be achieved by using the dynamic `component`, like we do for the icon. `<component :is="notification.routerLink ? RouterLink : 'div'"`

belljohan3: @laurensgroeneveld Thanks for your insight. I completely forgot that way to do.

app/src/App.vue
```diff
@@ -102,6 +102,30 @@ setTimeout(() => {
...
 }, 1000);
+
+setTimeout(() => {
+    watch(
+        [isConnected, isAuthenticated],
+        () => {
+            if (isConnected.value && !isAuthenticated.value) {
+                useNotificationStore().addNotification({
+                    id: "accountBanner",
+                    title: "You are missing out !",
```
laurensgroeneveld, line 113: Nitpick: remove the space before the exclamation mark.

```suggestion
                    title: "You are missing out!",
```

app/src/components/notifications/NotificationBanner.vue
```diff
@@ -51,26 +52,32 @@ switch (props.notification.state) {
...
-                <component :is="icon" class="h-5 min-h-5 w-5 min-w-5" />
-                <div class="flex flex-col md:inline-block md:align-middle">
-                    <span class="text-md md:text-sm">{{ notification.title }}</span>
-                    <span v-if="notification.description" class="text-xs md:ml-3">{{
-                        notification.description
-                    }}</span>
+        <component
+            :is="notification.routerLink ? RouterLink : 'div'"
+            :to="notification.routerLink"
+            @click="notification.id ? removeNotification(notification.id) : (show = false)"
```
laurensgroeneveld, line 58: I think this click handler should be moved - if I click the close button, it will also route me to the page, because this also triggers. That's not what I expect.

Perhaps two tests for this could also be added, to both the banner and toast:
- Clicking the link will start a router navigation (this will require a mocked router in the test)
- Clicking the close button while there is a link will only close the notification and not route the user

app/src/components/notifications/NotificationBanner.vue
```diff
@@ -51,26 +52,32 @@ switch (props.notification.state) {
...
+        >
+            <div class="flex items-center justify-between px-6 py-1">
+                <div class="flex items-center gap-2">
+                    <component :is="icon" class="h-5 min-h-5 w-5 min-w-5" />
```
laurensgroeneveld, line 62: With a fixed width and height I don't think the `min-` classes are needed here.

laurensgroeneveld (changes requested):
Found one issue with closing the notification, but looks good otherwise 👍

app/src/components/notifications/NotificationBanner.vue
belljohan3: How can we dismiss the notification when clicking on it ?

laurensgroeneveld: With the close button: 
image

belljohan3: Okay.

app/src/components/notifications/NotificationBanner.vue
```diff
@@ -51,26 +52,28 @@ switch (props.notification.state) {
...
+                    type="button"
+                    @click="notification.id ? removeNotification(notification.id) : (show = false)"
+                    class="h-6 min-h-6 w-6 min-w-6 cursor-pointer underline md:h-5 md:min-h-5 md:w-5 md:min-w-5"
+                    data-test="banner-close-button"
+                >
+                    <span class="sr-only">Close</span>
+                    <XMarkIcon class="h-5 w-5" aria-hidden="true" />
+                </button>
+            </div>
+        </component>
```
laurensgroeneveld, line 77: Because the button is inside the routerlink, clicking close will still bring me to the login page. The close button should probably be outside of the routerlink. Perhaps only the actual `description` should be clickable?

belljohan3: I found also that. I can fix it

app/src/components/notifications/NotificationToast.vue
```diff
@@ -54,37 +55,41 @@ switch (props.notification.state) {
...
-                        @click="
-                            notification.id ? removeNotification(notification.id) : (show = false)
-                        "
-                        class="inline-flex rounded-md bg-white text-zinc-400 hover:text-zinc-500 focus:outline-none focus:ring-2 focus:ring-indigo-500 focus:ring-offset-2"
-                        data-test="toast"
-                    >
-                        <span class="sr-only">Close</span>
-                        <XMarkIcon class="h-5 w-5" aria-hidden="true" />
-                    </button>
+        <component :is="notification.routerLink ? RouterLink : 'div'" :to="notification.routerLink">
```
laurensgroeneveld, line 58: If it is a router link, it should close the notification when clicked. I think adding a click handler here would solve that (I think it was there before but was removed in a commit).

laurensgroeneveld (approved):
Nice work 🙌

PR #317, opened 2024-08-31
Add search and sort ui

DirkDevelopmentFreelance:


DirkDevelopmentFreelance:
**Todo**

- [ ] Add test for search UI in content overview
- [ ] Add test for sort UI in content overview

ivanslabbert (dismissed):
@DirkDevelopmentFreelance looks good, and super nice to have sorting and searching working!

Found a few linting warnings - perhaps just have a quick look in the "files changed" tab?

image

- I am missing an indication whether we are sorting Ascending or Descending. Perhaps the selected option should be marked in some or other way?
- The text feels a bit too large and dark to me, but I think @laurensgroeneveld will be able to give better feedback on the visual side.

image

- Is there a reason for having the gap between the sorting button and the search box?
- I think the highlight colour in the search box is a bit dark, but again I think this is more something for @laurensgroeneveld to comment on.

laurensgroeneveld:
> Is there a reason for having the gap between the sorting button and the search box?

I actually like that more. I left a comment on the other PR to bring the filters into this bar as well. Then everything is together with filters and sorting on the right side and searching on the left side, which is quite a common arrangement.

cms/src/components/content/ContentOverview.vue
```diff
@@ -131,6 +171,104 @@ router.currentRoute.value.meta.title = `${capitaliseFirstLetter(titleType)} over
...
         /> -->
+        <div class="flex h-14 w-full gap-2 rounded-t-md bg-white p-2 shadow-lg">
+            <input
+                type="text"
+                class="h-full w-2/4 rounded-md border-[1px] border-zinc-200 p-4 text-sm focus:border-zinc-950 focus:bg-zinc-50 focus:outline-none focus:ring-0"
+                name="search"
+                placeholder="Search..."
+                data-test="search-input"
+                v-model="searchTerm"
+            />
```
laurensgroeneveld, line 182: Any reason why you didn't use the `LInput` component here? It looks a little different.

cms/src/components/content/ContentOverview.vue
```diff
@@ -131,6 +171,104 @@ router.currentRoute.value.meta.title = `${capitaliseFirstLetter(titleType)} over
...
+            <div>
+                <div class="relative flex h-full gap-1">
+                    <button
+                        class="flex h-full w-10 flex-row content-center justify-center rounded-md border-[1px] focus:border-black focus:outline-none"
+                        @click="() => (showSortOptions = true)"
+                        data-test="sort-toggle-btn"
+                    >
```
laurensgroeneveld, line 189: Same question, why is this not an `LButton`?

cms/src/components/content/ContentOverview.vue
```diff
@@ -131,6 +171,104 @@ router.currentRoute.value.meta.title = `${capitaliseFirstLetter(titleType)} over
...
+                name="search"
+                placeholder="Search..."
+                data-test="search-input"
+                v-model="searchTerm"
+            />
+            <div>
+                <div class="relative flex h-full gap-1">
+                    <button
+                        class="flex h-full w-10 flex-row content-center justify-center rounded-md border-[1px] focus:border-black focus:outline-none"
+                        @click="() => (showSortOptions = true)"
```
laurensgroeneveld, line 187: Nitpick: No need to make this a method.

```suggestion
                        @click="showSortOptions = true"
```

would work too.

cms/src/components/content/ContentOverview.vue
```diff
@@ -131,6 +171,104 @@ router.currentRoute.value.meta.title = `${capitaliseFirstLetter(titleType)} over
...
+                    >
+                        <ArrowsUpDownIcon class="h-full w-4" />
+                    </button>
+                    <div
+                        ref="sortOptionsAsRef"
+                        class="absolute right-0 top-full mt-2 h-max w-40 rounded-lg border border-gray-300 bg-white p-2 shadow-lg"
+                        v-if="showSortOptions"
+                        data-test="sort-options-display"
+                    >
+                        <h4>Sort By:</h4>
```
laurensgroeneveld, line 198: I agree with Ivan's comments on the styling, it's almost there just needs a few tweaks to the text color and size. Perhaps something we can pair program on?

cms/src/components/content/ContentOverview.vue
```diff
@@ -131,6 +171,104 @@ router.currentRoute.value.meta.title = `${capitaliseFirstLetter(titleType)} over
...
+                        <div class="flex flex-col">
+                            <label>
+                                <input
+                                    class="ml-2 text-zinc-800 focus:border-0 focus:outline-none focus:ring-0"
+                                    type="radio"
+                                    name="sortOption"
+                                    value="title"
+                                    v-model="selectedSortOption"
+                                    data-test="sort-option-title"
+                                />
```
laurensgroeneveld, line 208: I think an `LRadio` component would be nice so that we can easily reuse it in other places.

cms/src/components/content/ContentOverview.vue
```diff
@@ -131,6 +171,104 @@ router.currentRoute.value.meta.title = `${capitaliseFirstLetter(titleType)} over
...
+                        </div>
+                        <hr class="my-2" />
+                        <div class="flex flex-col gap-1">
+                            <button
+                                @click="handleSortAscending"
+                                class="flex w-full gap-1 rounded-md p-2 text-zinc-950 hover:bg-zinc-300"
+                                :style="{
+                                    backgroundColor: sortingAscending ? '#18181b' : 'white',
+                                    color: sortingAscending ? 'white' : '#18181b',
+                                }"
```
laurensgroeneveld, line 253: I have the same comment here as on the other PR. Since this doesn't necessarily make sense as an `LButton` (it's a situational button with different styling), I would in this case just style it with a dynamic `:class`. We could take that up with the rest of the styling.

cms/src/components/content/ContentOverview.vue
```diff
@@ -131,6 +171,104 @@ router.currentRoute.value.meta.title = `${capitaliseFirstLetter(titleType)} over
...
+                        data-test="sort-toggle-btn"
+                    >
+                        <ArrowsUpDownIcon class="h-full w-4" />
+                    </button>
+                    <div
+                        ref="sortOptionsAsRef"
+                        class="absolute right-0 top-full mt-2 h-max w-40 rounded-lg border border-gray-300 bg-white p-2 shadow-lg"
+                        v-if="showSortOptions"
+                        data-test="sort-options-display"
+                    >
```
laurensgroeneveld, line 197: This looks and works great. One small thing I'm missing is that it doesn't close on `Esc`. And as a nitpick, a small CSS animation would be cool. We have both in the profile menu. Perhaps we should use the [HeadlessUI menu component](https://headlessui.com/v1/vue/menu) here as well?

cms/src/components/content/ContentOverview.vue
```diff
@@ -77,6 +79,44 @@ if (!Object.entries(TagType).some((t) => t[1] == tagTypeString)) tagTypeString =
...
+        queryOptions.value.search = searchTerm.value;
+    }, 500);
+});
+
+watch(selectedSortOption, () => {
+    queryOptions.value.orderBy = selectedSortOption.value as
+        | "title"
+        | "updatedTimeUtc"
+        | "publishDate"
+        | "expiryDate";
```
laurensgroeneveld, line 102: Same comment as the other PR on the cast here, don't think it's needed.

cms/src/components/content/ContentOverview.vue
```diff
@@ -77,6 +79,44 @@ if (!Object.entries(TagType).some((t) => t[1] == tagTypeString)) tagTypeString =
 
 const titleType = tagTypeString ? tagTypeString : props.docType;
 router.currentRoute.value.meta.title = `${capitaliseFirstLetter(titleType)} overview`;
+
+const searchTerm = ref("");
+
+const sortOptionsAsRef = ref(null);
```
laurensgroeneveld, line 85: Please don't use `null`. Perhaps we should put that in an ADR 😅 But `null` in JavaScript is a bit problematic because `undefined` also exists, so if you want to check if something exists and only do `foo == undefined`, that might not return true if it's null. So instead, always initialize with `undefined` (if needed) and avoid the usage of `null`.

cms/src/components/content/ContentOverview.vue
```diff
@@ -77,6 +79,44 @@ if (!Object.entries(TagType).some((t) => t[1] == tagTypeString)) tagTypeString =
...
+
+const showSortOptions = ref(false);
+
+const selectedSortOption = ref("updatedTimeUtc");
+
+watch(searchTerm, () => {
+    setTimeout(() => {
+        queryOptions.value.search = searchTerm.value;
+    }, 500);
+});
```
laurensgroeneveld, line 95: I don't think this works the way you think it will: What I think happens here is that for every update to `searchTerm`, it will create a new timeout, so in the end we're still updating `queryOptions` every time the user types. Instead, what we want to do is called a "debounce" or "debouncing the input". This can be achieved in multiple ways. The lodash library, which we already use, has `debounce` method.

api/package-lock.json
laurensgroeneveld, line 1: I see in both PRs the package-lock is changed. This will give merge conflicts. Is that intended, was it needed to update a dependency to a minor change? Or is this simply the result of running `npm install`. As a rule of thumb, only `npm ci` should be used unless you specifically want to update the dependencies. `ci` will respect the lock file, `install` will just install new minor versions.

laurensgroeneveld (changes requested):
It works great, just some small comments, some are the same as the other PR. With the same caveat that we could also look at visual changes after the two are merged.

cms/src/components/content/ContentOverview.vue
DirkDevelopmentFreelance: Gave an error on my side. this fixed the error

cms/src/components/content/ContentOverview.vue
```diff
@@ -77,6 +79,44 @@ if (!Object.entries(TagType).some((t) => t[1] == tagTypeString)) tagTypeString =
...
+
+watch(selectedSortOption, () => {
+    queryOptions.value.orderBy = selectedSortOption.value as
+        | "title"
+        | "updatedTimeUtc"
+        | "publishDate"
+        | "expiryDate";
+});
+
+const sortingAscending = ref(false);
```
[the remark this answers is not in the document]
DirkDevelopmentFreelance: This is to have the default value inside the queryOptions "desc" represents false, thats why I default false
This makes it so when the user wants to sort, the current sorting mode is visual to them.

cms/src/components/content/ContentOverview.vue
DirkDevelopmentFreelance: Will keep that in mind, and change it to undefined. Thanks for the advice.

cms/src/components/content/ContentOverview.vue
laurensgroeneveld: As discussed, we could make this a follow-up ticket. Take the BCC component as an example: https://github.com/bcc-code/bcc-design/blob/main/design-library/src/components/BccRadio/BccRadio.vue

cms/src/components/content/ContentOverview.vue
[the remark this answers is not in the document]
laurensgroeneveld:
```vue
const sortingAscending = computed(() => queryOptions.value.orderDirection == 'asc');
```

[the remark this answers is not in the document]
DirkDevelopmentFreelance: Discussed on a meeting Friday

cms/src/components/content/ContentOverview.spec.ts
```diff
@@ -102,6 +102,54 @@ describe("ContentOverview.vue", () => {
...
+
+        //data-test="sort-option-title"
+
+        await waitForExpect(async () => {
+            const sortToggleBtn = wrapper.find('[data-test="sort-toggle-btn"]');
+            expect(sortToggleBtn.exists()).toBe(true);
+            await sortToggleBtn.trigger("click");
+
+            const sortOptionsDiv = wrapper.find('[data-test="sort-options-display"]');
+            expect(sortOptionsDiv.exists()).toBe(true);
```
laurensgroeneveld, line 139: I think this is of limited utility: testing that the actual options are displayed is fine, although it would perhaps be more valuable to test their functionality. I don't think it really matters that their container div is also rendered - perhaps we refactor the HTML in the future, and then this test would fail, even though the functionality could still work.

cms/src/components/content/ContentOverview.spec.ts
laurensgroeneveld, line 1: I'm missing tests for the functionality of search and sort, currently their are only tests that they display, but we don't know if they actually work. This would also come in handy when we add (a test for) an empty state.

laurensgroeneveld (changes requested):
I miss an empty state for when search/sort/filter returns nothing, but we can solve that in a followup PR when we put it all together as the same issue would apply to the filtering.

image

Edit: I see there is a TODO about it in ContentOverview already.

cms/src/components/forms/LRadio.vue
```diff
@@ -0,0 +1,83 @@
...
+
+const id = `l-radio-${useId()}`;
+
+const emit = defineEmits(["update:modelValue"]);
+
+const { attrsWithoutStyles } = useAttrsWithoutStyles();
+</script>
+
+<template>
+    <div class="flex items-center justify-stretch gap-2" :style="$attrs['style'] as StyleValue">
```
laurensgroeneveld, line 69: You're binding the `style` here, but the `class` should also be there (look at the `LInput`).

cms/src/components/content/ContentOverview.spec.ts
DirkDevelopmentFreelance: Added test functionality for search and sort options.

cms/src/components/content/ContentTable.vue
```diff
@@ -16,28 +16,18 @@ const contentDocs = contentOverviewQueryAsRef(props.queryOptions);
 </script>
 
 <template>
-    <LCard padding="none">
+    <LCard class="rounded-t-none" padding="none">
```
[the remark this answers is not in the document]
DirkDevelopmentFreelance: I think it will be better to make visual changes after the merge. I agree that it's something for the follow up PR.

cms/src/components/forms/LRadio.vue
DirkDevelopmentFreelance: Added class

cms/src/components/content/ContentOverview.vue
laurensgroeneveld: I see this still needs to be changed to `undefined`.

cms/src/components/forms/LRadio.vue
laurensgroeneveld, line 1: There's an open comment on this file. And perhaps add it to the sandbox.

laurensgroeneveld (approved):
Nice work 👍

cms/src/components/content/ContentOverview.vue
DirkDevelopmentFreelance: Changed

cms/src/components/content/ContentOverview.spec.ts
DirkDevelopmentFreelance: Functionality testing was added

cms/src/components/forms/LRadio.vue
```diff
@@ -0,0 +1,83 @@
...
+    label?: string;
+    variant?: RadioProps["variant"];
+    size?: RadioProps["size"];
+    context?: RadioProps["context"];
+    disabled?: boolean;
+};
+
+const props = withDefaults(defineProps<Props>(), {
+    context: "default",
+    variant: "primary",
```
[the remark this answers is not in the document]
DirkDevelopmentFreelance: As this has to do with the UI, I will be updating this LRadio component in the follow-up UI PR

PR #319, opened 2024-09-04
show related content and topics

belljohan3:


app/src/pages/RelatedContent.vue
laurensgroeneveld, line 1: This component does not belong in the `pages` directory: the only components here are top-level pages that the router routes to, like the SettingsPage, or the content page (SingleContent). This is a smaller component _on_ a page, so it should be put in the `components/content` folder.

app/src/components/tags/VerticalTagViewer.vue
```diff
@@ -0,0 +1,86 @@
...
+                        ' border-l-4 border-yellow-500 bg-yellow-300 bg-opacity-10':
+                            isContentSelected(content.slug),
+                    }"
+                >
+                    <div class="ml-4 mt-1 w-2/3">
+                        <h1 class="text-[0.9rem]">
+                            {{ content.title }}
+                        </h1>
+                        <p class="text-[0.8rem]">
+                            {{ content.summary }}
```
laurensgroeneveld, line 69: Please keep to Tailwind's font size system, unless there is a very good reason not to. Otherwise we get a lot of very small differences in font size. In this case, using `sm`/`xs` font size would work as well probably.

laurensgroeneveld: Actually do we need the `summary` here? That can be quite long. I suggest to only display the title, then that can have a bit bigger font size as well.

app/src/components/tags/VerticalTagViewer.vue
```diff
@@ -0,0 +1,86 @@
...
+            <RouterLink
+                v-for="content in taggedDocs"
+                :key="content._id"
+                :to="{
+                    name: 'content',
+                    params: { slug: content.slug },
+                }"
+            >
+                <div
+                    class="flex justify-between border-l-4 border-transparent bg-opacity-20 px-2 py-[2px] transition duration-200 hover:border-opacity-100 hover:bg-yellow-300 hover:bg-opacity-10"
```
laurensgroeneveld, line 58: Same point on using custom values (`py-[2px]`), there should be a good reason for it. In this case, `0.5` can be used instead of `2px`, as they are essentially the same value.

```suggestion
                    class="flex justify-between border-l-4 border-transparent bg-opacity-20 px-2 py-0.5 transition duration-200 hover:border-opacity-100 hover:bg-yellow-300 hover:bg-opacity-10"
```

laurensgroeneveld (changes requested):
Looks good, have a few minor comments.

app/src/components/content/RelatedContent.vue
```diff
@@ -0,0 +1,54 @@
...
+
+// Function to fetch content based on tags
+async function fetchContentForTags() {
+    const tagIds = tags.value.map((t: TagDto) => t._id);
+    const contentPromises = tagIds.map((tagId) =>
+        db.contentWhereTag(tagId, { languageId: appLanguageIdAsRef.value }),
+    );
+
+    content.value = (await Promise.all(contentPromises)).flat();
+}
```
laurensgroeneveld, line 25: Instead of running an additional query for each tag, can't we fetch all the content at once?

I'm also wondering why this even happens. It's only used for a check below to check if there is content. But isn't that always the case? I don't think there should be any orphaned tags (i.e. tags without any content) in the system.

app/src/components/tags/VerticalTagViewer.vue
```diff
@@ -0,0 +1,89 @@
...
+const isContentSelected = (slug: string) => {
+    if (router.currentRoute.value.params.slug === slug) return true;
+    return false;
+};
+
+const taggedDocs = db.contentWhereTagAsRef(props.tag?._id, props.queryOptions);
+
+const currentContent = computed(() => {
+    return taggedDocs.value.filter((doc) => doc._id !== props.currentPostId);
+});
```
laurensgroeneveld, line 27: Why is the current post filtered out here? Don't we want to display it with a highlight? This leads to two weird things on my end:

1. The sidebar looks very empty when there is only one other post in the category
image

2. If the post is the only one within a tag, the whole category is empty.
image

laurensgroeneveld (changes requested):

app/src/pages/SingleContent.vue
```diff
@@ -117,45 +148,74 @@ const text = computed(() => {
...
+                    )"
+                    :key="tag._id"
+                    :tag="tag"
+                    :currentPostId="content._id"
+                    :queryOptions="{
+                        filterOptions: { docType: DocType.Post },
+                        languageId: appLanguageIdAsRef,
+                    }"
+                />
+            </div>
```
laurensgroeneveld, line 212: Switching between the categories should not be the responsibility of `SingleContent`. This components should only be concerned with rendering out the content of an article. Anything else can happen in different components. So `VerticalTagViewer` should also do the tabs with different categories and switching between them.

app/src/pages/SingleContent.vue
```diff
@@ -117,45 +148,74 @@ const text = computed(() => {
...
             </div>
+
+            <div>
+                <VerticalTagViewer
+                    v-for="tag in tagCategory.filter(
+                        (t) => t.tagType == TagType.Category && t._id == selectedTag,
+                    )"
+                    :key="tag._id"
+                    :tag="tag"
+                    :currentPostId="content._id"
```
laurensgroeneveld, line 206: This is called `currentPostId` but you're passing the `content._id`, so this should probably be named `currentContentId`?

app/src/pages/SingleContent.vue
```diff
@@ -79,26 +93,43 @@ watch(
...
+const isLoading = computed(() => content.value === undefined);
 
 const text = computed(() => {
     if (!content.value.text) {
         return "";
     }
 
-    let text;
-
-    // Only parse text with TipTap if it's JSON, otherwise we render it out as HTML
```
laurensgroeneveld, line 93: This comment seems important, so it shouldn't be removed.

app/src/components/content/RelatedContent.vue
laurensgroeneveld: > I'm also wondering why this even happens. It's only used for a check below to check if there is content. But isn't that always the case? I don't think there should be any orphaned tags (i.e. tags without any content) in the system.

As discussed, this only happens because otherwise we show an empty "related content" if the current post only belongs to one topic with this one post in it.

laurensgroeneveld: > Instead of running an additional query for each tag, can't we fetch all the content at once?

This still can be fixed - the content is also fetched in `SingleContent`, so now we're doing it twice.

ivanslabbert (changes requested):
Works nice @belljohan3! It seems to be quite functional already.

- There are a couple of console.log() lines that you probably want to remove?
image

- SingleContent.vue has excessive padding at the bottom
image

- Please check that the category tab button colours are not too yellow. It should be less emphasized than the menu buttons. The background and vertical line can also be less yellow?
image

- In light mode the "related" background colour is the same colour as the menu bar, giving no visual separation. Perhaps we need to use the slight yellow tint as we did for pinned categories on the home page?
image

- In dark mode the "related" background looks like it is using the "zinc" palette? Perhaps we need to use the slight yellow tint as we did for pinned categories on the home page?
image

- I can open unpublished posts in the app. (Not related to this issue, but I thought this was fixed before? Did that logic get removed? Perhaps add a test for it as well.
image

- Switching between categories in SingleContent.vue causes the page to reload.
- If there are draft posts tagged with the same topic, the Related section shows the topic with no content
image

- In dark mode the hover colour when hovering over a category is dimming the background colour of selected categories and brightening the background colour of unselected categories. This feels a bit wrong.
- In light mode the hover colour when hovering over a category is brightening the background colour of selected categories and dimming the background colour of unselected categories. This feels a bit wrong.

app/src/components/tags/VerticalTagViewer.vue
```diff
@@ -0,0 +1,84 @@
...
+    return false;
+};
+
+const taggedDocs = db.contentWhereTagAsRef(props.tag?._id, props.queryOptions);
+
+const tagContent = props.tag
+    ? db.whereParentAsRef(props.tag._id, DocType.Tag, props.queryOptions.languageId, [])
+    : ref([]);
+
+const tagTitle = ref(props.title);
```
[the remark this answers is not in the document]
belljohan3: I am sure of what you mean here. I just ccpoy the same logic of HorizontalScrollableTagViewer component.

DirkDevelopmentFreelance (dismissed):
Nice job! Everything looks good to me, just 2 comments

- The background behind a pinned category and the related content doesn't fit the blueish theme for dark mode
- The background behind hovering and clicking on a post inside a category is too light, as discussed maybe try making the background the same color as the logo, or make the background just be a slightly lighter shade of the dark mode theme of the main page background.

But great job with the functionality it looks good to me!

ivanslabbert (approved):

PR #322, opened 2024-09-11
Make dark theme darker & more consistent

belljohan3:


app/src/components/navigation/TopBar.vue
```diff
@@ -8,7 +8,7 @@ const route = useRoute();
 
 <template>
     <header>
-        <div class="z-40 bg-white drop-shadow dark:bg-zinc-700">
+        <div class="z-40 bg-white drop-shadow dark:bg-slate-800">
```
DirkDevelopmentFreelance, line 11: I feel the top bar is a tiny bit too light still and that makes the contrast between the background and top bar a bit much, but that is a limitation in _**tailwindCSS**_ as there is no `bg-slate-850`. This could also just be my personal preference.

DirkDevelopmentFreelance (commented):
It looks great! Just one comment I made. But we can discuss it.

laurensgroeneveld (commented):
This looks pretty good to me. Perhaps the only thing that doesn't look great is the dark version of the card, it feels a little too light of a background.

image

belljohan3:
Thanks. So, do you suggest to make it a bit darker ?

laurensgroeneveld (commented):
Nice improvement on the card. I just noticed the notifications are still a white background. Perhaps they should be similar in color to the language/theme popup in dark mode?

image

ivanslabbert (approved):

PR #330, opened 2024-09-18
Post-tag-image-uploader

ivanslabbert:
- [x] Implement fallback image
- [ ] Redesign image editor component & add drag-drop uploading of images (create separate issue)


api/src/db/schemaUpgrade/v3.ts
```diff
@@ -0,0 +1,76 @@
...
+        await db.processAllDocs(async (doc: any) => {
+            if (!doc) {
+                return;
+            }
+            if (doc.error) {
+                console.error(`Unable to parse document: ${doc.parseError}`);
+                return;
+            }
+            if (doc.type && (doc.type == "post" || doc.type == "tag")) {
+                const d: PostDto | TagDto = doc;
```
DirkDevelopmentFreelance, line 31: I feel this variable should be more descriptive, or maybe a comment above it stating that it is used to make sure that Doc is PostDto | TagDto, or that it's just a placeholder to not directly edit the "doc" variable.

DirkDevelopmentFreelance (commented):
Everything in the code looks good to me, except for one comment.
Functionality wise I found some issues. Not sure if this PR causes them, but so far I think it does as it doesn't have these issues on the main branch or any of my other branches. @belljohan3 said he didn't experience these issues. 

1. When creating a post, buttons in the first block is disabled, they only become enabled when I add a group membership. Not sure if this is intended.
2. When you get the buttons to work after adding a group membership, and make the post, it doesn't get saved to the database after saving. 

I also found 2 other issues, but those are also on the main branch. I will note them down or add them as tickets.

api/src/db/schemaUpgrade/v3.ts
ivanslabbert: I need to check if the document is of type "Post" or "Tag". The first `doc.type` in the `if` statement is probably there to keep TypeScript happy as it would have worked in vanilla JS.

DirkDevelopmentFreelance: I am actually talking about the naming of the "d" variable. But maybe it being a single character already signifies that it is a copied version of the current "doc". The rest of the code is nice and descriptive.

belljohan3 (approved):
Everyting looks to me

PR #331, opened 2024-09-18
Add proper desktop navigation menu

belljohan3:


laurensgroeneveld (commented):
I have some comments that I'll fix myself together with a visual update, but I still think it's important for you to read them. Let me know if something isn't clear or if you don't agree with it.

laurensgroeneveld (approved):

PR #341, opened 2024-09-24
make URL clickable on text content

belljohan3:


app/package-lock.json
```diff
@@ -57,6 +58,47 @@
                 "wait-for-expect": "^3.0.2"
             }
         },
+        "../shared": {
+            "version": "0.0.3",
```
laurensgroeneveld, line 62: This doesn't seem intended. Running `npm install --install-links` instead of just `npm install` should solve it, and also the problem you have with the lock file that's out of sync.

PR #350, opened 2024-09-27
set date to match system settings

belljohan3:


shared/src/db/database.ts
```diff
@@ -466,17 +466,19 @@ class database extends Dexie {
...
+                suppressSeconds: true,
+            });
     }
 
     /**
      * Convert a DateTime object to a numeric (UNIX) date
      */
     fromDateTime(date: DateTime) {
-        return date.toMillis();
+        return date.setLocale(navigator.language || "en-US").toMillis();
```
ivanslabbert, line 481: not sure if the locale needs to be set when converting Luxon dates to Unix dates, but guess it does not do damage either.

shared/src/db/database.ts
```diff
@@ -485,7 +487,9 @@ class database extends Dexie {
      * @returns
      */
     fromIsoDateTime(date: string) {
-        return DateTime.fromISO(date).toMillis();
+        return DateTime.fromISO(date)
+            .setLocale(navigator.language || "en-US")
+            .toMillis();
```
ivanslabbert, line 492: not sure if the locale needs to be set when converting Luxon dates to Unix dates, but guess it does not do damage either.

ivanslabbert (approved):

PR #365, opened 2024-10-02
CMS: Add tag filter to content overview. 

DirkDevelopmentFreelance:
Just a commit to start the PR, added some initial code to get started with the PR/ticket.

### DIFF
--- a/cms/src/components/content/ContentOverview.vue
+++ b/cms/src/components/content/ContentOverview.vue
@@ -180,13 +180,13 @@ debouncedWatch(
         const tagIds = tags.value.map((t) => t._id);
 
         tagsContent.value = await db.whereParent(tagIds, DocType.Tag, selectedLanguage.value);
-        tagsContent.value.map((tagContent) =>
+        tagsContent.value.forEach(async (tagContent) => {
             tagsToDisplay.value.push({
                 label: tagContent.title,
-                value: tagContent._id,
+                value: tagContent.parentId,
                 isChecked: false,
-            }),
-        );
+            });
+        });
     },
     { debounce: 1 },
 );
--- a/cms/src/components/content/ContentOverview.vue
+++ b/cms/src/components/content/ContentOverview.vue
@@ -20,7 +20,7 @@ import {
     hasAnyPermission,
     type ContentDto,
 } from "luminary-shared";
-import { computed, ref, watch, type Ref } from "vue";
+import { computed, ref, watch } from "vue";
 import ContentTable from "@/components/content/ContentTable.vue";
 import LSelect from "../forms/LSelect.vue";
 import { capitaliseFirstLetter } from "@/util/string";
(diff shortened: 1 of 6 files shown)


ivanslabbert:
Hi @DirkDevelopmentFreelance 
I had a quick check, and it looks ok, but I found a few issues:

- Topics does not filter (not sure if it is query related?)
- When clicking the dropdown arrow the dropdown shows and hides.

Will the dropdown scroll if there are a lot of tags?

ivanslabbert:
The border colour of the dropdown is lighter on the tag filter compared to the sorting button. I think the tag filter looks better, so perhaps just update filter dropdown with the same colours?
image
image

belljohan3:
The categories are not displayed according to the selected language. 
image

cms/src/components/content/ContentOverview.vue
```diff
@@ -164,6 +167,54 @@ watch(selectedSortOption, () => {
...
+                label: tagContent.title,
+                value: tagContent.parentId,
+                isChecked: false,
+            });
+        } else {
+            tagsToDisplay.value[existingTagIndex].label = tagContent.title;
+        }
+    });
+});
+watch(tagsSelected.value, () => {
```
ivanslabbert, line 199: Is it correct to watch the value of a ref? Or should we rather watch the ref itself?

cms/src/components/forms/LChecklist.spec.ts
```diff
@@ -0,0 +1,79 @@
+//@ts-nocheck --> To prevent wrapper.vm.givenOptions typing error as it is valid code - block level ignore doesn't work so this is the best solution for that.
```
ivanslabbert, line 1: A bit dangerous, but as it is a test file we can live with it

cms/src/components/forms/LSelect.vue
```diff
@@ -5,7 +5,7 @@ export default {
 </script>
 
 <script setup lang="ts">
-import { type Component, type StyleValue } from "vue";
+import { ref, watch, type Component, type StyleValue } from "vue";
```
ivanslabbert, line 8: Getting a few linting errors here for imports that are not used

ivanslabbert (changes requested):
Good work Dirk!

A few small issues, but else it is good to go.

Small issue - the sort order dropdown is now a bit too tight to the button compared to the tag filter.
image

Could we sort the topic names alphabetically?
image

cms/src/components/content/ContentOverview.vue
DirkDevelopmentFreelance: From what I see it is actually a common practice, I used it because I wanted to trigger reactions.

Watching the ref's value (tagsSelected.value) allows you to specifically watch the inner value of the ref. This is useful when you want to trigger a reaction whenever the actual data inside the ref changes.

Watching the ref itself (tagsSelected) allows you to detect changes in the reference itself, such as if the whole ref object is replaced with a new one

cms/src/components/forms/LChecklist.spec.ts
DirkDevelopmentFreelance: I also don't like this approach but typescript is giving type issues for valid code and I used that code a lot of times so without this the file will be polluted with ```@ts-ignore```. If someone needs to add a test they can just remove this line temporarily, and then re-add it afterwards. From my research I discovered that  a lot of developers also struggle with this during testing.

cms/src/components/forms/LSelect.vue
DirkDevelopmentFreelance: Fixed.

ivanslabbert (approved):

PR #369, opened 2024-10-05
improve image uploader

belljohan3:


ivanslabbert (approved):
Nice work @belljohan3!

Just having a few comments, but approved the PR:

The padding on the containing div is quite big, and I think it should be smaller. You could rather add a bigger margin to the upload button to give it more space from the sides? 

The upload button is also to close to the image thumbnails.
telegram-cloud-photo-size-4-5776157652622099669-x

PR #373, opened 2024-10-08
Add inline notifaction for too large file uploads

DirkDevelopmentFreelance:
The `maxFileUploadSize` has been **set to 0.25mb** as I saw that _in socketIo it defaults to 0_, not sure if that value should have a default value other than 0? When the user tries to upload an image that is too large, it displays an exclamationcircle, this can either be hovered on, or clicked to display the error message. As soon as the user uploads an image that is the correct size, it disappears.

### DIFF
--- a/cms/src/components/images/ImageEditor.vue
+++ b/cms/src/components/images/ImageEditor.vue
@@ -1,7 +1,11 @@
 <script setup lang="ts">
 import { ref, computed, toRaw } from "vue";
 import LButton from "../button/LButton.vue";
-import { ArrowUpOnSquareIcon, QuestionMarkCircleIcon } from "@heroicons/vue/24/outline";
+import {
+    ArrowUpOnSquareIcon,
+    ExclamationCircleIcon,
+    QuestionMarkCircleIcon,
+} from "@heroicons/vue/24/outline";
 import ImageEditorThumbnail from "./ImageEditorThumbnail.vue";
 import { useNotificationStore } from "@/stores/notification";
 import {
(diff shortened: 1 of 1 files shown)


belljohan3 (commented):
Nice work @DirkDevelopmentFreelance

Just things:

- We need to really know if the max of image is correct
- I saw that for notification, you added two: The one from the store and the red icon before the help exclamationmark icon. i think that one would be enough.

DirkDevelopmentFreelance:
> * We need to really know if the max of image is correct
I agree. I just set it to 0.25mb(250000 bytes) temporarily for now as it was set no where else from what I could see. 
> * I saw that for notification, you added two: The one from the store and the red icon before the help exclamation mark icon. i think that one would be enough.
Yeah I was also wondering about the notification one(it was there) but we can remove it, but I feel like having both is good. But objectively it may be better to just have the inline notification. The red icon only appears when the user uploads a image > than the maxUploadSize, and then disappears when the user uploads a valid image.

belljohan3 (commented):
Just make sure to clean the code

belljohan3 (commented):
Hey @DirkDevelopmentFreelance !

I think we need this with @ivanslabbert. I am not entierely sure with the max size of the images.

DirkDevelopmentFreelance:
> I think we need this with @ivanslabbert. I am not entierely sure with the max size of the images.

The max size is set, it's 10mb. Just try to add a file that is bigger than 10mb.

belljohan3 (approved):
Just make to clean the code before merging

PR #379, opened 2024-10-09
add matomo analytics

oslabbie:


app/src/components/content/VideoPlayer.vue
```diff
@@ -88,6 +88,16 @@ onMounted(() => {
 
     // Get player user active states
     player.on(["useractive", "userinactive"], playerUserActiveEventHandler);
+
+    // start video player analytics on mounted
+    // @ts-expect-error window is a native browser api, and matomo is attaching _paq to window
+    if (window && window._paq) {
```
ivanslabbert, line 94: Strictly speaking you do not need to check for `window` as it is part of the browser API and always will be there.

ivanslabbert (changes requested):

ivanslabbert (approved):

PR #394, opened 2024-10-11
update vite config to copy the analytics service worker to dist

oslabbie:


oslabbie:
Hey @ivanslabbert 

Can you please review this. 
To be honest i am not sure if it will work, since i can only test if the file copies, but not if it will actually serve the file

PR #396, opened 2024-10-11
Create page document type

belljohan3:


shared/src/db/database.ts
```diff
@@ -181,20 +182,27 @@ class Database extends Dexie {
     /**
      * Get all IndexedDB documents of a certain type as Vue Ref
      * @param initialValue - The initial value of the ref while waiting for the query to complete
-     * @param tagType - Optional: The tag type to filter by (only used for tags)
+     * @param postOrTagType - Optional: The tag type to filter by (only used for tags)
```
ivanslabbert, line 185: The JSDoc descrition does not mention posts but only tags. It should mention both with the change to `postOrTagType`

ivanslabbert (changes requested):
Looks good @belljohan3 

Could you add tests to test the Post documents of postType "page" type specifically? I guess you can copy the Blog tests and modify them to test specifically for Page docs?

ivanslabbert (approved):

PR #404, opened 2024-10-15
Add Revert Changes Functionality to EditContent

DirkDevelopmentFreelance:


cms/src/components/content/EditContent.vue
```diff
@@ -184,6 +184,20 @@ const save = async () => {
...
+    // Restore the content documents to the previous versions
+    contentDocs.value = _.cloneDeep(contentDocsPrev.value!);
+
+    addNotification({
+        title: `${capitaliseFirstLetter(titleType)} reverted`,
+        description: `The changes to the ${titleType} have been reverted`,
+        state: "info",
+    });
+};
+
```
belljohan3, line 200: In this way, the notification will show even if there is changes. if you want to keep the notification, you can set differents messages according to the state of the content.

cms/src/components/content/EditContent.vue
```diff
@@ -230,10 +244,20 @@ watch(selectedLanguage, () => {
...
-
-            <LButton type="button" @click="save" data-test="save-button" variant="primary">
-                Save
-            </LButton>
+            <div class="flex gap-1">
+                <LButton
+                    type="button"
+                    @click="revertChanges"
+                    data-test="revert-changes-button"
+                    variant="tertiary"
```
belljohan3, line 252: How does the `secondary` attribute looks like ? 
I propose you to make the button title something like: `Revert Changes`

cms/src/components/content/EditContent.spec.ts
```diff
@@ -64,6 +64,38 @@ describe("EditContent.vue", () => {
...
+
+        await waitForExpect(() => {
+            const titleInput = wrapper.find('input[name="title"]');
+            expect((titleInput.element as HTMLInputElement).value).toBe(
+                mockData.mockEnglishContentDto.title,
+            );
+        });
+
+        const titleInput = wrapper.find('input[name="title"]');
+        await titleInput.setValue("New Title");
```
belljohan3, line 84: Can you try this: `await wrapper.find('input[name="title"]`).setValue("New Title");

belljohan3 (changes requested):
Nice job @DirkDevelopmentFreelance ! Just some few comments.

cms/src/components/content/EditContent.vue
DirkDevelopmentFreelance: The ```secondary``` attribute looks better.
In my opinion "Revert Changes" as to just "Revert" takes the attention away from "save". If you hover over the button it does say "Revert Changes" to hint to the user what the button will do. I think this will be enough to indicate to the user and keep it clean.

cms/src/components/content/EditContent.vue
DirkDevelopmentFreelance: Done.Good suggestion! It makes it much clearer to the user.

belljohan3 (approved):
What do you think if you make the notification with sucess state when they are changes to revert ?

PR #406, opened 2024-10-15
APP: Add appropriate message when no content is available in selected language.

DirkDevelopmentFreelance:


ivanslabbert (changes requested):
Nice, thanks Dirk!

Could you add a test to make sure we do not break this functionality if we make changes later?

ivanslabbert (approved):

PR #413, opened 2024-10-15
Add api rejection message in cms

DirkDevelopmentFreelance:


cms/src/App.vue
```diff
@@ -71,6 +72,17 @@ onBeforeMount(async () => {
...
         });
+
+        socket.on("changeRequestAck", (data: any) => {
+            if (data.ack == "rejected") {
+                useNotificationStore().addNotification({
+                    title: `Changed request: ${data.ack}`,
+                    description: data.message,
+                    state: "error",
+                    timer: 4000,
+                });
```
laurensgroeneveld, line 83: This does not yet fully comply with the acceptance criteria from the ticket:

> Create a (toast) notification with the error, and also state that the content has been reverted to its previous state.

I think we need to make sure this notification is understandable for non-technical users. "Change request" for example is also a concept that's foreign to those people, so I don't think it's helpful if we mention that.

Suggestion:

Title: Saving changes to server failed
Description: Your recent request to save changes has failed. The changes have been reverted, both in your CMS and on the server. Error message: `data.message`

Not sure if this is too long, but hopefully this doesn't happen too often in production anyways. I do think it's helpful to note the actual error as it's fairly readable in most cases, and can anyways then be screenshotted/forwarded to a developer if people have questions.

laurensgroeneveld (changes requested):
Have a comment on the notification message.

cms/src/App.vue
```diff
@@ -71,6 +72,17 @@ onBeforeMount(async () => {
...
             await loginRedirect();
         });
+
+        socket.on("changeRequestAck", (data: any) => {
+            if (data.ack == "rejected") {
+                useNotificationStore().addNotification({
+                    title: `Changed request: ${data.ack}`,
+                    description: data.message,
+                    state: "error",
+                    timer: 4000,
```
ivanslabbert, line 82: In error cases I think the error needs to stick around for a long time to give the user time to see it and take action (e.g. take a screenshot). I would set it to at least 1 minute.

ivanslabbert (changes requested):
Is it possible to add a test to test the error notifications? It is hard to test it in dev mode as these errors usually only will occur when there are bugs / version differences between the CMS and API (which should not happen if we do our job well).

In a test we can easily fake a response from the API through a test socket server.

cms/src/App.vue
DirkDevelopmentFreelance: Added

cms/src/App.vue
DirkDevelopmentFreelance: Changes made

DirkDevelopmentFreelance:
The test succeeds locally but not on github.

cms/src/App.spec.ts
```diff
@@ -49,4 +65,50 @@ describe("App", () => {
...
+        socket.on("changeRequestAck", changeRequestAckHandler);
+        changeRequestAckHandler({
+            ack: "rejected",
+            message: "Server error",
+        });
+
+        expect(addNotification).toHaveBeenCalledWith({
+            title: "Saving changes to server failed.",
+            description:
+                "Your recent request to save changes has failed. The changes have been reverted, both in your CMS and on the server. Error message: Server error",
```
ivanslabbert, line 109:
```suggestion
                "Your recent request to save changes has failed. The changes have been reverted. Error message: Server error",
```

cms/src/App.vue
```diff
@@ -71,6 +72,17 @@ onBeforeMount(async () => {
 
             await loginRedirect();
         });
+
+        socket.on("changeRequestAck", (data: any) => {
+            if (data.ack == "rejected") {
+                useNotificationStore().addNotification({
+                    title: "Saving changes to server failed.",
+                    description: `Your recent request to save changes has failed. The changes have been reverted, both in your CMS and on the server. Error message: ${data.message}`,
```
ivanslabbert, line 80:
```suggestion
                    description: `Your recent request to save changes has failed. The changes have been reverted. Error message: ${data.message}`,
```

ivanslabbert (approved):
Approved with 2 comments

PR #418, opened 2024-10-15
fix visual bugs of category view 

belljohan3:


DirkDevelopmentFreelance (commented):
It looks good, but maybe we should make the title wrap when it reaches the end of the div, when it wraps it can move up a bit in the div so it uses all the space it can so it doesn't make other elements look out of place?

app/src/pages/SingleContent.vue
```diff
@@ -207,6 +207,7 @@ function selectTag(parentId: Uuid) {
                 :tag="tag"
                 :queryOptions="{
                     filterOptions: { docType: DocType.Post },
+                    sortOptions: { sortBy: 'publishDate', sortOrder: 'desc' },
```
ivanslabbert, line 210: We need to sort in ascending order.

ivanslabbert (changes requested):
- [ ] Should we make the title wrap up to two lines, and if it needs more it should truncate the rest of the text?

image

ivanslabbert (approved):

PR #419, opened 2024-10-16
APP: Fix back button in mobile mode

DirkDevelopmentFreelance:


belljohan3 (commented):
.

app/src/components/navigation/TopBar.vue
```diff
@@ -1,17 +1,21 @@
 <script setup lang="ts">
 import { PlayCircleIcon, ChevronLeftIcon } from "@heroicons/vue/24/solid";
 import ProfileMenu from "./ProfileMenu.vue";
-import { useRoute, RouterLink } from "vue-router";
+import { useRoute } from "vue-router";
```
belljohan3, line 4:
```suggestion
import { useRouter } from "vue-router";
```

app/src/components/navigation/TopBar.vue
```diff
@@ -1,17 +1,21 @@
...
+    window.history.back();
+};
 </script>
 
 <template>
     <header>
         <div class="z-40 bg-zinc-100 dark:bg-slate-800">
             <div class="flex flex-row items-center justify-between space-x-8 px-6 py-5">
-                <RouterLink to="/" class="flex items-center">
+                <div @click="goBack" class="flex cursor-pointer items-center">
```
belljohan3, line 18:
```suggestion
                <div @click="router.back()" class="flex cursor-pointer items-center">
```

app/src/components/navigation/TopBar.vue
```diff
@@ -1,17 +1,21 @@
 <script setup lang="ts">
 import { PlayCircleIcon, ChevronLeftIcon } from "@heroicons/vue/24/solid";
 import ProfileMenu from "./ProfileMenu.vue";
-import { useRoute, RouterLink } from "vue-router";
+import { useRoute } from "vue-router";
 import DesktopMenu from "./DesktopMenu.vue";
 
 const route = useRoute();
```
belljohan3, line 7:
```suggestion
const router = useRouter();
```

belljohan3 (changes requested):
Nice job @DirkDevelopmentFreelance. Just few suggestions.

app/src/components/navigation/TopBar.vue
DirkDevelopmentFreelance: useRoute is still needed for some of the css, but I did instead use router instead of window.history.back() now, because useRouter() makes more sense here like you suggested.

app/src/components/navigation/TopBar.vue
DirkDevelopmentFreelance: Implemented the change, much cleaner code, nice suggestion.

app/src/components/navigation/TopBar.vue
DirkDevelopmentFreelance: useRoute is still needed for some of the css so that needs to be kept in, but I did instead use router instead of window.history.back() now, because useRouter() makes more sense here like you suggested.

ivanslabbert (approved):

belljohan3 (approved):

PR #420, opened 2024-10-16
Fix topic open from topics explorer should not show related heading

DirkDevelopmentFreelance:


belljohan3:
Hey @DirkDevelopmentFreelance ! It is still show the Related header when clicking on topic in Explore page. Could you have times today for reviewing it together ?

PR #422, opened 2024-10-16
Topic open from topics explorer should not show related heading

belljohan3:


ivanslabbert (changes requested):
It works as expected.

Not related to the task, but think we should fix it here:

- [x] The "Related" heading (when it should be showing) is a bit insignificant / small in Desktop mode. The colour could also be a shade darker (in light mode)?
image

- [x] When opening a topic, it shows the topic name twice. Think we need to be able to pass an optional property to the HorizontalScrollableTagViewer to hide the title?
image

ivanslabbert (approved):

PR #423, opened 2024-10-16
add code error for page and content

belljohan3:


belljohan3:
Not sure if this PR is still needed as it is already implemented.

PR #424, opened 2024-10-16
Remove change documents

oslabbie:
…he changeDoc's only slow down the api)

### DIFF
--- a/api/src/changeRequests/processChangeRequest.ts
+++ b/api/src/changeRequests/processChangeRequest.ts
@@ -1,7 +1,5 @@
-// import { randomUUID } from "crypto";
 import { validateChangeRequest } from "./validateChangeRequest";
 import { DbService } from "../db/db.service";
-// import { ChangeDto } from "../dto/ChangeDto";
 import { ChangeReqDto } from "../dto/ChangeReqDto";
 import { DocType, Uuid } from "../enums";
 import { validateSlug } from "./validateSlug";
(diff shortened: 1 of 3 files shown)


api/src/changeRequests/processChangeRequest.ts
```diff
@@ -80,37 +80,38 @@ export async function processChangeRequest(
     // Insert / update the document in the database
     const upsertResult = await db.upsertDoc(doc);
 
+    // ******************* tmp disable changeDoc, until it will be implimented on the CMS (ATM the changeDoc's only slow down the api) ********************
```
laurensgroeneveld, line 83: As for this comment, can you link to a Github issue that will change this back? Then it's easy in the future to see if this comment is still relevant. "tmp" might not be so temporary 😅 Also I don't think it needs to be so extensive and attention drawing but that might just be personal preference 😄

```suggestion
    // TODO: Reactivate change diffs in change requests - https://github.com/bccsa/luminary/issues/...
```

api/src/changeRequests/processChangeRequest.ts
```diff
@@ -80,37 +80,38 @@ export async function processChangeRequest(
...
-        }
-
-        // Apply the language of the included document to the change document
-        if (upsertResult.changes.language) {
-            changeDoc.language = upsertResult.changes.language;
-        }
-
-        return db.insertDoc(changeDoc);
-    }
+    // if (upsertResult.changes) {
```
laurensgroeneveld, line 85: Related to the previous comment, how sure are we that this will be changed soon? If not, I would suggest to remove this code rather than comment it out, so it doesn't keep laying around if we never get to actually solving this problem. We can always easily find the code back in Git/Github, especially if you create an issue for solving this that then can link to this PR.

laurensgroeneveld (changes requested):
Be sure to read all the comments as the last one is perhaps the most important. I would actually prefer removing code to commenting it out, as we can find it back later (but be sure to reference the code in a new issue on the backlog).

oslabbie:
Hey @laurensgroeneveld 

Thanx for the review, i made the changes you suggested. 
I removed the test instead of just skipping them though, since we an any case removed the code that the test is related to

laurensgroeneveld (approved):
image

This always looks great, less code is less bugs 😁

PR #426, opened 2024-10-17
remove hard-coded language membership.

belljohan3:


cms/src/components/languages/CreateOrEditLanguageModel.spec.ts
```diff
@@ -80,22 +90,6 @@ describe("CreateOrEditLanguageModal.vue", () => {
             // check if the button has the right text
             expect(wrapper.html()).toContain("Save Changes");
         });
-
-        it("disables save button if form fields are not filled", async () => {
```
DirkDevelopmentFreelance, line 84: Just realised this test was not a duplicate, this test tests if the button becomes disables if all the other values are empty, so you need to check if disabled is truthy

DirkDevelopmentFreelance (changes requested):

DirkDevelopmentFreelance (commented):
Everything looks good to me.

cms/src/components/languages/CreateOrEditLanguageModel.spec.ts
```diff
@@ -80,22 +90,22 @@ describe("CreateOrEditLanguageModal.vue", () => {
...
         });
+    });
 
-        it("disables save button if form fields are not filled", async () => {
-            const wrapper = mount(CreateOrEditLanguageModal, {
-                props: {
-                    isVisible: true,
-                },
-            });
+    it.skip("disables save button if form fields are not filled", async () => {
```
laurensgroeneveld, line 95: Why is this test skipped? If there is a very good reason for it, add a comment, but otherwise rather fix the reason why it needs to be skipped 😅

cms/src/components/languages/CreateOrEditLanguageModel.spec.ts
```diff
@@ -52,13 +55,20 @@ describe("CreateOrEditLanguageModal.vue", () => {
...
+                groupSelector = wrapper.findComponent(GroupSelector);
+                expect(groupSelector.exists()).toBe(true);
+            });
+
+            await groupSelector!.findComponent(ComboboxInput).setValue("Languages");
+
             // Assert that the save button is disabled
             const saveButton = wrapper.find("[data-test='save-button']");
-            expect(saveButton.attributes("disabled")).toBeUndefined();
+            expect(saveButton.attributes("disabled")).toBeFalsy();
```
laurensgroeneveld, line 71: Not a fan of "falsy" as it doesn't check for an actual value - it could be `false`, `null`, `0`, etc. So I would rather put it as `toBeFalse` (or if that doesn't exist, `toBe(false)`).

laurensgroeneveld (dismissed):
Just two comments on the tests.

cms/src/components/languages/CreateOrEditLanguageModel.spec.ts
ivanslabbert: @DirkDevelopmentFreelance could you help to Johan to get the test working? You found a way to get tests to work with HeadlessUI components?

ivanslabbert (changes requested):

ivanslabbert:
Rewriting the combobox control in #443. Please create a new task to fix the test as soon as the new combobox control is ready.

cms/src/components/languages/CreateOrEditLanguageModel.spec.ts
ivanslabbert: Discussed today that one can test if the disabled prop is set on the LButton component.

ivanslabbert (approved):

DirkDevelopmentFreelance (approved):

PR #427, opened 2024-10-17
417 add seo title and summary fields

belljohan3:


laurensgroeneveld (commented):
This works fine, however from a UX perspective I think it might need a bit more work:

- When a specific SEO title/string are not set, the app will probably use the title/summary of the content. Currently, they are empty by default, leading me as a user to believe it's not set at all:

image

- At the same time I feel a lot of editors might actually not have good insight into what SEO is. It's also not relevant for all types of post - for internal content it doesn't matter. But it does take up quite a bit of space in the interface.

To solve this requires a little bit of thinking. I'm thinking about two things:
- Show that by default the translation title/summary will be taken. Either by prefilling them (although that might also not make sense because `seoTitle`/`seoString` are not set by default I think), or having some explanation text?
- Make the SEO less prominent in the interface. Perhaps we can use tabs within "Basic translation settings", with a separate tab for SEO?

Just some thoughts, I think it would be good to discuss this with Ivan as well.

belljohan3:
Hey @laurensgroeneveld ! 

I think you are right. 
Check out this [comment](https://github.com/bccsa/luminary/issues/432#issuecomment-2425158353). According to me, we can do it like that. It will simplify things.

ivanslabbert:
> Hey @laurensgroeneveld !
> 
> I think you are right. Check out this [comment](https://github.com/bccsa/luminary/issues/432#issuecomment-2425158353). According to me, we can do it like that. It will simplify things.

I commented to this here: https://github.com/bccsa/luminary/issues/432#issuecomment-2425185501
In short, we need both options.

belljohan3:
Hey @laurensgroeneveld. I went for this: 

image

laurensgroeneveld:
> Hey @laurensgroeneveld. I went for this:
> 
> image

I'll try to review it functionally when I have time (just answering this quickly in between work tasks 😅) but that was what I had in mind, looks good 👍

cms/src/components/common/LCard.vue
```diff
@@ -53,7 +63,34 @@ function collapse() {
             </button>
         </div>
         <div v-show="!collapsed" data-test="collapsible-container">
+            <!-- Tab Navigation -->
+            <div v-if="tabs?.length" class="border-b">
```
laurensgroeneveld, line 67: I think the tabs should rather be a separate component. The `LCard` doesn't need to be aware of any tabs, that can just be in the slot that gets passed into it. There is already a `LTabs` component in the codebase, but it's a bit half baked as it doesn't fully support switching tabs. It should be fairly easy to add that though, as it's already rendering slots. You can view the component in the [sandbox](http://localhost:4175/sandbox)

```html
                <LTabs :tabs="tabs" v-model:currentTab="selectedTab">
                    <template #tab-1>
                        Tab 1
                    </template>               
                    <template #tab-2>
                        Tab 2
                    </template>
                </LTabs>
```

-> this currently renders both tabs, but I think it will be pretty easy to add the correct functionality inside `LTabs`.
image

Please also add tests for this component.

cms/src/components/content/EditContentBasic.vue
```diff
@@ -1,133 +1,47 @@
...
-                @click="startEditingSlug"
-                class="flex h-5 w-5 min-w-5 items-center justify-center rounded-md py-0.5 hover:bg-zinc-200 active:bg-zinc-300"
-                title="Edit slug"
-            >
-                <component :is="PencilIcon" class="h-4 w-4 text-zinc-500" />
-            </button>
+    <LCard title="General Settings" collapsible v-if="content">
+        <!-- Tab Navigation -->
+        <div>
+            <nav class="flex space-x-4">
```
laurensgroeneveld, line 26: This is the correct place for the tabs instead of inside `LCard`, but just use the `LTabs` component for it (see my previous comment).

cms/src/components/content/EditContentBasic.vue
```diff
@@ -1,133 +1,47 @@
...
+const content = defineModel<ContentDto>("content");
 
-const validateSlug = async () => {
-    if (!content.value) return;
-    content.value.slug = await Slug.generate(content.value.slug, content.value._id || "");
-};
+const activeTab = ref(0);
+const tabs = [
+    { title: "General ", component: ContentBasic },
+    { title: "SEO ", component: ContentSEO },
```
laurensgroeneveld, line 18: I would keep the naming with the prefix `Edit` so all edit components are grouped together. Perhaps rename this component `EditContentGeneral`, because that is the title of it?

- `EditContentGeneral`
    -`EditContentBasic`
    - `EditContentSeo` (<- please don't capitalize SEO, we follow camel case even there).

laurensgroeneveld (changes requested):

ivanslabbert (changes requested):
Blocking merging until reviewed by Dirk and checked by Ivan

DirkDevelopmentFreelance (commented):
The code looks good and the interface in the cms.

I just have 1 suggestion to do with this topic, maybe the SEO title and summary should be an array of titles and summaries for the current content(the interface can then be something where there is a text area and titles/summaries can be entered and separated by a ``;``, because then editors can have more thorough coverage on alternate things users can type in that can be relevant. Hope this makes sense.

cms/src/components/content/EditContentGeneral.vue
```diff
@@ -0,0 +1,37 @@
...
+type Props = {
+    disabled: boolean;
+};
+defineProps<Props>();
+
+const content = defineModel<ContentDto>("content");
+
+const currentTab = ref("basic"); // Default tab key
+const tabs = [
+    { title: "Basic", key: "basic", component: EditContentBasic },
```
ivanslabbert, line 18:
```suggestion
    { title: "Visible title & summary", key: "visible", component: EditContentBasic },
```

cms/src/components/content/EditContentGeneral.vue
```diff
@@ -0,0 +1,37 @@
...
+    { title: "Seo", key: "seo", component: EditContentSeo },
```
ivanslabbert, line 19:
```suggestion
    { title: "SEO title & summary", key: "seo", component: EditContentSeo },
```

cms/src/components/content/EditContentSeo.vue
```diff
@@ -0,0 +1,33 @@
+<script setup lang="ts">
```
ivanslabbert, line 1: Do we really need a separate component here? Could it not be included in EditContentGeneral (and perhaps rename EditContentGeneral to EditContentTitleSummary)?

ivanslabbert (changes requested):

cms/src/components/content/EditContentTitleSummary.vue
```diff
@@ -0,0 +1,58 @@
+<script setup lang="ts">
```
ivanslabbert, line 1: Can you move the logic and HTML in this file to EditContentBasic.vue?

cms/src/components/content/EditContentTitleSummary.spec.ts
```diff
@@ -0,0 +1,61 @@
+import "fake-indexeddb/auto";
```
ivanslabbert, line 1: Can you move the tests in this file to EditContentBasic.spec.ts?

ivanslabbert (changes requested):

ivanslabbert (approved):

PR #436, opened 2024-10-19
APP: Show 404 error when content is expired / scheduled but available in IndexedDb, or when not available in IndexedDb

belljohan3:


app/src/pages/SingleContent.vue
```diff
@@ -141,11 +149,10 @@ function selectTag(parentId: Uuid) {
         </div>
     </div>
 
-    <div v-if="isLoading">
-        <LoadingSpinner />
-    </div>
+    <NotFoundPage
+        v-if="isContentNotFound || isExpiredOrScheduled || content.status == PublishStatus.Draft"
+    />
```
laurensgroeneveld, line 154: From just a quick glance at this: doesn't this mean their could be a short flash of "404" before the content is shown? I haven't tested this yet though.

ivanslabbert: This should not happen under normal circumstances as content should already be in IndexedDB when routing from from the home page.

In cases where people open a link that is routing to the article without having the document in IndexedDB yet, this might happen and we will need a loading state. But I am not sure how we would be able to determine that without having a live API which can query the API directly, so perhaps we need to add a follow-up task to look at this once we have a live API available.

ivanslabbert (commented):
We need a 404 HTTP response when building the site to static pages (which we will do sometime soon with Vite for the public website). @belljohan3 could you research how one would implement that with Vue?

ivanslabbert (changes requested):
Please also add tests to test that scheduled and expired content gives a 404?

ivanslabbert (approved):

belljohan3:
can i merge this @ivanslabbert @DirkDevelopmentFreelance ?

PR #437, opened 2024-10-19
Solve some security warnings using npm audit fix

laurensgroeneveld:
This solves some of the open security warnings on all projects using `npm audit fix`. 

On both the app and CMS the only moderate warning is related to the `vue-tsc` package, of which we're using an older version. However, when I upgrade this from 1.8 to 2.1, it suddenly generates tons of type errors in the code, that don't actually seem to be correct. Not entirely sure what's going on here and that needs a bit more investigation.

### DIFF
--- a/api/package-lock.json
+++ b/api/package-lock.json
@@ -2358,10 +2358,11 @@
             }
         },
         "node_modules/@nestjs/core": {
-            "version": "10.4.4",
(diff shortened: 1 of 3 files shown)


laurensgroeneveld:
Hmm this seems to break `npm ci --install-links`. `npm update` on a clean `main` version has the same result. Don't have the bandwidth right now to investigate, so I'll close this for now.

PR #449, opened 2024-10-22
CMS: Make EditContent saveable when in draft without img

DirkDevelopmentFreelance:


laurensgroeneveld (approved):
Works fine, nice fix 👍 I see the unit tests for this component are already skipped because of https://github.com/bccsa/luminary/issues/313 so I won't complain about it lacking a test. Added a note to that ticket with a link to this PR.

PR #456, opened 2024-10-23
add seoTitle and seoSummary as meta data

belljohan3:


ivanslabbert (changes requested):
Blocking merging until reviewed

DirkDevelopmentFreelance (commented):
The code looks nice and clean, nice research done. @ivanslabbert, just double check the functionality as I am not entirely sure how to find the metadata inside the browser, tried an extension but it doesn't work on local sites. Nice Work @belljohan3

belljohan3:
> The code looks nice and clean, nice research done. @ivanslabbert, just double check the functionality as I am not entirely sure how to find the metadata inside the browser, tried an extension but it doesn't work on local sites. Nice Work @belljohan3

You can inspect the current doc and check the meta data in the `head` tag.

app/src/pages/SingleContent.vue
```diff
@@ -15,6 +15,7 @@ import RelatedContent from "../components/content/RelatedContent.vue";
 import VerticalTagViewer from "@/components/tags/VerticalTagViewer.vue";
 import Link from "@tiptap/extension-link";
 import LImage from "@/components/images/LImage.vue";
+import { useHead } from "@vueuse/head";
```
ivanslabbert, line 18: According to the plugin's GitHub page, it has been replaced with Unhead/vue, so the best will be to use the new package.

But to keep our application as lightweight as possible, I would suggest not using a plugin to set the title and metadata, but rather set the title as currently done in line 44, and setting the metadata as per https://stackoverflow.com/questions/2568760/is-it-possible-to-use-javascript-to-change-the-meta-tags-of-the-page

app/src/pages/SingleContent.vue
```diff
@@ -37,32 +38,38 @@ const isExpiredOrScheduled = computed(() => {
...
 
-        if (isExpiredOrScheduled.value) return;
+    if (isExpiredOrScheduled.value) return;
 
-        const tagIds = content.value.parentTags.concat([content.value.parentId]); // Include this content's parent ID to include content tagged with the parent (if the parent is a tag document).
+    useHead({
+        title: content.value.seoTitle || content.value.title,
+        meta: [
+            {
+                name: "summary",
```
ivanslabbert, line 54:
```suggestion
                name: "description",
```
the "summary" metadata tag is not supported by Google's indexer

app/src/pages/SingleContent.vue
```diff
@@ -37,32 +38,38 @@ const isExpiredOrScheduled = computed(() => {
...
-        if (!content.value) return;
+watch(content, async () => {
+    if (!content.value) return;
 
-        document.title = isExpiredOrScheduled.value
-            ? `Page not found - ${appName}`
-            : `${content.value.title} - ${appName}`;
+    document.title = isExpiredOrScheduled.value
+        ? `Page not found - ${appName}`
+        : `${content.value.title} - ${appName}`;
```
ivanslabbert, line 46: We are now setting the title in two locations in the code, both direcetly using the "document" api as well as through the useHead plugin. We should combine the logic.

ivanslabbert (changes requested):

belljohan3:
@ivanslabbert  Here is also a nice way to set it. 

`[
    { name: "title", content: content.value.seoTitle || content.value.title || "" },
    { name: "description", content: content.value.seoString || content.value.summary || "" },
].forEach(attrs => {
    const metaTag = Object.assign(document.createElement("meta"), attrs);
    document.head.appendChild(metaTag);
});`

app/src/pages/SingleContent.vue
```diff
@@ -37,32 +37,37 @@ const isExpiredOrScheduled = computed(() => {
...
 
-        if (isExpiredOrScheduled.value) return;
+    // Seo meta tag settings
+    [
+        { name: "title", content: content.value.seoTitle || content.value.title || "" },
```
ivanslabbert, line 49: Google does not support the "title" metadata (see https://developers.google.com/search/docs/crawling-indexing/special-tags). We could rather set the document.title to the SEO title if available (in line 43)

Your implementation for the description metadata tag looks good.

ivanslabbert (changes requested):
Added 1 change request.

Can we also add a test to see if the title is being set correctly at the <head> title tag, and if the "description" metadata is set correctly?

ivanslabbert (approved):

PR #460, opened 2024-10-25
get historical data from newly accesibble groups

belljohan3:


api/src/test/socketioTestClient.ts
```diff
@@ -9,6 +9,7 @@ export type socketioTestRequest = {
     changeRequest?: ChangeReqDto;
     timeout?: number;
     getAccessMap?: boolean;
+    accessMap?: any;
```
ivanslabbert, line 12: Can you add a comment why use need to set this to `any`?

api/src/test/mockdata.ts
```diff
@@ -0,0 +1,227 @@
+// import { AccessMap } from "src/permissions/permissions.service";
+
+export const superAdminAccessMap = {
```
ivanslabbert, line 3: This is not used, so the file can probably be deleted

api/src/socketio.spec.ts
```diff
@@ -288,6 +288,45 @@ describe("Socketio", () => {
...
+                            tag: {
+                                view: true,
+                            },
+                            language: {
+                                view: true,
+                            },
+                        },
+                    };
+
+                    // Indicate to the API that the client only have access to public content
```
ivanslabbert, line 318:
```suggestion
                    // Indicate to the API that the client only had access to public content
```

ivanslabbert (approved):
Approved with some small issues to be fixed

api/src/test/socketioTestClient.ts
```diff
@@ -9,6 +9,7 @@ export type socketioTestRequest = {
     changeRequest?: ChangeReqDto;
     timeout?: number;
     getAccessMap?: boolean;
+    accessMap?: any; // Using 'any' here to accommodate varied structures for permissions data, which may differ based on client requirements or evolve over time.
```
ivanslabbert, line 12: This sounds like AI...
```suggestion
    accessMap?: any; // The accessMap is converted to a AccessMap object by the API.
```

ivanslabbert (approved):

PR #465, opened 2024-10-28
Fix preferred language in APP

DirkDevelopmentFreelance:


app/src/globalConfig.ts
```diff
@@ -27,13 +27,21 @@ export const appLanguageAsRef = readonly(_appLanguageAsRef);
...
             newVal.length > 0 &&
             (!appLanguageIdAsRef.value || !newVal.some((l) => l._id === appLanguageIdAsRef.value))
         ) {
-            appLanguageIdAsRef.value = newVal[0]._id;
+            const languagePreferredByBrowser = navigator.language;
+            newVal.forEach((val) => {
+                if (val.languageCode == languagePreferredByBrowser) {
+                    appLanguageIdAsRef.value = val._id;
+                } else {
+                    appLanguageIdAsRef.value = "lang-eng";
```
ivanslabbert, line 42: We should not hardcode references to documents which may be deleted. I suggest adding a field to LanguageDto which can mark a language document as default, and then finding the default document instead of hardcoding it.

This obviously requires some API and CMS work as well...

app/src/globalConfig.ts
```diff
@@ -27,13 +27,21 @@ export const appLanguageAsRef = readonly(_appLanguageAsRef);
...
-    // Set the preferred language to the first language in the list if it is not set
+    // Set the preferred language to the preferred language returned by the browser if it is not set
+    // The language is only set if there is a supported language for it otherwise it defaults to english
     watch(languages, (newVal) => {
         if (
             newVal.length > 0 &&
             (!appLanguageIdAsRef.value || !newVal.some((l) => l._id === appLanguageIdAsRef.value))
         ) {
-            appLanguageIdAsRef.value = newVal[0]._id;
+            const languagePreferredByBrowser = navigator.language;
```
ivanslabbert, line 37: There should be a list of preferred browser languages? We should try to get a match to any of those languages, and only if none are matching revert to the default language.

ivanslabbert (changes requested):

app/src/globalConfig.ts
DirkDevelopmentFreelance: Yeah that makes sense. Sorry I didn't think of that scenario

cms/src/components/languages/CreateOrEditLanguageModel.spec.ts
```diff
@@ -4,7 +4,7 @@ import { mount } from "@vue/test-utils";
 import CreateOrEditLanguageModal from "./CreateOrEditLanguageModal.vue";
 import { setActivePinia } from "pinia";
 import { createTestingPinia } from "@pinia/testing";
-import { accessMap, db } from "luminary-shared";
+import { accessMap, db, DocType, type LanguageDto } from "luminary-shared";
```
ivanslabbert, line 7: Lint warning: Unused import `LanguageDto`

cms/src/components/languages/CreateOrEditLanguageModal.vue
```diff
@@ -5,6 +5,8 @@ import LInput from "@/components/forms/LInput.vue";
 import LButton from "@/components/button/LButton.vue";
```
ivanslabbert, line 5: - Are we ensuring that controls are disabled for users without edit access to the given language?
- Are we ensuring that setting a language to Default is disabled for users without language edit access to all languages?

The same checks should also be implemented in the API.
The API should also remove the "default" field from other language documents if set on a given doc. (CMS checks are only best effort but cannot be relied on - the API does the final check and quality control)

ivanslabbert (commented):
Good work so far @DirkDevelopmentFreelance 

Left some comments

cms/src/components/languages/CreateOrEditLanguageModal.vue
```diff
@@ -22,29 +24,37 @@ const emit = defineEmits(["close", "created", "updated"]);
 // Check if we are in edit mode (if a language is passed)
 const isEditMode = computed(() => !!props.language);
 
+const canEditOrCreate =
+    verifyAccess(["group-private-editors"], DocType.Language, AclPermission.Edit) ||
+    verifyAccess(["group-private-editors"], DocType.Language, AclPermission.Edit);
```
ivanslabbert, line 29: We should never use hard-coded references to data that might not be available in the database in production code. (In tests it is obviously ok and needed.)

cms/src/components/languages/CreateOrEditLanguageModel.spec.ts
```diff
@@ -14,6 +14,7 @@ import {
 import waitForExpect from "wait-for-expect";
 import GroupSelector from "../groups/GroupSelector.vue";
 import { ComboboxInput } from "@headlessui/vue";
+import LToggle from "../forms/LToggle.vue";
```
ivanslabbert, line 17: lint warning

ivanslabbert (changes requested):
Did not do a functional check (running in my dev env) but added some comments

ivanslabbert (approved):
Add tests for: (can do in a new ticket)

- GroupSelector.vue: The list of assignable groups should only include groups to which the user has group assign access, and edit access to the docType in question (passed as prop)
- API validateChangeRequestAccess: Add tests for Language doctype validations
- API processChangerequest: Check if the API can reset the default language flag (see line 134) - if not implemented already

PR #469, opened 2024-10-28
App homepage query caching

ivanslabbert:


app/src/pages/HomePage/HomePageNewest.vue
laurensgroeneveld, line 1: The convention I tried to have is that all components in `pages` are top-level page components, i.e. they get rendered directly in the `RouterView`. These subcomponents for Homepage should then rather belong in `components`.

laurensgroeneveld (approved):
Just one small comment, but looks good otherwise 👍 I don't have enough content in my local db to test this properly, but I trust you did 😄

app/src/pages/HomePage/HomePageNewest.vue
ivanslabbert: Sure, thanks. I can move them.

ivanslabbert:
> Just one small comment, but looks good otherwise 👍 I don't have enough content in my local db to test this properly, but I trust you did 😄

I copied the staging database to my local dev environment to do testing and got good results (using the performance slowdown settings in chrome's dev tools). But will be nice to test on staging in a production environment.

PR #472, opened 2024-10-30
feat: add access control to disable save on published content for users without publish permissions

belljohan3:
… publish permissions

### DIFF
--- a/cms/src/components/content/EditContent.vue
+++ b/cms/src/components/content/EditContent.vue
@@ -171,7 +171,7 @@ const save = async () => {
     if (!canPublish.value) {
         addNotification({
             title: "Insufficient Permissions",
-            description: "You do not have permission to save published content",
+            description: "You do not have publish permissions",
             state: "error",
         });
         return;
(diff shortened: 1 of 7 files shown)


cms/src/components/content/EditContent.vue
```diff
@@ -164,6 +168,15 @@ const save = async () => {
         return;
     }
 
+    if (!canPublish.value) {
+        addNotification({
+            title: "Insufficient Permissions",
+            description: "You do not have permission to save published content",
```
ivanslabbert, line 174:
```suggestion
            description: "You do not have "publish" permissions",
```

ivanslabbert (changes requested):
I think when a user (without publish permissions) opens a document where the status is already set to published, he should not be able to edit any of the fields in the document.

The validation changes in this PR looks ok and we can keep them.

ivanslabbert:
This is not working as expected. In the screenshot below the user does not have publish access to the document. As the document is already published, all fields should be disabled.

telegram-cloud-photo-size-4-5861455488372294123-y

PR #473, opened 2024-10-30
add acl entry validation in api

belljohan3:


api/src/enums.ts
```diff
@@ -18,6 +18,7 @@ export enum DocType {
     Language = "language",
     Group = "group",
     Change = "change",
+    Image = "image",
```
ivanslabbert, line 21: We do not have image documents anymore, so this line can be removed.

ivanslabbert (changes requested):
The new functions need to be implemented in the change request processing chain

api/src/changeRequests/validateChangeRequest.ts
```diff
@@ -62,6 +64,15 @@ export async function validateChangeRequest(
         return validationResult;
     }
 
+    // check ACL entries for group documents
+    if (changeRequest.doc.type === DocType.Group) {
+        validateAclEntry(changeRequest.doc, new GroupAclEntryDto());
+        return {
```
ivanslabbert, line 70: This will always fail validation. I think you need to check the output of the validateAclEntry function?

ivanslabbert:
Wait until #471 is merged before finalizing this PR. The new Redirect document type should be added to the API side ACL validation before merging.

ivanslabbert (approved):

PR #484, opened 2024-11-04
Optimize explore page query

ivanslabbert:


DirkDevelopmentFreelance (approved):
Looks good, explore page now loads faster on 20x throttle on my side.

PR #485, opened 2024-11-05
removed capacitor packages, and added a event listener for the video …

oslabbie:
…player, in order to trigger capacitor actions from plugins on the player

### DIFF
--- a/app/src/components/content/VideoPlayer.vue
+++ b/app/src/components/content/VideoPlayer.vue
@@ -8,8 +8,6 @@ import type { ContentDto } from "luminary-shared";
 import px from "./px.png";
 import LImage from "../images/LImage.vue";
 import { getMediaProgress, removeMediaProgress, setMediaProgress } from "@/globalConfig";
-import { StatusBar } from "@capacitor/status-bar";
-import { Capacitor } from "@capacitor/core";
 
 type Props = {
     content: ContentDto;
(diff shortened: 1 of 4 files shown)


app/src/components/content/VideoPlayer.vue
```diff
@@ -73,6 +64,10 @@ onMounted(() => {
 
     player = videojs(playerElement.value!, options);
 
+    // emit event with player on mount
```
ivanslabbert, line 67:
```suggestion
    // emit event with player on mount to notify subscribing plugins that the player has been created
```

app/src/components/content/VideoPlayer.vue
```diff
@@ -137,7 +129,9 @@ onUnmounted(() => {
     player?.off(["mousemove", "click"], autoHidePlayerControls);
     player?.off("play", playerPlayEventHandler);
     player?.off(["useractive", "userinactive"], playerUserActiveEventHandler);
-    player?.off("fullscreenchange", playerFullscreenEventHandler);
+    // emit event when player is Destroyed
```
ivanslabbert, line 132:
```suggestion
    // emit event to notify subscribing plugins that the player has been destroyed
```

ivanslabbert (approved):
Approved with two suggestions to change comments

PR #498, opened 2024-11-07
add error handling to plugin loader

oslabbie:


app/src/util/pluginLoader.ts
```diff
@@ -3,16 +3,20 @@
...
-    });
+        const _a: any = [];
+        _p.forEach(async (p) => {
+            _a.push(dynamicLoadPlugin(p));
+        });
 
-    await Promise.all(_a);
+        await Promise.all(_a);
+    } catch (err: any) {
+        console.log(err.message);
```
ivanslabbert, line 18: Change to console.error so that we can catch the errors with Sentry (https://stackoverflow.com/questions/50633580/how-to-report-console-error-with-sentry)

ivanslabbert (approved):

PR #502, opened 2024-11-09
disabled image field when no edit access

belljohan3:


cms/src/components/images/ImageEditorThumbnail.vue
```diff
@@ -8,6 +8,7 @@ import fallbackImage from "@/assets/fallback-image-cms.webp";
 type Props = {
     imageFileCollection?: ImageFileCollectionDto;
     imageUploadData?: ImageUploadDto;
+    editable?: boolean;
```
ivanslabbert, line 11: Please rename to `disabled` to keep the naming convention.
(The name `editable` is used elsewhere in the code to name DTO objects that can be edited)

ivanslabbert (approved):
Approved, but please have a look at a couple comments?

cms/src/components/images/ImageEditorThumbnail.vue
belljohan3: Thanks.

PR #503, opened 2024-11-09
Disallow duplicate redirects

DirkDevelopmentFreelance:


api/src/db/db.service.ts
```diff
@@ -473,15 +473,38 @@ export class DbService extends EventEmitter {
         });
     }
 
+    getDocsByType(docType: DocType): Promise<DbQueryResult> {
```
ivanslabbert, line 476: Please add JSDoc documentation for this function?

cms/src/util/slug.ts
```diff
@@ -38,8 +38,18 @@ export class Slug {
...
+
+        if (docType) {
+            existingDocWithSlug = await db.docs
+                .where("slug")
+                .equals(text)
+                .and((doc) => doc.type === docType)
+                .first();
+        } else {
+            existingDocWithSlug = await db.docs.where("slug").equals(text).first();
+        }
```
ivanslabbert, line 52:
```suggestion
let query = db.docs.where("slug").equals(text);
if (docType) query = query.and((doc) => doc.type === docType);
const res = await query.first();
```

ivanslabbert (approved):

PR #507, opened 2024-11-11
Add configurable favicon

DirkDevelopmentFreelance:


cms/.env.example
```diff
@@ -5,6 +5,8 @@ VITE_CLIENT_APP_URL="http://localhost:4174"
 VITE_CLIENT_IMAGES_URL="http://localhost:9000/images"
 VITE_LOGO="../src/assets/logo.svg"
 
+VITE_FAV_ICON="/favicon.png"
```
ivanslabbert, line 8: Can we rename this to `VITE_LOGO_FAVICON`? Then it aligns with the logo naming. Perhaps also move the favicon to the `/src/assets/ directory`.

ivanslabbert (approved):
I added new favicon logos to the`asset` folders in the CMS and app, which can be used as the new favicon logos. Please see some comments on naming?

PR #513, opened 2024-11-12
Add LModal Component

DirkDevelopmentFreelance:


app/src/components/form/LModal.vue
```diff
@@ -0,0 +1,64 @@
...
+
+type Props = {
+    isVisible: boolean;
+    heading: string;
+    bLeftText?: string;
+    bRightText?: string;
+};
+defineProps<Props>();
+
+const emit = defineEmits(["close", "b-left-click", "b-right-click"]);
```
ivanslabbert, line 12: If we remove all buttons from the footer are (see next comment) we only need the "close" event (which the developer can choose to implement or not)

app/src/components/form/LModal.vue
```diff
@@ -0,0 +1,64 @@
...
+                    <LButton
+                        name="close-button"
+                        :variant="!bLeftText ? 'primary' : 'tertiary'"
+                        size="lg"
+                        rounding="less"
+                        :class="bLeftText || bRightText ? 'ml-auto w-1/4' : 'w-full'"
+                        @click="emit('close')"
+                    >
+                        Close
+                    </LButton>
```
ivanslabbert, line 59: What if we remove all buttons, but rather expose the footer section as a named slot? In this way a developer can add anything into the slot and hook up to events on the buttons directly (without having to interact with the modal's logic).

app/src/components/form/LModal.vue
```diff
@@ -0,0 +1,64 @@
...
+    bLeftText?: string;
+    bRightText?: string;
+};
+defineProps<Props>();
+
+const emit = defineEmits(["close", "b-left-click", "b-right-click"]);
+</script>
+
+<template>
+    <div :open="isVisible" @close="emit('close')">
```
ivanslabbert, line 16: This is nice - giving the possibility to close the modal by clicking outside of the popup area.

ivanslabbert (commented):
For a modal to work, it needs to be teleported to a higher-level element. Vue has support for teleporting. Otherwise, the modal will be contained inside any relative or absolute parent element.

app/src/components/form/LModal.vue
```diff
@@ -0,0 +1,28 @@
...
+        <div v-if="isVisible" @click.self="emit('close')">
+            <div class="fixed inset-0 z-50 bg-slate-800 bg-opacity-50 backdrop-blur-sm"></div>
+            <div class="fixed inset-0 z-50 flex items-center justify-center rounded-lg p-2">
+                <div class="w-full max-w-md rounded-lg bg-white p-5 shadow-xl dark:bg-slate-700">
+                    <h2 class="mb-4 text-lg font-semibold">{{ heading }}</h2>
+                    <div class="divide-y divide-zinc-200 dark:divide-slate-600">
+                        <slot name="header"></slot>
+                    </div>
+                    <div class="mt-4">
+                        <slot name="footer">Default content for footer</slot>
```
ivanslabbert, line 22: Shall we leave this empty (i.e. no default text). If a developer does not want to use the footer, he needs to add content to it at the moment to override the default text (as far as I can see how it will work...)

app/src/components/form/LModal.vue
```diff
@@ -0,0 +1,28 @@
...
+
+<template>
+    <Teleport to="body">
+        <div v-if="isVisible" @click.self="emit('close')">
+            <div class="fixed inset-0 z-50 bg-slate-800 bg-opacity-50 backdrop-blur-sm"></div>
+            <div class="fixed inset-0 z-50 flex items-center justify-center rounded-lg p-2">
+                <div class="w-full max-w-md rounded-lg bg-white p-5 shadow-xl dark:bg-slate-700">
+                    <h2 class="mb-4 text-lg font-semibold">{{ heading }}</h2>
+                    <div class="divide-y divide-zinc-200 dark:divide-slate-600">
+                        <slot name="header"></slot>
```
ivanslabbert, line 19: Is this the default slot? (From the name it looks like it belongs to the area where the heading text is, but from the HTML it looks like the "body" part of the modal. If possible, we should keep it as a default slot (meaning that any HTML added inside the `LModal` component without specifying the slot name should end up in this slot.

ivanslabbert (changes requested):
Could not test this in runtime, but added a couple of comments

app/src/components/form/LModal.vue
```diff
@@ -0,0 +1,28 @@
...
+};
+defineProps<Props>();
+
+const emit = defineEmits(["close"]);
+</script>
+
+<template>
+    <Teleport to="body">
+        <div v-if="isVisible" @click.self="emit('close')">
+            <div class="fixed inset-0 z-50 bg-slate-800 bg-opacity-50 backdrop-blur-sm"></div>
```
ivanslabbert, line 14: In light mode we should perhaps choose a zinc style colour?
```suggestion
            <div class="fixed inset-0 z-50 bg-zink-800 dark:bg-slate-800 bg-opacity-50 backdrop-blur-sm"></div>
```

ivanslabbert (approved):

PR #519, opened 2024-11-13
moved syncversion from local storage to luminaryInternals index db

oslabbie:


shared/src/db/database.ts
```diff
@@ -12,6 +12,7 @@ import {
     TagDto,
     TagType,
     Uuid,
+    luminaryInternalsDto,
```
ivanslabbert, line 15: A type is normally spelled starting with a capital letter. As it is not a type used to communicate data to external systems, it is not really a DTO.
```suggestion
    LuminaryInternals,
```

shared/src/types/dto.ts
```diff
@@ -151,3 +151,8 @@ export type RedirectDto = ContentBaseDto & {
     slug: string;
     toSlug?: string;
 };
+
+export type luminaryInternalsDto = {
+    id: string;
+    value: any;
+};
```
ivanslabbert, line 158: Move to types folder, and remove `Dto` from name

ivanslabbert (approved):

PR #522, opened 2024-11-13
add bookmark on SingleContent page

belljohan3:


app/src/pages/SingleContent.vue
```diff
@@ -60,6 +62,26 @@ const is404 = computed(() => {
     return false;
 });
 
+const userPreferences = ref<{ bookmarks: { [key: string]: { ts: number } } }>({
+    bookmarks: JSON.parse(localStorage.getItem("userPreferences") || "{}").bookmarks || {},
+});
```
ivanslabbert, line 67: Can we move this to a central location that is easily accessible in other parts of the code? The place it makes most sense according to me is in globalConfig.ts

ivanslabbert (changes requested):

app/src/pages/SingleContent.vue
belljohan3: Yes we can do it.

ivanslabbert:
Created a new ticket (#524) for the bookmarks page as this ticket only includes the bookmark button

belljohan3:
> Created a new ticket (#524) for the bookmarks page as this ticket only includes the bookmark button

Can i use the same design as the actual AC website or ?

ivanslabbert (commented):
Can you add a test for globalConfig, and also a test to test the bookmark function on SingleContent?

app/src/globalConfig.ts
```diff
@@ -108,3 +108,10 @@ export const removeMediaProgress = (mediaId: string, contentId: Uuid) => {
         localStorage.setItem("mediaProgress", JSON.stringify(_mediaProgress));
     }
 };
+
+/**
+ * The user preferences.
+ */
+export const userPreferences = ref<{ bookmarks: { [key: string]: { ts: number } } }>({
```
ivanslabbert, line 115: Can we create (and export) a type for `{ bookmarks: { [key: string]: { ts: number } } }`?
We should also use naming that is use else in the project (e.g. use `id` instead of `key`)

app/src/pages/BookmarksPage.vue
```diff
@@ -0,0 +1,52 @@
...
+                undefined,
+                appLanguageIdAsRef.value,
+            );
+        } else {
+            bookmarkContent.value = [];
+        }
+    } catch (error) {
+        console.error("Error fetching bookmarked content:", error);
+    }
+};
```
ivanslabbert, line 28: This query can be re-written using `useDexieLiveQueryWithDeps`

app/src/pages/BookmarksPage.vue
```diff
@@ -0,0 +1,52 @@
...
+        <div class="flex flex-wrap gap-4">
+            <ContentTile
+                v-for="content in bookmarkContent"
+                :key="content._id"
+                :content="content"
+                class="flex w-auto justify-start"
+            />
+        </div>
+        <div v-if="bookmarkContent.length === 0" class="text-gray-500">
+            No bookmarks available for the selected language.
```
ivanslabbert, line 49:
```suggestion
            You should try this! Click [add bookmark icon here] on any post, and it will show up here.
```

app/src/pages/SingleContent.spec.ts
```diff
@@ -329,4 +332,21 @@ describe("SingleContent", () => {
             });
         });
     });
+
+    it("can bookmark a content", async () => {
```
ivanslabbert, line 336: Can you also add a test to check if it can remove a bookmark?

The test should also check if the bookmark value is added to the bookmarks list in the global config

app/src/pages/SingleContent.vue
```diff
@@ -60,6 +64,23 @@ const is404 = computed(() => {
...
+    const parentId = content.value.parentId;
+    if (!parentId) return;
+    if (userPreferences.value.bookmarks[parentId]) {
+        delete userPreferences.value.bookmarks[parentId];
+    } else {
+        userPreferences.value.bookmarks[parentId] = {
+            ts: Date.now(),
+        };
+    }
+    localStorage.setItem("userPreferences", JSON.stringify(userPreferences.value));
```
ivanslabbert, line 78: This should happen in globalConfig

app/src/globalConfig.spec.ts
```diff
@@ -65,4 +72,20 @@ describe("globalConfig.ts", () => {
...
+        userPreferencesAsRef.value = { bookmarks: { "content-post1-eng": { ts: Date.now() } } };
+        expect(userPreferencesAsRef.value).toEqual({
+            bookmarks: { "content-post1-eng": { ts: Date.now() } },
+        });
+    });
+
+    it("can remove a bookmark", async () => {
+        userPreferencesAsRef.value.bookmarks = {};
+        expect(userPreferencesAsRef.value).toEqual({ bookmarks: {} });
+    });
```
ivanslabbert, line 90: As we are not really testing if the preferences are saved to localStorage, these tests are not really adding value, so we can remove them.

app/src/pages/BookmarksPage.vue
```diff
@@ -0,0 +1,59 @@
...
+    DocType,
+    useDexieLiveQueryWithDeps,
+    type ContentDto,
+    type Uuid,
+} from "luminary-shared";
+import { computed } from "vue";
+import { BookmarkIcon } from "@heroicons/vue/24/solid";
+
+// Extract bookmark keys
+const bookmarkKeys = computed(() => Object.keys(userPreferencesAsRef.value.bookmarks));
```
ivanslabbert, line 15: you need to handle an empty userPreferences object (with no `bookmarks` key)
```suggestion
const bookmarkKeys = computed(() => Object.keys(userPreferencesAsRef.value.bookmarks || []));
```

app/src/pages/BookmarksPage.vue
```diff
@@ -0,0 +1,59 @@
...
+
+const bookmarkContent = useDexieLiveQueryWithDeps(
+    appLanguageIdAsRef,
+    (appLanguageId: Uuid) =>
+        db.docs
+            .where("parentId")
+            .anyOf(bookmarkKeys.value)
+            .filter((c) => {
+                const content = c as ContentDto;
+                if (content.type !== DocType.Content) return false;
```
ivanslabbert, line 25: Documents with a `parentId` are always of type `ContentDto`, so this line is redundant

app/src/pages/SingleContent.spec.ts
ivanslabbert: This test should check if the userPreferences ref has been updated with the bookmark.

A test where a bookmarked document is "unbookmarked" should also be added (can be done in the same test)

ivanslabbert (changes requested):
Can we also add a notification when a bookmark is added, stating that the bookmark is now added to the bookmarks page available from the user profile menu.

### Bookmark added

You can find your bookmarks from the user profile menu in the top right of the screen.

app/src/pages/BookmarksPage.vue
```diff
@@ -0,0 +1,59 @@
...
+        <div class="flex flex-wrap gap-4">
+            <ContentTile
+                v-for="content in bookmarkContent"
+                :key="content._id"
+                :content="content"
+                class="flex w-auto justify-start"
+            />
+        </div>
+        <div v-if="bookmarkContent.length === 0" class="text-gray-500">
+            You should try this ! <br />Click "<BookmarkIcon class="inline h-5 w-5" />" on any post,
```
ivanslabbert, line 55:
```suggestion
            You should try this! <br />Click "<BookmarkIcon class="inline h-5 w-5" />" on any post
```

app/src/pages/BookmarksPage.vue
```diff
@@ -0,0 +1,59 @@
...
+
+    {
+        initialValue: [],
+    },
+);
+</script>
+
+<template>
+    <div>
+        <h1 class="mb-4 text-xl font-medium">Bookmarks</h1>
```
ivanslabbert, line 45:
```suggestion
        <h1 class="mb-4 text-xl font-medium text-zinc-700 dark:text-slate-100">Bookmarks</h1>
```

Would be nice to also add the bookmarks icon to the heading? (Should then also do it for the settings page)

app/src/pages/BookmarksPage.vue
```diff
@@ -0,0 +1,59 @@
...
+        <div class="flex flex-wrap gap-4">
+            <ContentTile
+                v-for="content in bookmarkContent"
+                :key="content._id"
+                :content="content"
+                class="flex w-auto justify-start"
+            />
+        </div>
+        <div v-if="bookmarkContent.length === 0" class="text-gray-500">
```
ivanslabbert, line 54:
```suggestion
        <div v-if="!bookmarkContent.length" class="text-zinc-500 dark:text-slate-200">
```

PR #539, opened 2024-11-20
set video player according to the selected app language

belljohan3:


app/src/components/content/VideoPlayer.vue
```diff
@@ -33,6 +38,32 @@ function autoHidePlayerControls() {
...
+                            // Enable the matched audio track
+                            track.enabled = true;
+                            console.log(`Enabled audio track: ${track.label} (${track.language})`);
+                        } else {
+                            // Disable other audio tracks
+                            track.enabled = false;
+                        }
+                    }
+                }
+            }
```
ivanslabbert, line 63: I am not sure this is the best way to set the language. We have more than one audio track per language (different qualities), and in this way we are not sure which one is being selected.

I think it should be possible to change the player language (https://videojs.com/guides/options/#language)

ivanslabbert (approved):
Works as expected on my computer and iPhone simulator. Have you checked that it works correctly on an actual iPhone?

belljohan3:
Yes it works on my iPhone also.

PR #541, opened 2024-11-20
add skip button on the video player

belljohan3:
No that The following values are valid for the skip buttons options: 5 | 10 | 30. I set it to 10 sec instead of 15 as requested

### DIFF
--- a/app/src/components/content/VideoPlayer.css
+++ b/app/src/components/content/VideoPlayer.css
@@ -22,6 +22,26 @@
     @apply !text-5xl;
 }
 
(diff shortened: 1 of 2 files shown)


ivanslabbert (commented):
Will it be possible to move these buttons to the middle of the player? Could perhaps also use nicer looking icons...
telegram-cloud-photo-size-4-5909197168013591990-y

ivanslabbert (approved):
Works nice. I added a couple of options (smooth seek and play rate). Please check before merging?

PR #556, opened 2024-11-26
Display only one banner notification at a time; Fix API CORS error; Update IndexedDB default indexes;

belljohan3:


DirkDevelopmentFreelance (changes requested):
I don't see the "You are offline" banner anymore after accepting the Notification banner anymore.

ivanslabbert:
I think we need to do a little more here. When one notification is dismissed (either by the user or programmatically), the next notification in the queue should pop up after say 3 seconds.

DirkDevelopmentFreelance (commented):
Looks like it's working to me

app/src/App.vue
```diff
@@ -81,54 +81,54 @@ onBeforeMount(async () => {
...
-            if (isConnected.value) {
-                useNotificationStore().removeNotification("offlineBanner");
+            if (isConnected.value && isAuthenticated.value) {
+                useNotificationStore().removeNotification("accountBanner");
             }
         },
         { immediate: true },
     );
-}, 5000);
+}, 1000);
```
ivanslabbert, line 131: Can we make this longer?

ivanslabbert (changes requested):
We need a pause of 2 seconds before displaying the next queued banner notification.

belljohan3:
When we make the notification longer the tests fails for some of the reason.

ivanslabbert (approved):

PR #535, opened 2024-11-19
APP: Review skipped modal tests

DirkDevelopmentFreelance:

PR #557, opened 2024-11-26
add isPublished function

belljohan3:


DirkDevelopmentFreelance (commented):
Looks good to me. Putting the theme toggle in SingleContent is also nice. Good work @belljohan3!

ivanslabbert:
Found some issues with the theme switcher:

z-index problem:
image

Should we perhaps use an outline version of the dark (moon) icon? It feels a bit heavy / dark?

The selector feels a bit out of place where it is now (and a bit tight to the bookmark icon). Should we rather move it here? (It should still only show when in SingleContent.vue.)
telegram-cloud-photo-size-4-5924782659437970667-x

ivanslabbert:
According to @DirkDevelopmentFreelance, VueUse has a function to detect whether the system is in dark mode or not.

ivanslabbert:
Remove theme switcher functionality from this branch, and add to #535

DirkDevelopmentFreelance:
> According to @DirkDevelopmentFreelance, VueUse has a function to detect whether the system is in dark mode or not.

https://v4-11-2.vueuse.org/core/usedark/

ivanslabbert (approved):
Please also add a test for isPublished.ts

PR #561, opened 2024-11-29
add theme toggle icon

belljohan3:


app/src/components/navigation/ThemeSelectorModal.vue
```diff
@@ -3,13 +3,18 @@ import { ref, watch, onMounted } from "vue";
 import LButton from "../button/LButton.vue";
```
DirkDevelopmentFreelance, line 3: The theme selector just has to be updated to sync with useDark, see ``TopBar.vue``

app/src/components/navigation/ThemeSelectorModal.vue
```diff
@@ -3,13 +3,18 @@ import { ref, watch, onMounted } from "vue";
...
 import { CheckCircleIcon } from "@heroicons/vue/20/solid";
 import LModal from "../form/LModal.vue";
+import { SunIcon, MoonIcon, AdjustmentsHorizontalIcon } from "@heroicons/vue/24/outline";
```
DirkDevelopmentFreelance, line 6: Remove unused import

app/src/components/navigation/TopBar.vue
```diff
@@ -13,10 +15,14 @@ const LOGO_DARK = import.meta.env.VITE_LOGO_DARK;
 const LOGO_SMALL_DARK = import.meta.env.VITE_LOGO_SMALL_DARK;
 
 const isSmallScreen = ref(false);
+const isVisible = ref(false);
 
 const logo = computed(() => (isSmallScreen.value ? LOGO_SMALL : LOGO));
 const logoDark = computed(() => (isSmallScreen.value ? LOGO_SMALL_DARK : LOGO_DARK));
 
+// Current theme state
+const isDarkMode = useDark();
```
DirkDevelopmentFreelance, line 24: I added this, useDark works with user preference when loading up the website for the first time

app/src/components/navigation/TopBar.vue
```diff
@@ -59,8 +65,17 @@ onMounted(() => {
 
                 <DesktopMenu class="hidden lg:flex" />
                 <div class="flex-1" />
+                <div class="mx-4">
+                    <div @click="isDarkMode = !isDarkMode" class="hidden items-center lg:flex">
```
DirkDevelopmentFreelance, line 69: Replaced the component element with a normal div as a component here wasn't necessary

DirkDevelopmentFreelance (dismissed):

belljohan3:
Thanks @DirkDevelopmentFreelance . But don't you think taht it should open the modal first ? @ivanslabbert what is your take on this ? According to me, toogling directly on the icon should be fine.

DirkDevelopmentFreelance:
> Thanks @DirkDevelopmentFreelance . But don't you think taht it should open the modal first ? @ivanslabbert what is your take on this ? According to me, toogling directly on the icon should be fine. 
> 

You can do what you feel will work, I don't have a specific idea for the implementation for the toggle, just wanted to give you an example to work with😇

app/src/components/navigation/TopBar.vue
ivanslabbert, line 24: Should this not perhaps be moved to globalConfig?

ivanslabbert (changes requested):
- [ ] When changing the theme from the theme selector modal, the theme is not saved and reverts back to the old theme after reload. The modal however saves it's state, so it seems as if we are saving the state in two places now.
- [ ] Theme selector shortcut button is only showing in desktop mode, not in mobile.
- [ ] Theme selector shortcut button is too emphasized (should be less black / less white)
- [ ] Feels like the theme selector button is a bit too close to the profile menu in desktop mode
- [x] Add pointer icon (hand with pointing index finger) when hovering the theme selector button

ivanslabbert (approved):

PR #562, opened 2024-11-29
fix unavailable language to show toast

belljohan3:


DirkDevelopmentFreelance (commented):
Looks like it's working

ivanslabbert (changes requested):

PR #565, opened 2024-11-29
remove vee-validate from the cms

belljohan3:
vee-validate is only use in LInput. May be, we need to find a way to replace with something else ? @ivanslabbert

### DIFF
--- a/cms/package.json
+++ b/cms/package.json
@@ -32,7 +32,7 @@
         "pinia": "^2.1.7",
         "socket.io-client": "^4.7.4",
         "transliteration": "^2.3.5",
-        "vue": "^3.3.11",
+        "vue": "^3.5.13",
         "vue-router": "^4.2.5"
     },
     "devDependencies": {
--- a/cms/package.json
+++ b/cms/package.json
@@ -32,8 +32,7 @@
         "pinia": "^2.1.7",
         "socket.io-client": "^4.7.4",
         "transliteration": "^2.3.5",
-        "vee-validate": "^4.12.5",
-        "vue": "^3.5.13",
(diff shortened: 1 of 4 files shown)


ivanslabbert:
> vee-validate is only use in LInput. May be, we need to find a way to replace with something else ? @ivanslabbert

I think vee-validate is a bit of an overkill for inline validation of one field. Check if it can be done manually in a simple way, so that we can remove the package and reduce weight.

cms/package.json
```diff
@@ -32,8 +32,7 @@
         "pinia": "^2.1.7",
         "socket.io-client": "^4.7.4",
         "transliteration": "^2.3.5",
-        "vee-validate": "^4.12.5",
-        "vue": "^3.5.13",
+        "vue": "^3.3.11",
```
ivanslabbert, line 35: Why is the vue package being backdated from 3.5.13 to 3.3.11?

ivanslabbert (changes requested):
Looks good @belljohan3 but left one change request.

@DirkDevelopmentFreelance please do a functional check?

cms/package.json
belljohan3: It is probably a mistake from the rebase process

DirkDevelopmentFreelance (approved):
Seems good

cms/package.json
DirkDevelopmentFreelance: This still needs to be changed I think

DirkDevelopmentFreelance (changes requested):

ivanslabbert (changes requested):

PR #574, opened 2024-12-06
568-fix-scrolling-for-older-browsers

ivanslabbert:


ivanslabbert:
@oslabbie could you review this bug fix? It might affect your Capacitor app (even though I think it should not...)

To support older browsers I had to set the height of App.vue using absolute positioning instead of using the newer dynamic viewport height units (not supported on Chromium < 108)

Is there a way of patching the production app with the same (without pushing Staging to production)?

belljohan3 (approved):

PR #576, opened 2024-12-10
Fix bug in db.service where fields with an undefined value causes the update comparison to believe the updated doc has changed.

ivanslabbert:


belljohan3 (approved):
NIce fix

PR #585, opened 2024-12-13
hide playback rate button for live stream

belljohan3:


ivanslabbert (approved):
Looks good. I think we should also hide the 10s forward and backward buttons when live.

PR #590, opened 2024-12-18
517 api shared live api

oslabbie:


shared/src/socket/socketio.ts
```diff
@@ -38,12 +43,13 @@ class Socketio {
...
     constructor(apiUrl: string, cms: boolean = false, token?: string) {
         this.isCms = cms;
+        console.log(this.isCms);
 
         this.socket = io(apiUrl, token ? { auth: { token } } : undefined);
 
         this.socket.on("connect", () => {
             isConnected.value = true;
-            this.requestData();
+            this.socket.emit("joinSocketGroups", { cms: this.isCms });
```
ivanslabbert, line 52: Change this to send the list of needed docs to the API to only send the needed documents to the client.

shared/src/socket/socketio.ts
```diff
@@ -118,11 +124,23 @@ class Socketio {
...
+        //     userId: "user-super-admin",
+        //     reqData: {
+        //         version: db.syncVersion,
+        //         cms: this.isCms,
+        //         accessMap: accessMap.value,
+        //     },
+        // });
+        // await db.bulkPut(res.docs);
+        // if (res.version != undefined) db.syncVersion = res.version;
     }
```
ivanslabbert, line 144: Remove function

shared/src/types/dto.ts
```diff
@@ -154,3 +154,31 @@ export type RedirectDto = ContentBaseDto & {
     slug: string;
     toSlug?: string;
 };
+
+export type apiConnectionOptions = {
```
ivanslabbert, line 158:
```suggestion
export type ApiConnectionOptions = {
```

ivanslabbert (changes requested):

shared/src/socket/socketio.ts
oslabbie: This is already implemented with the isCms option 
We will have to re-impliment if we remove the isCms option

api/src/docs/docs.service.ts
```diff
@@ -0,0 +1,117 @@
...
+
+        // Determine which doc types to get
+        const docTypes = req.cms ? [...this.cmsDocTypes, ...this.appDocTypes] : this.appDocTypes;
+
+        // Get user accessible groups
+        const userViewGroups = PermissionSystem.accessMapToGroups(
+            accessMap,
+            AclPermission.View,
+            docTypes,
+        );
```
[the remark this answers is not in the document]
oslabbie: This does not make sense to move into a separate class, since the implementation on the nest and socket side, is a specific implimentation, and not a universal implementation

ivanslabbert (approved):
Approved with comments

PR #591, opened 2025-01-02
add autofullscreen and autoplay query string parameters

ivanslabbert:


ivanslabbert:
Need to add tests for the two query string parameters.

belljohan3:
I didn't manage to get it work. @DirkDevelopmentFreelance can you also look on your side ?

DirkDevelopmentFreelance:
> I didn't manage to get it work. @DirkDevelopmentFreelance can you also look on your side ? 

Sure! Will try to get to it ASAP before 20:00 SAST

DirkDevelopmentFreelance (commented):
The auto fullscreen works, but I can't seem to get the autoplaying to work? Not sure if it just doesn't work, or if I should write a plugin manually for it?

app/src/components/content/VideoPlayer.vue
```diff
@@ -162,10 +166,22 @@ onMounted(() => {
         if (progress > 60) player.currentTime(progress - 30);
     });
 
-    // Remove player progress on ended
```
DirkDevelopmentFreelance, line 165: Just a general question, why do we remove the progress of the video the user is currently watching?  Don't we want to save their progress, because what if they accidentally exit out of the page or video? Or want to come back to it later after they have bookmarked it?

ivanslabbert: Yes, we are saving the player progress. The removing logic will only remove the state if the player reaches the end of the video (i.e. that the user is done watching the video)

DirkDevelopmentFreelance: I see, yes that makes sense
I will then perhaps suggest changing it to something like "on video ended" or something to just make it more clear.

ivanslabbert:
> The auto fullscreen works, but I can't seem to get the autoplaying to work? Not sure if it just doesn't work, or if I should write a plugin manually for it?

When opening a URL which includes a slug (with `?autoplay=true&autofullscreen=true`), auto-play and auto-full-screen will only work if playing video / audio without user interaction is enabled in the browser (e.g. in chrome://flags)

However, when enabling on page load (e.g. when opening the home page http://localhost:4174?autoplay=true&autofullscreen=true) the video will play automatically and go to full screen when clicking on a post, as the user interaction was triggered just before the video is loaded.

DirkDevelopmentFreelance:
> > The auto fullscreen works, but I can't seem to get the autoplaying to work? Not sure if it just doesn't work, or if I should write a plugin manually for it?
> 
> When opening a URL which includes a slug (with `?autoplay=true&autofullscreen=true`), auto-play and auto-full-screen will only work if playing video / audio without user interaction is enabled in the browser (e.g. in chrome://flags)
> 
> However, when enabling on page load (e.g. when opening the home page http://localhost:4174?autoplay=true&autofullscreen=true) the video will play automatically and go to full screen when clicking on a post, as the user interaction was triggered just before the video is loaded.

Autofullscreen only works when I click on the video's pause/play buttons, I would expect it to also go full screen when it is auto playing but it's only auto playing when I load in and then when I click on the pause button, it pauses, then i have to click play for it to go full screen

ivanslabbert:
Need to add error handling to `exitFullscreen`
image

PR #595, opened 2025-01-06
Add toggle button for grid/list view in ExplorePage

johan-bell:


DirkDevelopmentFreelance (changes requested):
Some comments
Screenshot From 2025-01-07 11-29-01
In mobile view the toggle is inside the search input, which makes it inaccessable, when you click on it  it just focuses the search input.

Screenshot From 2025-01-07 11-29-17
Screenshot From 2025-01-07 11-30-59
In both mobile and desktop view, the grid is not displaying correctly. The button on top to toggle grid and list view causes the image under it to also display lower than the rest of the grid on the left. 
And there is a tremendous gap between items.

DirkDevelopmentFreelance (changes requested):
It looks good, but the space between the search bar and the toggle icon is too large still:
image
The search input also has a massive gap in between the icon and the element in desktop mode, and when tabbing through the page, the focus outline is much bigger than the icon:
image

DirkDevelopmentFreelance (changes requested):
image
I suggest maybe putting the button toggle on left side and extending the search bar the full width of its container, or extending the search bar up until the button toggle
Also maybe make the default layout on desktop be the grid view already and on mobile the list view. 
image
I believe making the toggle icon's width be the same as the icon will fix the issue with the search bar
I believe I made the same comments in my previous review so maybe also look at them?

PR #598, opened 2025-01-07
597 app video home page

belljohan3:


ivanslabbert (commented):
- [ ] Need test for VideoPage.vue (including the test(s) removed from HomePage.spec.ts)

image
- [ ] remove white padding above pinned categories in the video page. (Check that distance of text to top bar matches that of the home page - might need to find a way to bleed the yellow background to the top)

image
- [ ] Video button missing in the menu bar

ivanslabbert (approved):

PR #619, opened 2025-01-17
Make image non mandatory

DirkDevelopmentFreelance:
The app will display a fallback image, resulting in the mandatory image in ``EditContent`` being removed by simply removing the code that enforces the rule.

### DIFF
--- a/app/src/components/IgnorePagePadding.vue
+++ b/app/src/components/IgnorePagePadding.vue
@@ -3,10 +3,12 @@ withDefaults(
     defineProps<{
         mobileOnly?: boolean;
         ignoreTop?: boolean;
+        ignoreBottom?: boolean;
     }>(),
     {
         mobileOnly: false,
         ignoreTop: false,
+        ignoreBottom: false,
     },
 );
 </script>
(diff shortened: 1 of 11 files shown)


ivanslabbert (commented):
image

- [x] If images are not available, it is not showing the default image unless when hovered over
- [x] Image height dynamic instead of fixed. Content with longer titles expands image height.

image

- [ ] Fallback images not shown on Explore page if not available from S3 storage / offline

image

- [ ] Fallback images not shown on SingleContent page if not available from S3 storage / offline, but still keeping image area open. Should perhaps collapse the area? If the video field is in use, we could use the same look and feel as used in the content tiles?

DirkDevelopmentFreelance:
On My side it works fine 
image
image

But realised text isn't on other parts of the app and then it's only a blur which looks weird. So I am going to fix that

belljohan3 (commented):
LGTM @DirkDevelopmentFreelance 
You did a good job. 
Don't you think that the app should show the default image when i the image is not set in the cms ? Unless @ivanslabbert wanted it like this.
I pushed some changes to fix a layout issue with CopyrihtBanner in the app. But the rest is ok

ivanslabbert (approved):

PR #624, opened 2025-01-21
hotfix, calculate syncMap when client data is called, to avoid functi…

oslabbie:
…on running without a syncMap

### DIFF
--- a/shared/src/docs/sync.spec.ts
+++ b/shared/src/docs/sync.spec.ts
@@ -413,4 +413,22 @@ describe("rest", () => {
         },
         { timeout: 10000 },
     );
+
+    // Test not working, syncMap does not update on test side, only in the syncFile
+    it.skip(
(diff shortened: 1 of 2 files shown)


shared/src/docs/sync.spec.ts
```diff
@@ -396,16 +396,39 @@ describe("rest", () => {
...
+                const post = syncMap.value.get("post_group-recursive-test");
                 expect(post?.blocks[0].blockStart).toBe(20000);
                 expect(post?.blocks[0].blockEnd).toBe(1000);
             }, 9000);
         },
         { timeout: 10000 },
     );
+
+    // Test not working, syncMap does not update on test side, only in the syncFile
+    it.skip(
```
ivanslabbert, line 418: Create task to come back to this test

ivanslabbert (approved):

Issue #575, opened 2024-12-09
APP: Add categories to explore (topics) page


ivanslabbert:

ivanslabbert:
Suggestions for layout including topics organised into categories:
- Ivan: use same layout as current home page
- Dirk: will come with suggestion
- Johan: Text only - will come with suggestion

PR #633, opened 2025-01-23
Implement infinite scroll on allTopics

belljohan3:


app/src/components/ExplorePage/AllTopics.vue
```diff
@@ -42,22 +36,35 @@ watch(allTopics, async (value) => {
...
 
+// Infinite scroll setup
+const scrollElement = ref<HTMLElement | null>(null);
+const scrollPosition = ref(10);
+const infiniteScrollData = computed(() => filteredTopics.value.slice(0, scrollPosition.value));
+
+useInfiniteScroll(
+    scrollElement,
+    () => {
+        if (scrollPosition.value < filteredTopics.value.length) {
```
DirkDevelopmentFreelance, line 59: Maybe this should check the length of the infiniteScrollData instead of filteredTopics? Maybe that's why it's not working

DirkDevelopmentFreelance (changes requested):
I am not experiencing the intended functionality

ivanslabbert (commented):
If feels like the infinite scroll is not doing its job correctly. I think it has to be applied to the scrolling parent element, but in this case it is applied to an element that is overflowing the parent, so it might not be able to detect correctly where it is at.

We could wait with this until #575 is done, or just include the infinite scroll in the same PR and close this one?

PR #634, opened 2025-01-23
577 app category content popup on singlecontent page max height not set correctly

belljohan3:


DirkDevelopmentFreelance (commented):
Nice implementation.
image
Maybe we should also customise the scrollbar here and make it thinner? To me it feels out of place

app/src/pages/SingleContent.vue
```diff
@@ -352,7 +352,7 @@ const selectedCategory = computed(() => {
         @close="showCategoryModal = false"
         :heading="selectedCategory?.title || ''"
     >
-        <div class=" ">
+        <div class="max-h-[80vh] overflow-y-auto">
```
ivanslabbert, line 355: using `vh` does not work on older browsers (I had issues with the home page not being able to scroll on older browsers). Please make use of more compatible CSS?

ivanslabbert (changes requested):

PR #635, opened 2025-01-23
589 app rethink 404 when not logged in

belljohan3:


DirkDevelopmentFreelance (commented):
Nice!
But I am experiencing that the SingleContent "loading" appears right before the 404 message for a about 2 seconds, I think this shouldn't happen maybe?

belljohan3:
@DirkDevelopmentFreelance 

The `404` code has been removed as requested. 

If you are logged in and you are navigating to a route with an invalid slug, it shows this:

image

If you are not logged in and you are navigating to a route with an invalid slug, it shows this:

image

belljohan3:
> Nice! But I am experiencing that the SingleContent "loading" appears right before the 404 message for a about 2 seconds, I think this shouldn't happen maybe?

I think it should happens. We set it in the `SingleContent.vue` component.

DirkDevelopmentFreelance:
> @DirkDevelopmentFreelance 
> 
> The `404` code has been removed as requested. 
> 
> If you are logged in and you are navigating to a route with an invalid slug, it shows this:
> 
> image
> 
> If you are not logged in and you are navigating to a route with an invalid slug, it shows this:
> 
> image
> 

Yes that is working. I am talking about the loading state.

belljohan3:
I think that it is normal to have to the loading state. @ivanslabbert !?

DirkDevelopmentFreelance:
> I think that it is normal to have to the loading state. @ivanslabbert !?

Before the "you are not logged in" or "404". The loading state quickly flashes. Yes the loading should be there. But if there is no data to load I personally don't think it makes sense to have a loading state flash to the user because it never loaded anything, it seems like a visual artifact. Does this make more sense to you?😇

oslabbie:
i Agree with @DirkDevelopmentFreelance in the visual sense, but I am not sure how possible it is hiding the loading, since the app need to query the db to check if the document exists, before it can know that it does not exists

PR #638, opened 2025-01-23
Add tests for pr 465

### LINKED ISSUE #571: Add tests for PR-465
https://github.com/bccsa/luminary/pull/465

MrDirkelz:


api/src/changeRequests/processChangeRequest.spec.ts
```diff
@@ -266,27 +267,42 @@ describe("processChangeRequest", () => {
...
+
+        const languages = await db.getDocsByType(DocType.Language);
+
+        languages.docs.forEach((language) => {
+            if (language._id === "lang-fra") {
+                expect(language.default).toBe(1);
+            } else {
+                expect(language.default).toBe(0);
+            }
+        });
```
belljohan3, line 306: I think the way we are expecting here is not really. As `English` is the default language, you should check if the value of the french is set to `0` as the logic says in the TS file.

DirkDevelopmentFreelance: 1 = True, 0 = False. 
``ts
languages.docs.forEach((language) => {
            if (language._id === "lang-fra") {
                expect(language.default).toBe(1); // This means french is the new default language
            } else {
                expect(language.default).toBe(0); // All other languages now has to be false
            }
        });
``

I added comments here to explain the code.

PR #652, opened 2025-01-24
Cms language selection

DirkDevelopmentFreelance:


belljohan3:
When i am selecting French for e.g., i switch to swahili, it shows french. shouldn't it fallback on English or to the default language if it is set ?

DirkDevelopmentFreelance:
> I suggest having this oder: Settings > Language > Sign out > .... Sanbox will probably be removed
> 
> image

I agree with you. Thanks for reviewing!

belljohan3:
That means will also need the strings translations here @ivanslabbert ?

belljohan3:
image

What do you think if we set the color of the black like the close button ? (Check the image below)

image

ivanslabbert:
> That means will also need the strings translations here @ivanslabbert ?

We will not require strings translation here for the time being

ivanslabbert:
> I suggest having this oder: Settings > Language > Sign out > .... Sanbox will probably be removed
> 
The sandbox only shows when in local dev mode, and should be kept as a testing ground for components (for now at least)

cms/src/components/navigation/ProfileMenu.spec.ts
```diff
@@ -5,11 +5,21 @@ import * as auth0 from "@auth0/auth0-vue";
...
-vi.mock("vue-router", () => ({
-    useRouter: vi.fn().mockImplementation(() => ({
-        push: routePushMock,
-    })),
-}));
+// vi.mock("vue-router", () => ({
+//     useRouter: vi.fn().mockImplementation(() => ({
+//         push: routePushMock,
+//     })),
+// }));
```
ivanslabbert, line 12: Remove commented code

ivanslabbert (changes requested):
The functionality works, but I want to add some requests

- [x] Load set language to browser language (if available) - the same way as done in the app. If nothing is set (first load) it loads Chichewa in my dev environment.
- [x] Language selection is currently only done when navigating to a content overview. I think it should be set on a higher level. (In the app it is done in main.ts)

cms/src/components/navigation/ProfileMenu.vue
```diff
@@ -1,26 +1,57 @@
...
 
 const { user, logout } = useAuth0();
 const router = useRouter();
 
-const userNavigation: {
-    name: string;
-    action: Function;
-}[] = [
-    { name: "Settings", action: () => router.push({ name: "settings" }) },
+const shouldDisplayLanguageModal = ref(false);
```
ivanslabbert, line 22:
```suggestion
const showLanguageModal = ref(false);
```

cms/src/components/modals/LanguageModal.vue
```diff
@@ -0,0 +1,58 @@
+<script setup lang="ts">
+import { db, DocType, type LanguageDto } from "luminary-shared";
+import LButton from "../button/LButton.vue";
+import { CheckCircleIcon } from "@heroicons/vue/20/solid";
+import { cmsLanguageIdAsRef } from "@/globalConfig";
+import LModal from "./LModal.vue";
+
+type Props = {
+    isVisible: boolean;
```
ivanslabbert, line 9: Using a v-model for isVisible could be a bit nicer (then you will not need to use events for resetting it from the parent control)

cms/src/components/modals/LModal.vue
```diff
@@ -0,0 +1,37 @@
...
+import App from "@/App.vue";
+
+type Props = {
+    heading: string;
+};
+defineProps<Props>();
+
+const isVisible = defineModel<boolean>("isVisible");
+const isTestEnviroment = import.meta.env.MODE === "test";
+const emit = defineEmits(["close"]);
```
ivanslabbert, line 11: Could the modal not jus set `isVisible` to false to close itself (e.g. when clicking close / clicking outside of the window)? Then the parent control does not need to handle the event etc.

ivanslabbert (changes requested):
- [ ] Bug: selected language reverts back to English on page reload

image
- [x] Language dropdown can be removed from content overview page
- [ ] When changing the language, the content overview should reload with the correct language selection (think you will need to implement the queries with useDexieLiveQuery)

cms/src/components/navigation/ProfileMenu.vue
DirkDevelopmentFreelance: Done

cms/src/components/modals/LModal.vue
DirkDevelopmentFreelance: The Lmodal used here is the same as the app's, the Close button is made outside this component, inside the LanguageModal itself

cms/src/components/modals/LanguageModal.vue
```diff
@@ -0,0 +1,58 @@
...
+const setLanguage = (id: string) => {
+    cmsLanguageIdAsRef.value = id;
+    emit("close");
+};
+</script>
+
+<template>
+    <LModal
+        name="lModal-languages"
+        heading="Select Language"
```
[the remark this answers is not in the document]
DirkDevelopmentFreelance: done

ivanslabbert:
Found a bug: When deleting the selected language from localstorage and refreshing the page, no language is selected. In such a case it should automatically select the browser language (if existing) or revert to select the default language (as set in the CMS languages)

DirkDevelopmentFreelance:
> Found a few bugs:
> 
> * [ ]  When deleting the selected language from localstorage and refreshing the page, no language is selected. In such a case it should automatically select the browser language (if existing) or revert to select the default language (as set in the CMS languages)
It does this on my side
> * [ ]  After selecting a for the first time (e.g. after clearing language from local storage and refreshing the page), the content overview does not automatically refresh with the selected language content.
It also does this on my side

cms/src/globalConfig.ts
```diff
@@ -13,3 +14,25 @@ export const cmsLanguageIdAsRef = ref(localStorage.getItem("cms_selectedLanguage
 watch(cmsLanguageIdAsRef, (newVal) => {
     localStorage.setItem("cms_selectedLanguage", newVal);
 });
+
+const languageInitialised = JSON.parse(localStorage.getItem("language_initialised") || "false");
+
```
ivanslabbert, line 19: I do not think it is needed to set an extra flag. We can probably just check if cmsLanguageIdAsRef is set or not?

cms/src/components/content/ContentOverview.spec.ts
```diff
@@ -118,90 +127,6 @@ describe("ContentOverview.vue", () => {
         });
     });
 
-    it("should switch languages correctly", async () => {
```
ivanslabbert, line 121: We need to add a test (in LanguageModal perhaps) to ensure language switching is working as expected

cms/src/globalConfig.ts
```diff
@@ -13,3 +14,25 @@ export const cmsLanguageIdAsRef = ref(localStorage.getItem("cms_selectedLanguage
...
+        const preferredLang = languages.find(
+            (lang) => lang.languageCode === browserPreferredLanguage,
+        );
+        const defaultLang = languages.find((lang) => lang.default === 1);
+
+        cmsLanguageIdAsRef.value = preferredLang?._id || defaultLang?._id || "";
+    } else {
+        cmsLanguageIdAsRef.value = languages.filter((lang) => lang.default === 1)[0]._id;
+    }
+    console.log(cmsLanguageIdAsRef.value);
```
ivanslabbert, line 36: Remove console log

cms/src/components/modals/LanguageModal.spec.ts
```diff
@@ -0,0 +1,98 @@
+import "fake-indexeddb/auto";
+import { describe, it, expect, beforeEach, afterEach } from "vitest";
+import { mount } from "@vue/test-utils";
+import LanguageModal from "./LanguageModal.vue";
+import { db } from "luminary-shared";
+import { mockLanguageDtoEng, mockLanguageDtoFra, mockLanguageDtoSwa } from "@/tests/mockdata";
+import { cmsLanguageIdAsRef } from "@/globalConfig";
+import waitForExpect from "wait-for-expect";
```
ivanslabbert, line 8: Unused import

cms/src/globalConfig.ts
ivanslabbert: Still using the languageInitialized flag? I think it is a bit risky to hard-store flags like this, where we should be able to get it from checking if cmsLanguageIdAsRef is set or not.

DirkDevelopmentFreelance: I thought I removed that flag. Maybe didn't commit it and forgot about it.

DirkDevelopmentFreelance (commented):
Fix conflicts

ivanslabbert (approved):
Works as expected. Please check unused imports, and resolve conflicts before merging?

PR #653, opened 2025-01-24
Fix EditContentText not being disabeld

DirkDevelopmentFreelance:


oslabbie (approved):
this looks good, perhaps just update the branch before you commit?

PR #655, opened 2025-01-27
Make layout changes to EditContent

DirkDevelopmentFreelance:
Moved the sidebar from the right to the left, following the principle that users should see important information first
Moved groupSelector on top of the LImage component to make it clearer to the user, maybe add some bottom margin to seperate them more?
Moved "view live version" inside of each translation in the validation container, but only if the post is publishable, and contains a title. 
When I first started using the CMS, I remember that I skimmed over it and didn't ever really used and actually forgot about that feature, so now users will be able to find it more easily and it gives more space for the EditParent component

### DIFF
--- a/cms/src/components/content/EditContent.vue
+++ b/cms/src/components/content/EditContent.vue
@@ -319,6 +319,32 @@ watch(selectedLanguage, () => {
             </div>
         </template>
         <div class="relative grid grid-cols-3 gap-8">
+            <!-- Sidebar -->
+            <div class="col-span-3 md:col-span-1" v-if="parent">
+                <div class="sticky top-20 space-y-6">
+                    <EditContentParentValidation
+                        :can-translate="canTranslate"
+                        :can-publish="canPublish"
+                        :can-edit="canEditParent"
+                        v-if="contentDocs"
+                        v-model:parent="parent"
+                        v-model:contentDocs="contentDocs"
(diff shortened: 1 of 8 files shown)


DirkDevelopmentFreelance:
[] Enable editing without group membership

cms/src/components/content/EditContent.vue
ivanslabbert, line 1: Remove unused import (EditContentPreview), or add back in project

ivanslabbert (commented):
image

- [x] The "view live version" button is a bit too prominent / big. I clicked it by mistake when wanting to switch languages. Perhaps the button could be moved to next to the title (with "view live version" text in tooltip, and button only enabled when online and state is "published")
- [x] Unselected languages' buttons are too light - it is hard to see which language is selected
- [x] When adding a language, the new language is not automatically selected

image

- [ ] The list of languages in the placeholder is not filtering out languages already added. This list should only show languages to which the user has translate access.

image

- [x] Spacing between the group selector and image editor is a bit too tight

- [x] "View live version" button should open the link in a new tab / window, and not navigate away from the CMS.
- [ ] The left (settings) and right (editing) columns should scroll individually. You currently have to scroll to the bottom of the editing column (when there is a long text) to get to the bottom of the settings column.

ivanslabbert (approved):

PR #661, opened 2025-01-29
Convert login page into modal & updated tests

belljohan3:


ivanslabbert (changes requested):
Few change requests:

- Rename LoginPage.vue to LoginModal.vue, and move to a more appropriate folder
- The modal should be added in the profile menu (it's primary location) and not in App.vue.
- Add a ref to globalConfig to control the modal's visibility, so that we can call the modal from other locations (e.g. the "you are missing out" notification banner, the home page if there is no content, etc.)

app/src/pages/LoginPage.vue
```diff
@@ -27,21 +31,37 @@ const loginAsGuest = async () => {
...
+        :isVisible="showLoginModal || false"
+        heading="Log in"
+        class="mx-auto max-w-lg"
+        @close="showLoginModal = false"
+    >
+        <p class="my-4" v-if="!isConnected">
+            You are offline. Please connect to the internet to log in.
+        </p>
+
+        <LCard>
```
ivanslabbert, line 44: Add a `v-else` here to hide the login section when offline

ivanslabbert (approved):

PR #664, opened 2025-01-30
Check for browser compatibility

MrDirkelz:


app/postcss.config.js
```diff
@@ -1,6 +1,23 @@
+import tailwindcss from "tailwindcss";
+import autoprefixer from "autoprefixer";
+import doiuse from "doiuse";
+
 export default {
```
belljohan3, line 5: Can you explain how `doiuse` works here ?

johan-bell (commented):
Hi @DirkDevelopmentFreelance left a review.

app/postcss.config.js
DirkDevelopmentFreelance: See original ticket on the board. Left a link for the documentation of the package.

app/postcss.config.js
```diff
@@ -1,6 +1,23 @@
...
-        autoprefixer: {},
-    },
+    plugins: [
+        tailwindcss(),
+        autoprefixer(),
+        doiuse({
+            browsers: [
+                "chrome >= 50",
+                "firefox >= 45",
+                "edge >= 12",
```
ivanslabbert, line 13: Not sure if edge refers to the old MS edge or the new Chromium based Edge? version numbers on Edge are now similar to Chrome's version numbers, so supporting Edge Chromium V12 would be going back very far...

ivanslabbert (changes requested):
Hi @DirkDevelopmentFreelance 
Is the check failing the Lint testing, or just giving a warning? I think it should be failing tests if features are not supported by the target browser versions.

PR #669, opened 2025-01-30
Fix EditContent Bugs

DirkDevelopmentFreelance:


ivanslabbert (approved):
Found another issue when a new blog post is created:

- [ ] When adding the first content doc (e.g. for English), the slug is empty:
image

The rest of the issues seems to be fixed

PR #674, opened 2025-01-31
redesign of explore page

belljohan3:


DirkDevelopmentFreelance (commented):
It looks good and feels smooth. I just think we need to look at the visuals a bit more in depth maybe. But it looks good. Just one lint warning that was probably removed it another PR.

DirkDevelopmentFreelance (changes requested):
This PR is not up to date with the latest branch commits I think making it not work. It is not showing my topics on the explore page, and the package.json has conflicts it seems...

PR #679, opened 2025-02-03
Redo startup sequencing

ivanslabbert:


ivanslabbert (commented):
* Simplify App.vue by moving startup logic to main.ts
* Simplify main.ts by moving functions to separate file (i18n) / globalConfig (default language computed function)
* Change functions to async / promised based functions to give more control on startup sequencing (initLanguage)

oslabbie (approved):

PR #687, opened 2025-02-04
App: fix buggy language selector

DirkDevelopmentFreelance:
Changes have been commented.

### DIFF
--- a/app/src/components/navigation/LanguageModal.vue
+++ b/app/src/components/navigation/LanguageModal.vue
@@ -21,7 +21,7 @@ const emit = defineEmits(["close"]);
 const defaultLanguage = computed(() => languages.value.find((lang) => lang.default == 1));
 
 const setLanguage = (id: string) => {
-    if (!appLanguageIdsAsRef.value.includes(id)) {
+    if (!new Set(appLanguageIdsAsRef.value).has(id)) {
         appLanguageIdsAsRef.value.push(id);
     }
 };
--- a/app/src/components/navigation/LanguageModal.vue
+++ b/app/src/components/navigation/LanguageModal.vue
@@ -1,7 +1,7 @@
 <script setup lang="ts">
 import { db, DocType, type LanguageDto } from "luminary-shared";
 import LButton from "../button/LButton.vue";
-import { appLanguageIdsAsRef } from "@/globalConfig";
+import { appLanguageIdsAsRef, appLanguagesPreferredAsRef } from "@/globalConfig";
 import LModal from "../form/LModal.vue";
 import { ArrowDownIcon, ArrowUpIcon } from "@heroicons/vue/24/solid";
 import { CheckCircleIcon } from "@heroicons/vue/20/solid";
@@ -82,6 +82,8 @@ const removeFromSelected = (id: string) => {
         :is-visible="isVisible"
         @close="emit('close')"
     >
+        {{ appLanguageIdsAsRef }}
+
         <transition-group
             name="language"
             tag="div"
--- a/app/src/components/navigation/LanguageModal.vue
+++ b/app/src/components/navigation/LanguageModal.vue
@@ -82,8 +82,6 @@ const removeFromSelected = (id: string) => {
         :is-visible="isVisible"
         @close="emit('close')"
     >
-        {{ appLanguageIdsAsRef }}
-
         <transition-group
             name="language"
             tag="div"
--- a/app/src/components/navigation/LanguageModal.vue
+++ b/app/src/components/navigation/LanguageModal.vue
@@ -1,7 +1,7 @@
 <script setup lang="ts">
 import { db, DocType, type LanguageDto } from "luminary-shared";
 import LButton from "../button/LButton.vue";
-import { appLanguageIdsAsRef, appLanguagesPreferredAsRef } from "@/globalConfig";
+import { appLanguageIdsAsRef } from "@/globalConfig";
 import LModal from "../form/LModal.vue";
 import { ArrowDownIcon, ArrowUpIcon } from "@heroicons/vue/24/solid";
 import { CheckCircleIcon } from "@heroicons/vue/20/solid";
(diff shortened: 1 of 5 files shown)


app/src/components/navigation/LanguageModal.vue
```diff
@@ -1,7 +1,7 @@
 <script setup lang="ts">
 import { db, DocType, type LanguageDto } from "luminary-shared";
 import LButton from "../button/LButton.vue";
-import { appLanguageIdsAsRef } from "@/globalConfig";
+import { appLanguageIdsAsRef, appLanguagesPreferredAsRef } from "@/globalConfig";
```
belljohan3, line 4: remove unused import ?

app/src/components/navigation/LanguageModal.vue
```diff
@@ -82,6 +82,8 @@ const removeFromSelected = (id: string) => {
         :is-visible="isVisible"
         @close="emit('close')"
     >
+        {{ appLanguageIdsAsRef }}
```
belljohan3, line 85: probably forgot to remove this log ?

belljohan3 (changes requested):

app/src/components/navigation/LanguageModal.vue
```diff
@@ -21,7 +21,7 @@ const emit = defineEmits(["close"]);
 const defaultLanguage = computed(() => languages.value.find((lang) => lang.default == 1));
 
 const setLanguage = (id: string) => {
-    if (!appLanguageIdsAsRef.value.includes(id)) {
+    if (!new Set(appLanguageIdsAsRef.value).has(id)) {
```
DirkDevelopmentFreelance, line 24: Create a set quickly in the if statement to ensure that the language can't be duplicated

app/src/globalConfig.ts
```diff
@@ -22,6 +22,7 @@ export const appLanguageIdsAsRef = ref<string[]>(
  * Set the default language of the app.
  */
 function setAppDefaultLanguage(languageId: Uuid) {
+    if (appLanguageIdsAsRef.value.length > 1) return;
```
DirkDevelopmentFreelance, line 25: We can assume that the array has changed and the default language has been set if the length is no longer less than 1

app/src/globalConfig.ts
```diff
@@ -36,7 +37,9 @@ function setAppDefaultLanguage(languageId: Uuid) {
 watch(
     appLanguageIdsAsRef,
     (newVal) => {
-        localStorage.setItem("languages", JSON.stringify(newVal));
+        if (appLanguageIdsAsRef.value.includes(typeof null))
```
DirkDevelopmentFreelance, line 40: Clean up the array to ensure null values are not in the array as a safety precaution

app/src/globalConfig.ts
```diff
@@ -36,7 +37,9 @@ function setAppDefaultLanguage(languageId: Uuid) {
...
+            appLanguageIdsAsRef.value = appLanguageIdsAsRef.value.filter((id) => id !== null);
+        localStorage.setItem("languages", JSON.stringify(newVal.filter((id) => id != null)));
```
DirkDevelopmentFreelance, line 42: This ensures that there are no null values in local storage if all else fails

app/src/globalConfig.ts
```diff
@@ -130,15 +133,10 @@ export const initLanguage = () => {
...
-                // If the browser preferred language does not match any of the available content languages,
-                // set the CMS defined default language as the preferred language. If no default language is defined
-                // in the CMS, set the first available language as the preferred language.
-                appLanguageIdsAsRef.value[0] = cmsDefaultLanguage?._id || _languages[0]._id; // ??
-
                 // Add the CMS defined default language to the list of preferred languages if it is not already there
                 if (
                     cmsDefaultLanguage &&
-                    !appLanguageIdsAsRef.value.some((l) => l === cmsDefaultLanguage._id)
+                    !appLanguageIdsAsRef.value.includes(cmsDefaultLanguage._id)
```
DirkDevelopmentFreelance, line 139: Includes makes more sense here in my opinion as I think we want to check all elements in the array to enforce uniqueness

app/src/components/navigation/LanguageModal.vue
ivanslabbert: Using a set will not be improving anything (where we are just checking for the existance of an id). Sets are useful to prevent duplicates, but it would require creating and converting it on every cycle.

app/src/globalConfig.ts
```diff
@@ -36,7 +37,9 @@ function setAppDefaultLanguage(languageId: Uuid) {
 watch(
     appLanguageIdsAsRef,
     (newVal) => {
-        localStorage.setItem("languages", JSON.stringify(newVal));
+        if (appLanguageIdsAsRef.value.includes(typeof null))
+            appLanguageIdsAsRef.value = appLanguageIdsAsRef.value.filter((id) => id !== null);
```
ivanslabbert, line 41: Do not change a ref inside a watcher that is watching it...
You can rather filter for nulls where you are creating / modifying the array

app/src/globalConfig.ts
ivanslabbert:
```suggestion
        localStorage.setItem("languages", JSON.stringify(newVal.filter((id) => id)));
```
should also work (and will work for undefined values as well)

ivanslabbert (commented):
Left some comments

app/src/components/navigation/LanguageModal.vue
DirkDevelopmentFreelance: This set just quickly creates a new unique array that doesn't have duplicated values to check for the id, this ensures that only one id of each is in there.

app/src/globalConfig.ts
DirkDevelopmentFreelance: This code only has to run 1 time I believe, the only time should be when the user runs the website for the first time. If the array is larger than 1, it means the default language was added, and the browser preferred languages was added or the user made changes to their preference. I don't think it is neccesarry to set a default language everytime. But only if you think the admins will change the default language a lot of times?

ivanslabbert: Think there is some confusion in naming here. The app default language would be the language in the top of the array, so perhaps this function should be removed altogether, and the one place where the logic is used it can be implemented directly.

ivanslabbert (approved):

PR #696, opened 2025-02-10
Update Vue and Dexie

ivanslabbert:
Update Vue to 3.5.13
Update Dexie to V4.0.11

Note that minification of the share library is disabled in this PR to simplify troubleshooting. When bundling with the main project (app / cms) the bundler should automatically minify it together with other modules.


DirkDevelopmentFreelance (approved):
This looks good to me, @belljohan3, I think you can give it a look as well and then just approve it too

belljohan3:
> This looks good to me, @belljohan3, I think you can give it a look as well and then just approve it too

Yes i agree

PR #705, opened 2025-02-12
Add document delete function

ivanslabbert:
- [x] Handle rejection of Post/Tag deletes where the associated content documents are already removed from IndexedDb. As these documents were manually removed and no updatedUtcTime properties have been updated, the sync protocol will not know to sync back the docs to the client.
- [x] Delete associated content documents from CouchDB when a post/tag is deleted.
- [x] Implement delete functionality in CMS
- [x] Remove available languages from sibling docs when content doc is set to draft
- [x] Generate delete requests for content docs when parent post/tag doc is deleted
- [x] Delete images from S3 storage on post/tag deletion


api/src/changeRequests/validateChangeRequestAccess.ts
```diff
@@ -177,10 +299,32 @@ export async function validateChangeRequestAccess(
             };
         }
 
-        // Get the previous document to check if the default flag has been changed
-        const getRequest = await dbService.getDoc(doc._id);
-        const prevDefault =
-            getRequest.docs.length && (getRequest.docs[0] as LanguageDto).default == 1;
+        // Check if the user has assign access to all added groups in the memberOf property
+        // TODO: Write tests for this
```
DirkDevelopmentFreelance, line 303: Have to maybe make a ticket for this/Remove comment if  no longer relevant

ivanslabbert:
> Not reactively showing deleted documents changes on the app (can maybe do in seperate PR). Rest of it looks like it's working.

Need to check if it is implemented when receiving updated documents through Socket.io

api/src/changeRequests/validateChangeRequestAccess.ts
ivanslabbert: Task added #737

ivanslabbert:
> Not reactively showing deleted documents changes on the app (can maybe do in seperate PR). Rest of it looks like it's working.

This should be fixed now. Might need to add some tests for the additional logic.

api/src/changeRequests/processChangeRequest.ts
```diff
@@ -189,6 +189,19 @@ export async function processChangeRequest(
...
+            ? parentQuery.docs[0]
+            : undefined;
+
+        // We still need to replicate the parent document's memberOf property to the content document, as this is not (and should not be) done in the CMS.
+        if (parentDoc) {
+            const contentDoc = doc as ContentDto;
+            contentDoc.memberOf = parentDoc.memberOf;
+        }
+    }
+
```
ivanslabbert: We need to update sibling contentDocs with the updated list of available languages to prevent content not showing in second-choice languages (e.g. when not available in the user's first prefererred language)

api/src/changeRequests/validateChangeRequest.ts
```diff
@@ -94,7 +94,7 @@ export async function validateChangeRequest(
 }
 
 async function dtoValidate(data: any, message: string): Promise<ValidationResult> {
-    // Try-catch is needed to handle nested validation errors (speficially for arrays?) which throws an exception instead of giving a meaningful validation result.
+    // Try-catch is needed to handle nested validation errors (specifically for arrays?) which throws an exception instead of giving a meaningful validation result.
     // TODO: Might be possible to work around the exception according to https://dev.to/avantar/validating-nested-objects-with-class-validator-in-nestjs-1gn8 (see comments) - but seems like they are just handling the error in any case.
```
DirkDevelopmentFreelance, line 98: TODO left. Maybe we should create a ticket?

DirkDevelopmentFreelance (changes requested):
It correctly deletes content on the app as well now, but I had to delete local cache to see changes, this will maybe be a bit bad for users as they would have data in their local db and won't delete local cache very often.

api/src/changeRequests/validateChangeRequest.ts
ivanslabbert: I think we we do work on the API in the future we will pick it up again.

DirkDevelopmentFreelance (commented):
Hey @ivanslabbert, I still have to clear local cache to see changes

DirkDevelopmentFreelance (approved):
Nice work! Seems to be working well now

PR #710, opened 2025-02-14
Shared: Use selected languages list to remove unselected languages from indexDB

belljohan3:


shared/src/db/database.ts
```diff
@@ -143,6 +143,44 @@ class Database extends Dexie {
...
+                const languageIds = languages.map((l) => l._id);
+
+                // Identify language IDs to delete
+                const idsToDelete = languageIds.filter((id) => !selectedLanguageIds.includes(id));
+                console.log("Languages to be Deleted:", idsToDelete);
+
+                if (idsToDelete.length > 0) {
+                    try {
+                        // Delete language documents in bulk
+                        await db.docs.where("language").anyOf(idsToDelete).delete();
```
ivanslabbert, line 164: The language doc itself should not be deleted, otherwise the user will not be able to select it again in the future

shared/src/db/database.ts
```diff
@@ -143,6 +143,44 @@ class Database extends Dexie {
...
+
+                        // Delete content documents related to removed languages
+                        await db.docs
+                            .where("type")
+                            .equals(DocType.Content)
+                            .and((doc) => idsToDelete.includes(doc.language!))
+                            .delete();
+
+                        console.log("Deleted languages and related content documents.");
```
ivanslabbert, line 173: Remove console log

shared/src/db/database.ts
```diff
@@ -143,6 +143,44 @@ class Database extends Dexie {
...
+                        // Your deletion logic
```
ivanslabbert, line 174: Remove comment?

shared/src/db/database.ts
```diff
@@ -143,6 +143,44 @@ class Database extends Dexie {
...
+                // Fetch all language documents
+                const languages = await db.docs.where("type").equals(DocType.Language).toArray();
+                const languageIds = languages.map((l) => l._id);
+
+                // Identify language IDs to delete
+                const idsToDelete = languageIds.filter((id) => !selectedLanguageIds.includes(id));
+                console.log("Languages to be Deleted:", idsToDelete);
+
+                if (idsToDelete.length > 0) {
+                    try {
```
ivanslabbert, line 162: I do not think a try-catch is needed here. We need to verify the data so that there never will be an exception

ivanslabbert (changes requested):

ivanslabbert (approved):

Issue #612, opened 2025-01-09
CMS: Fix skipped test in EditContent.spec.ts: "enables post/tag settings editing when no groups are set"


ivanslabbert:

DirkDevelopmentFreelance:
May be fixed #655

PR #711, opened 2025-02-15
Remove test?

DirkDevelopmentFreelance:
My suggestion is that we remove this test? Because the post/tag settings remain enabled wether or not groups are assigned? Or maybe this ticket was not given enough of a description or the test naming is wrong? Not sure what this test is supposed to test?


ivanslabbert:
Added a more detailed description in ticket #612

DirkDevelopmentFreelance:
Opened new one

PR #714, opened 2025-02-17
712 api cms create users overview editor

belljohan3:


ivanslabbert (commented):
Hi @belljohan3 

It looks good so far and it is functional! Made some comments on things I could see that need a bit more work.

- [ ] Add tests for UserOverview.vue
- [ ] Add tests for UserTable.vue
- [x] Add tests for UserRow.vue
- [ ] Add a query option to the `search` API endpoint to query for a user by userID (or perhaps for any doc by id and type)
- [ ] The search API query is including deleteCmd documents in it's result
image
image

- [ ] Add delete button with confirmation (see EditContent)
- [ ] Show a loading indicator (see GroupOverview) while loading data from API in UserOverview

cms/src/pages/UsersPage.vue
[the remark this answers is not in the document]
belljohan3: This file is not needed.

api/src/dto/SearchReqDto.ts
```diff
@@ -64,4 +64,15 @@ export class SearchReqDto {
     @IsOptional()
     @Expose()
     languages?: Array<string>;
+
+    @IsOptional()
+    @IsString()
+    @Expose()
+    userId?: string;
```
ivanslabbert, line 71: We should be able to use DocId here?

cms/src/components/users/EditUser.vue
```diff
@@ -0,0 +1,277 @@
...
+const save = async () => {
+    // Bypass save if the user is new and marked for deletion
+    if (isNew.value && editable.value.deleteReq) {
+        return;
+    }
+
+    original.value = _.cloneDeep(editable.value);
+    editable.value.updatedTimeUtc = Date.now();
+
+    await db.upsert(editable.value);
```
ivanslabbert, line 156: When saving we should not save to the docs table, as users are read directly from the API. We do not yet have a REST API endpoint for commiting docs to the API, so this will need to be saved directly in the localChanges table for uploading via Socket.io

cms/src/components/users/EditUser.vue
```diff
@@ -0,0 +1,277 @@
...
+    }, 0);
+
+    latest ? (usersQuery.from = latest) : delete usersQuery.from;
+    const _q = await getRest().search(usersQuery);
+    _q &&
+        _q.docs &&
+        _q.docs.forEach((d: UserDto) => {
+            users.value.set(d._id, d);
+        });
+};
```
ivanslabbert, line 55: This query should be rewritten to only get one document from the API

ivanslabbert (changes requested):

api/src/db/db.service.ts
```diff
@@ -525,6 +526,10 @@ export class DbService extends EventEmitter {
                 });
             }
 
+            if (options.docId) {
+                selectors.push({ _id: options.docId });
+            }
```
ivanslabbert, line 531: Add a test to make sure this is working

ivanslabbert (approved):

PR #724, opened 2025-02-20
Make sure minification can work using an environmental variable.

DirkDevelopmentFreelance:


app/.env.example
```diff
@@ -27,4 +27,6 @@ VITE_PLUGINS=
 VITE_PLUGIN_PATH=
 
 # Hide the explore page (This is a temporary solution)
-VITE_HIDE_EXPLORE=true
\ No newline at end of file
+VITE_HIDE_EXPLORE=true
+
+VITE_MINIFY=true
```
ivanslabbert, line 32: Can you change this to something that by default will bypass minification?
e.g. `VITE_BYPASS_MINIFY`

ivanslabbert (changes requested):

app/.env.example
DirkDevelopmentFreelance: Implemented

app/vite.config.ts
```diff
@@ -48,6 +49,6 @@ export default defineConfig({
     build: {
         target: "es2015",
         sourcemap: true,
-        minify: true,
+        minify: env.VITE_BYPASS_MINIFY === "false",
```
belljohan3, line 52: Why do you assign a value here here ? Isn't supposed to just the value of that variable ?

belljohan3 (commented):
@DirkDevelopmentFreelance Check if it is possible to just read the value of the variable. otherwise it is not usefull to the variable here

app/vite.config.ts
DirkDevelopmentFreelance: vite.config runs before the npm packages are run, so enviromental variables has to be done this way because it's out of the actual app enviroment, so it works different here

ivanslabbert, line 52: Change to the following to ensure that it always runs (even without setting the environmental variable) unless set to true:
```suggestion
        minify: env.VITE_BYPASS_MINIFY !== "true",
```

ivanslabbert (changes requested):

ivanslabbert (approved):

PR #736, opened 2025-02-26
726 app clickable images

belljohan3:


app/src/components/tags/VerticalTagViewer.vue
```diff
@@ -57,6 +57,7 @@ const tagged = useDexieLiveQueryWithDeps(
                     <div class="flex items-center">
                         <div class="relative overflow-hidden rounded">
                             <LImage
+                                v-if="content.parentImageData"
```
ivanslabbert, line 60: This is preventing fallback images to show when a tag does not have an image. Fixing this will include reverting some changes on LImage and LImageProvider (setting the image prop as optional again)

app/src/components/images/ImageModal.vue
```diff
@@ -0,0 +1,127 @@
...
+    size: "post",
+});
+
+const emit = defineEmits(["close"]);
+const container = ref<HTMLDivElement | null>(null);
+
+const scale = ref(1);
+const translateX = ref(0);
+const translateY = ref(0);
+const MAX_SCALE = 5;
```
ivanslabbert, line 33: Max scale of 5 is perhaps a bit much? 2 or 3 should do I think

app/src/components/images/ImageModal.vue
ivanslabbert, line 1: Add clamping to prevent dragging the image edge too far away from the viewport edges (I would think something like max 10px from the border will be enough.

image

ivanslabbert (changes requested):

ivanslabbert:
Tested it on dev from my phone, and seems to work ok. Acts a bit strange when pinching - I think it is then dragging simultaneously. Try to disable dragging while pinching?

ivanslabbert (commented):
It works a lot better already!

The default zoom in mobile view should be more than 100% (perhaps 150%)?
image

DirkDevelopmentFreelance (changes requested):
It looks good on desktop mode but I had some thoughts regarding desktop mode

Maybe the escape button can be pressed and it will exit out of the modal, as right now if the image is fully zoomed on desktop mode it is hard to exit it quickly
The dragging is extremely slow on desktop, and there is a weird dragging visual artifact that happens when trying to drag on desktop mode.

On mobile it works and looks nice! Nice work @belljohan3!

ivanslabbert (approved):

PR #740, opened 2025-02-26
Css compatibility

DirkDevelopmentFreelance:


app/src/main.ts
```diff
@@ -75,3 +78,40 @@ async function Startup() {
...
+            .process(css, { from: undefined })
+            .catch((err) => {
+                console.error("PostCSS processing error", err);
+            });
+    }
+}
+
+if (import.meta.env.DEV == true) {
+    checkCSSCompatibility();
+}
```
ivanslabbert, line 117: Nice work @DirkDevelopmentFreelance 

But I think this should be in a separate CSS testing file (imported into main.ts). This separate file should only be imported into main.ts if the project is running in dev mode. The separate CSS testing file can be excluded from production builds.

ivanslabbert (changes requested):

app/src/main.ts
DirkDevelopmentFreelance: Made the changes

ivanslabbert (approved):
I see the tests are failing on an out-of-sync package-lock.json file.
Might have to revert the package-lock.json and package.json files and re-install the doiuse package?

PR #754, opened 2025-03-05
Fix EditContentText Reactivity

DirkDevelopmentFreelance:


ivanslabbert (changes requested):
Works as expected, but I see there is a test on the RichTextEditor component failing...

ivanslabbert (approved):

PR #763, opened 2025-03-10
Cms keep filter search life cycle

MrDirkelz:


DirkDevelopmentFreelance:
TODO:
 - [ ] Fix the new LSelect component
 - [ ] Create a slot in parent, so we have a row below the search bar row
 - [ ] Fix bug where some filter options are not always working
 - [ ] Fix bug where when you click outside the new component it adds the top most one
 - [ ] Fix margin in new component
 - [ ] Fix shadows in top left corner
 - [ ] Write suitable tests where needed with new changes

PR #770, opened 2025-03-10
App: Fix Language Selection Bug

DirkDevelopmentFreelance:


app/src/components/navigation/LanguageModal.vue
```diff
@@ -19,6 +19,18 @@ const { t } = useI18n();
 
 const languages = db.whereTypeAsRef<LanguageDto[]>(DocType.Language, []);
 
+watch(
+    languages,
+    () =>
+        appLanguageIdsAsRef.value.filter((id) => {
```
ivanslabbert, line 25: A filter does not overwrite the array it is filtering, but returns the filtered array. So this routine is not doing anything?

ivanslabbert (commented):
I think the filtering should happen in globalConfig to correct the problem at source. I suggest adding some logic in the cmsLanguages watcher that overwrites `appLanguageIdsAsRef` with an array that only includes languages available in the CMS as soon as the CMS languages are available.

[](https://github.com/bccsa/luminary/blob/f052a6508de21ff640d7698f8eac0a61d8fe780c/app/src/globalConfig.ts#L83)

ivanslabbert (approved):

PR #777, opened 2025-03-17
APP:  Only show "translation not avaiable" banner  when opening directly to a slug

DirkDevelopmentFreelance:


app/src/pages/SingleContent.vue
```diff
@@ -205,15 +205,18 @@ watch(
...
-                    id: "translation-not-published",
-                    title: t("notification.content_not_available.title"),
-                    description: t("notification.content_not_available.description", {
-                        language: appLanguagesPreferredAsRef.value[0].name,
-                    }),
-                    state: "error",
-                    type: "toast",
-                });
+                //Only display notification when directly navigating to a slug
+                if (router.currentRoute.value.name == "content") {
```
ivanslabbert, line 209: It does not seem like this is working. I am still getting a notification when navigating from the home page to a post which is not translated.

ivanslabbert (approved):

PR #782, opened 2025-03-19
CMS: Add group membership filter to ContentOverview

MrDirkelz:


ivanslabbert (commented):
The group selector is now not working at all (the checkboxes in LChecklist are not showing either?). Perhaps the group selector should also work in the same way as the tag selector by showing the selected groups as labels below.

LChecklist and LCombobox are now essentialy the same, so we should rather just use one component? (Have made some nice updates to LCombobox in #714 so I suggest rather going for LCombobox instead of improving LChecklist further)

ivanslabbert:
Please move the files back their original folders? Restructuring should be done in it's own PR during a time where there is not much movement in the project to aviod a lot of conflicts

cms/src/components/content/ContentRow.vue
```diff
@@ -27,7 +30,35 @@ const contentDocs = db.whereParentAsRef(props.contentDoc.parentId, props.parentT
...
+        db.docs
+            .where("parentId")
+            .anyOf(_props.contentDoc.parentTags)
+            .filter((t) => {
+                if (
+                    t.language === cmsLanguageIdAsRef.value ||
+                    t.language === cmsDefaultLanguage.value
+                )
+                    return true;
+                return false;
```
ivanslabbert, line 45: Can use `getPreferredContentLanguage` to filter the returned contet docs (see `util` folder)

cms/src/components/users/EditUser.spec.ts
```diff
@@ -103,7 +103,7 @@ describe("EditUser.vue", () => {
         });
     });
 
-    it("should update and save the current user", async () => {
+    it.skip("should update and save the current user", async () => {
```
ivanslabbert, line 106: skipped test?

DirkDevelopmentFreelance:
Created a new PR, as this one had unresolvable conflicts

PR #783, opened 2025-03-19
Only sync deleteCmd documents when doing an update sync

ivanslabbert:
deleteCmd documents are not synced on initial sync as the API will not return previously deleted documents, and no cleanup is needed in the client.
Also added an export with a ref indicating when sync is active


DirkDevelopmentFreelance (approved):
Everything seems to work as expected

PR #784, opened 2025-03-19
Cms-keep-filter-search-life-cycle-2

ivanslabbert:


ivanslabbert:
State saving seems to work well. There is however an issue with the tag selector where the first tag in the list is always selected, no matter which tag is clicked on...
image

ivanslabbert:
- [ ] Remove unused "state" condition

image

DirkDevelopmentFreelance (approved):

PR #785, opened 2025-03-19
App: Add Multiple fallback images support

DirkDevelopmentFreelance:


DirkDevelopmentFreelance (commented):
Fallback images can be changed with other suggestions

ivanslabbert (commented):
Looks good @DirkDevelopmentFreelance 

See there is a conflict that should be resolved. Also check notes below on fallback images included in the Luminary repo.

---

Please make sure default fallback images are royalty free / not copyrighted? Also encode them as webp and not jpeg.
I would just use nature images and not images of people to keep it as generic as possible.

ivanslabbert (changes requested):
Looks nice with the images!

Have some change requests however.

app/src/components/images/LImage.vue
```diff
@@ -39,6 +42,19 @@ const rounding = {
...
+
+onBeforeMount(() => {
+    if (!_fallbackImages.value.includes({ seed: fallbackImageSeed })) {
+        const randomIndex = Math.floor(Math.random() * fallbackImages.length);
+        const randomUrl = fallbackImages[randomIndex];
+        _fallbackImages.value = [
+            ..._fallbackImages.value,
+            { url: randomUrl, seed: fallbackImageSeed },
+        ];
+    }
```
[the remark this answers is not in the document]
DirkDevelopmentFreelance: I will have to find a different way to handle it then, as I am trying to use the contentparentid to link the images throughout the webpage so it's the same image on each page for each content

app/src/main.ts
```diff
@@ -41,8 +41,8 @@ async function Startup() {
         token,
         appLanguageIdsAsRef,
         docTypes: [
-            { type: DocType.Tag, contentOnly: true, syncPriority: 2 },
-            { type: DocType.Post, contentOnly: true, syncPriority: 2 },
+            { type: DocType.Tag, contentOnly: false, syncPriority: 2 },
```
DirkDevelopmentFreelance, line 44: I made this false, as having it foreach on each content document will become slower much faster, and as well as the fact it would then populate the array 27 times, x times, depending on the amount of contentdocs. so directly using parent documents makes the amount of documents that has to be looped through less. 
If someone can think of a faster solution please suggest it.

DirkDevelopmentFreelance: Reverted and did something else

DirkDevelopmentFreelance (commented):
image
image

Result on my side. Working as expected to me. The images are persisting, and the code looks ok, could be a faster way maybe? But this was the best way to only have it look through unique parent Ids and not through each content doc, which makes it faster.

app/src/globalConfig.ts
```diff
@@ -320,3 +327,61 @@ export const loginModalVisible = ref(false);
...
+            fallbackImages.value = [
+                ...fallbackImages.value,
+                { id: randomDoc.parentId, url: randomUrl },
+            ];
+        });
+    });
+};
+watch(fallbackImages, () => {
+    localStorage.setItem("_fallback_images", JSON.stringify(fallbackImages.value));
+});
```
belljohan3, line 387: remove this watch to avoid troubles if it gets remove from the local storage ?

belljohan3 (dismissed):

DirkDevelopmentFreelance:
The flashing fallback image does not seem to be an issue in this PR anymore. Tested on 20x CPU slowdown and slow 4G network with caching disabled.

app/src/globalConfig.ts
```diff
@@ -315,3 +323,54 @@ export const loginModalVisible = ref(false);
...
+                fallbackImages.value = [
+                    ...fallbackImages.value,
+                    { id: doc.parentId, url: randomUrl },
+                ];
+
+                processedParentDocs.add(doc.parentId);
+            });
+        },
+    );
+};
```
ivanslabbert, line 376: Move to own function in utils folder

app/src/components/images/LImage.vue
ivanslabbert, line 1: Do we need tests?

app/src/globalConfig.ts
```diff
@@ -315,3 +323,54 @@ export const loginModalVisible = ref(false);
...
+
+                fallbackImages.value = [
+                    ...fallbackImages.value,
+                    { id: doc.parentId, url: randomUrl },
+                ];
+
+                processedParentDocs.add(doc.parentId);
+            });
+        },
+    );
```
ivanslabbert, line 375: This should happen on-demand as images are loaded, and not at app startup for all images. Otherwise it will affect the initial app loading time

app/src/components/images/LImage.vue
```diff
@@ -48,12 +50,20 @@ onMounted(() => {
...
+    if (!currentFallbackImage) return;
+    return currentFallbackImage.url;
+});
 </script>
 
 <template>
     <div ref="parentRef" :class="sizes[size]">
         <div
-            :style="{ 'background-image': 'url(' + fallbackImg + ')' }"
+            :style="{ 'background-image': 'url(' + fallbackImage + ')' }"
```
ivanslabbert, line 66: Only set fallback background image if:

- No image is set
- An image is set, and both the primary and secondary images failed to display (e.g. due to being offline)

ivanslabbert (commented):
- [x] Replace animal pics with nature-only pics (does not need to be African only)
- [ ] All images to be 640px wide
- [ ] Check client cache control headers settings in nginx (in docker) for fallback images

app/src/components/images/LImage.vue
DirkDevelopmentFreelance: Added a test. Not sure if more should be added.

DirkDevelopmentFreelance:
> * Check client cache control headers settings in nginx (in docker) for fallback images
I added a line of code in the nginx.conf to re-evaluate the fallbackImages caching. Not sure if this is what you meant?

ivanslabbert (approved):
Approved, but please fix a few small issues?

PR #787, opened 2025-03-21
738 app logout confirmation dialog

belljohan3:


app/src/components/common/LDialog.vue
ivanslabbert, line 1: This could have been done in a simpler way by using LModal. We can still have the LDialog component, but LDialog can use LModal internally

ivanslabbert (changes requested):

ivanslabbert (approved):

PR #789, opened 2025-03-24
Add test to RichTextEditor

DirkDevelopmentFreelance:


cms/src/components/editor/RichTextEditor.spec.ts
```diff
@@ -16,4 +16,30 @@ describe("RichTextEditor", () => {
...
+        editor.commands.setContent("Testing Testing 123");
+
+        editor.options.onUpdate?.({ editor });
+
+        await waitForExpect(() => {
+            //@ts-expect-error
+            const updatedText = JSON.parse(wrapper.vm.text);
+            expect(updatedText.content[0].content[0].text).toBe("Testing Testing 123");
+        });
+    });
```
ivanslabbert, line 44: This test should rather test if the document has been updated in IndexedDB. At the moment it just tests if the text is showing visually, but the issue we had (and need to test to prevent happening again) is that the updated text in the RichTextEditor was never saved to the editable object, and hence not saved back to IndexedDb

ivanslabbert (changes requested):

ivanslabbert (approved):

PR #793, opened 2025-03-25
APP: fix scrolling

belljohan3:


app/src/App.vue
```diff
@@ -84,14 +84,21 @@ onErrorCaptured((err) => {
     console.error(err);
     Sentry.captureException(err);
 });
+
+onMounted(() => {
+    document.querySelector("main")?.focus(); // Focus to enable keyboard scrolling
+});
```
ivanslabbert, line 90: I think this needs to be done every time after routing. After clicking on a navigation button in the menu bar scrolling is currently disabled, probably because the button in the top bar has focus.

app/index.html
```diff
@@ -7,7 +7,7 @@
         <title>Loading...</title>
     </head>
     <body
-        class="overscroll-none bg-white text-slate-900 dark:bg-slate-900 dark:text-slate-50"
+        class="bg-white text-slate-900 dark:bg-slate-900 dark:text-slate-50"
```
ivanslabbert, line 10: Do not think we need to enable overscroll. We are trying to give the Web-app an app-like feel.

ivanslabbert (changes requested):

app/index.html
belljohan3: Then we won't be able to scroll with keyboard. Because actually it is preventing the scroll with keyboard as soon as the document is ready

ivanslabbert (approved):

PR #796, opened 2025-03-26
APP: Update VideoPlayer to hide volume control on mobile devices

belljohan3:


DirkDevelopmentFreelance (dismissed):
Nice work @belljohan3.

I have one suggestion..Maybe there could be a button next to the timeline at the bottom at the left hand side, and when you click it the volume picker appears. This would make it more consistent regarding the fact that every video platform does this, and users would most likely look there first as that is what they are used to.

ivanslabbert (approved):

PR #808, opened 2025-03-28
Add test to group selector to only display groups to which the user has assign access

ivanslabbert:


DirkDevelopmentFreelance (approved):
Looks good to me.

PR #812, opened 2025-04-02
CMS: save changes in draft mode with no publish access

belljohan3:


ivanslabbert (changes requested):
This looks like it will work, but I think we should rather disable the controls if the user cannot edit it (e.g. disable the publish button if the user does not have publish access, and disable all the editing components if the content doc is published and the user does not have publish access)

In addition we can display an information icon with a message "no publish access" in the parent  validation pane so that the user is informed.

Issue #791, opened 2025-03-25
APP: Review CSS Warnings


DirkDevelopmentFreelance:
Think about how far back we want versioning support to go for now. Once decided, change browser versions to those and review css warnings in the console. Fix them if they can be fixed. If they can't find a better alternative and implement it as a suggestion.

DirkDevelopmentFreelance:
CSS Warnings for browsers from early 2023 onward:

1. ui-serif, ui-sans-serif, ui-monospace and ui-rounded values for font-family (extended-system-fonts)
2. text-decoration styling only partially supported
3. CSS text-indent only partially supported
4. CSS-resize property not supported for Safari on iOS
5. CSS3 Multiple column layout only partially supported (multicolumn)
6. CSS scrollbar styling not supported on safari iOS, and only partially supported on other browsers
7. CSS cursors not supported on safari on iOS and samsung internet
8. Intrinsic-width only partially supported by firefox
9. CSS touch-action property not supported by Safari

Question:
Could it be running through all tailwindCSS classes seen as it is imported into the CSS file? And in reality not looking at the used classes we actually use

PR #814, opened 2025-04-03
Review CSS warnings

DirkDevelopmentFreelance:


app/src/pages/NotFoundPage.vue
```diff
@@ -22,12 +22,14 @@ const { isAuthenticated } = useAuth0();
                 }}
             </p>
             <div v-if="isAuthenticated" class="mt-10 flex items-center justify-center gap-x-6">
-                <RouterLink to="/" class="text-yellow-700 underline"> Back to home </RouterLink>
+                <RouterLink to="/" class="border-b-[1px] border-yellow-700 text-yellow-700">
```
laurensgroeneveld, line 25: What's the reason for avoiding `underline`? That's one of the most basic CSS features so should be supported 😄

app/postcss.config.js
```diff
@@ -1,6 +1,23 @@
...
+                "chrome >= 109",
+                "firefox >= 109",
+                "edge >= 109",
+                "opera >= 95",
+                "safari >= 16.3",
+                "> 1%",
+            ],
+            onFeatureUsage: (usageInfo) => {
+                throw new Error(`Unsupported CSS Detected: ${usageInfo.featureData.title}`);
+            },
```
laurensgroeneveld, line 20: Personally I'd argue that using unsupported CSS features is not that problematic, as long as you're mindful of them and think of fallbacks for browsers that don't support that feature. One popular mobile browser is Opera Mini for example, which deliberately doesn't support many features to save bandwidth or processing power. In that case I'd say it's expected the site might have a degraded experience, even though it should still be usable.

app/tailwind.config.ts
```diff
@@ -8,19 +8,7 @@ export default {
...
-                "bcc-brand-50": "#f5f8f8",
-                "bcc-brand-100": "#eaeeef",
-                "bcc-brand-200": "#d2dadb",
-                "bcc-brand-300": "#acbdbd",
-                "bcc-brand-400": "#8ea6a6",
-                "bcc-brand-500": "#437571",
-                "bcc-brand-600": "#004e48",
-                "bcc-brand-700": "#023d38",
-                "bcc-brand-800": "#021f1c",
-                "bcc-brand-900": "#020a0b",
```
laurensgroeneveld, line 23: Is it intended to remove these colors? They are still used in the `LButton` for the `bcc` variant button.

laurensgroeneveld (commented):
Was looking at the PR list and the title piqued my interest 😄 From the accompanying ticket (#791) it seems the underlying goal is not to use CSS features that are not fully supported in all browsers. Just left a few comments with that in mind.

app/tailwind.config.ts
DirkDevelopmentFreelance: These colors will also be removed in LButton... think they should've been removed in an earlier ticket as we want to have no Organisation specific references in the project

DirkDevelopmentFreelance:
> Was looking at the PR list and the title piqued my interest 😄 From the accompanying ticket (#791) it seems the underlying goal is not to use CSS features that are not fully supported in all browsers. Just left a few comments with that in mind.

Thanks @laurensgroeneveld! Glad this ticket peaked your interests as the input helps a lot on this one. Lots of CSS features, like for example "underline" that are unsupported on some browsers, or only partially supported on others, but aren't that breaking and won't really make a difference persay; Atleast I can edit the script to ignore certain css features, so we can have that control atleast. Once again, thanks for reviewing! It really helped to get a clearer picture on this ticket

app/src/pages/NotFoundPage.vue
laurensgroeneveld: Turns out it's not as well supported as I thought surprisingly: https://caniuse.com/mdn-css_properties_text-decoration-line_underline - but I'd still argue that this would be fine to use, since it's not that big a problem if it doesn't show up in some browsers (as long as your links also have a different color).

app/tailwind.config.ts
laurensgroeneveld: Fair just wanted to note that currently it's still used, so should not be removed yet 😄.

DirkDevelopmentFreelance:
Test failing

app/tailwind.config.ts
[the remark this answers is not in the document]
DirkDevelopmentFreelance: This seems to already have been removed

api/src/db/seedingDocs/lang-eng.json
```diff
@@ -80,6 +80,7 @@
         "logout.modal.title": "Logout",
         "logout.modal.description": "Are you sure you want to log out?",
         "logout.modal.button_cancel": "Cancel",
-        "logout.modal.button_logout": "Logout"
+        "logout.modal.button_logout": "Logout",
+        "navigation.cta.home": "Back to home"
```
ivanslabbert, line 84:
```suggestion
        "notfoundpage.navigation.home": "Back to home"
```
We have to think about the translators, that they will easily be able to know where the text is use din the app

api/src/db/seedingDocs/lang-fra.json
```diff
@@ -80,6 +80,7 @@
         "logout.modal.title": "Déconnexion",
         "logout.modal.description": "Êtes-vous sûr de vouloir vous déconnecter ?",
         "logout.modal.button_cancel": "Annuler",
-        "logout.modal.button_logout": "Déconnexion"
+        "logout.modal.button_logout": "Déconnexion",
+        "navigation.cta.home": "Retour à la maison"
```
ivanslabbert, line 84: Same as previous comment

api/src/db/seedingDocs/lang-fra.json
```diff
@@ -80,6 +80,14 @@
...
-        "logout.modal.button_logout": "Déconnexion"
+        "logout.modal.button_logout": "Déconnexion",
+        "notfoundpage.navigation.home": "Retour à la maison",
+        "notfoundpage.unauthenticated.title": "Oups ! Vous n'êtes pas connecté.",
+        "notfoundpage.unauthenticated.loginPrompt.before": "Veuillez cliquer",
+        "notfoundpage.unauthenticated.loginPrompt.linkText": "ici",
+        "notfoundpage.unauthenticated.loginPrompt.after": "pour vous connecter ou créer un compte.",
+        "notfoundpage.unauthenticated.description": "Le contenu peut être uniquement disponible pour les utilisateurs connectés.",
+        "notfoundpage.authenticated.title": "Hé, nous n'avons pas pu trouver cette page",
+        "notfoundpage.authenticated.description": "Le contenu que vous recherchez a peut-être été supprimé ou n'est pas disponible pour le moment."
```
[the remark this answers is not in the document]
DirkDevelopmentFreelance: @belljohan3 I am still waiting for the review on these translations...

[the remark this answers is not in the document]
belljohan3: After reviewing, i think the translation is fine like this for now. @DirkDevelopmentFreelance can merge it.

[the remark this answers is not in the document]
DirkDevelopmentFreelance: Thanks @belljohan3!

PR #815, opened 2025-04-03
cms-group-selector-updates

belljohan3:


cms/src/components/groups/GroupSelector.vue
```diff
@@ -1,6 +1,6 @@
 <script setup lang="ts">
-import { computed, ref } from "vue";
-
+// Import necessary utilities and types from Vue and shared libraries
```
ivanslabbert, line 2: Remove comment

ivanslabbert (approved):
This seems to work as expected.

We need to do the same / similar for the TagSelector. But this can be on a separate PR, unless it is quick to fix here.
image

belljohan3:
> This seems to work as expected.
> 
> We need to do the same / similar for the TagSelector. But this can be on a separate PR, unless it is quick to fix here. image

Issue created #817

PR #827, opened 2025-04-08
Fix Group changes not showing in the validation component and not reverted when reverting in EditContent bug

DirkDevelopmentFreelance:


cms/src/components/content/EditContent.vue
```diff
@@ -34,7 +34,7 @@ import EditContentStatus from "@/components/content/EditContentStatus.vue";
 import EditContentBasic from "@/components/content/EditContentBasic.vue";
 import EditContentText from "@/components/content/EditContentText.vue";
 import EditContentVideo from "@/components/content/EditContentVideo.vue";
-import EditContentParentValidation from "@/components/content/EditContentParentValidation.vue";
+import EditContentParentValidation from "../../components/content/EditContentParentValidation.vue";
```
ivanslabbert, line 37: Not sure if this change was intentional?

PR #828, opened 2025-04-08
feat: add link management functionality to rich text editor

belljohan3:


cms/src/components/editor/RichTextEditor.spec.ts
```diff
@@ -59,4 +59,22 @@ describe("RichTextEditor", () => {
...
+            },
+        });
+
+        await waitForExpect(() => {
+            expect(wrapper.text()).toContain("Gandalf the Grey");
+            // expect the word to be a link
+            expect(wrapper.find("a").exists()).toBe(true);
+            expect(wrapper.find("a").attributes("href")).toBe("https://example.com");
+        });
+    });
```
ivanslabbert, line 77: This test is not testing if pasting links is working. If not possible / very difficult to test pasting (eg. from Word), then we can rather just remove this test.

ivanslabbert (commented):
General:
When clicking the link button, the link modal opens but is not focussed. The text input on the modal should get focus.

ivanslabbert (approved):

PR #831, opened 2025-04-10
Added v-if to button for access control. Ran test for functionality.

Rapotukganakga:


cms/src/components/content/EditContent.spec.ts
```diff
@@ -535,5 +535,22 @@ describe("EditContent.vue", () => {
...
+            const wrapper = mount(EditContent, {
+                props: {
+                    docType: DocType.Post,
+                    id: mockData.mockPostDto._id,
+                    languageCode: "eng",
+                    tagOrPostType: PostType.Blog,
+                },
+            });
+            await waitForExpect(async() => {
+                const deletebutton = wrapper.find('[data-test="delete-button"]')
```
ivanslabbert, line 551: This const is not used? Not sure what the reason is for the const?

ivanslabbert (changes requested):
Please check that the visibility of the translation delete icons are also set correctly?
image

cms/src/components/content/EditContent.spec.ts
```diff
@@ -535,5 +535,22 @@ describe("EditContent.vue", () => {
                 });
             });
         });
+
+        it("Check if the user does not have delete access", async () => {
+        delete accessMap.value["group-public-content"].post?.delete
```
DirkDevelopmentFreelance, line 540: Does this work? As I actually never thought of doing it this way, if it works you just tought me something @Rapotukganakga, nice.

ivanslabbert (commented):
Few more comments...

cms/src/components/content/EditContentValidation.vue
```diff
@@ -32,6 +36,12 @@ const sortedLanguages = computed(() => {
 });
 
 const showDeleteModal = ref(false);
+
+const canDeleteLang = computed(() => {
+    if (!editableContent.value) return false;
+    return verifyAccess(editableContent.value.memberOf, DocType.Language, AclPermission.Delete, "all");
```
ivanslabbert, line 42: Think this should be:
```suggestion
    return verifyAccess(editableContent.value.memberOf, editableContent.value.parentType, AclPermission.Delete, "all");
```

api/src/db/seedingDocs/group-public-users.json
```diff
@@ -26,7 +26,7 @@
         {
             "type": "language",
             "groupId": "group-super-admins",
-            "permission": ["view", "edit", "delete", "assign", "translate"]
+            "permission": ["view", "edit", "assign", "translate"]
```
ivanslabbert, line 29: Please revert this change? Super admins need full access to other groups

ivanslabbert (approved):
Approved, but please fix the outstanding change request

PR #832, opened 2025-04-10
817 cms tag selector updates

belljohan3:


ivanslabbert (changes requested):
It works correctly when the user has view access to a tag without assign access, showing the tag without a remove button.

It however does not work correctly when the user does not have view access to the tag. In this case the tag's UUID must be shown (which is not happening at the moment).

DirkDevelopmentFreelance (commented):
It looks good, but why do we need to show the UUID's of the tags? To me it feels wrong to display data like that to users. But the ticket itself looks good to me.

ivanslabbert:
> It looks good, but why do we need to show the UUID's of the tags? To me it feels wrong to display data like that to users. But the ticket itself looks good to me.

Users need to know that there are tags set (even if they do not have access to see the tags). This is perhaps more important for Groups than for Tags, but it will help to avoid confusion

ivanslabbert:
> > It looks good, but why do we need to show the UUID's of the tags? To me it feels wrong to display data like that to users. But the ticket itself looks good to me.
> 
> Users need to know that there are tags set (even if they do not have access to see the tags). This is perhaps more important for Groups than for Tags, but it will help to avoid confusion

Looking more into this... It is not practical to show tags of which we do not have the tag content documents, as it will show those tags under all tag type selectors, making it more confusing for the users

ivanslabbert (approved):

belljohan3:
@ivanslabbert everything look good for me.

PR #834, opened 2025-04-11
Delete drafted documents on sync on non-cms clients

MrDirkelz:


DirkDevelopmentFreelance:
Reopened as I accedintally used this PR again to push

DirkDevelopmentFreelance:
This implementation is not required as it seems to already not include drafted documents on non-cms clients

PR #845, opened 2025-04-15
Remove the div for a canPublish tmp render for debugging

MrDirkelz:


johan-bell (commented):
Hi @DirkDevelopmentFreelance i will close this for now as it is already removed.

PR #863, opened 2025-04-29
Add group membership filter to content overview in the CMS

DirkDevelopmentFreelance:


DirkDevelopmentFreelance:
Display the groups in the ContentTable

cms/src/components/content/ContentOverview/ContentOverview.vue
```diff
@@ -36,6 +39,7 @@ import LRadio from "../../forms/LRadio.vue";
 import { cmsLanguageIdAsRef } from "@/globalConfig";
 import LChecklist from "@/components/forms/LChecklist.vue";
```
ivanslabbert, line 40: Unused import

cms/src/components/content/ContentOverview/ContentOverview.vue
```diff
@@ -348,6 +377,35 @@ const resetQueryOptions = () => {
...
+        >
+            <ul class="flex w-full flex-wrap gap-2">
+                <TransitionGroup
+                    enter-active-class="transition duration-150 delay-75"
+                    enter-from-class="transform scale-90 opacity-0"
+                    enter-to-class="transform scale-100 opacity-100"
+                    leave-active-class="transition duration-100"
+                    leave-from-class="transform scale-100 opacity-100"
+                    leave-to-class="transform scale-90 opacity-0"
+                >
```
ivanslabbert, line 392: Remove transition group?

cms/src/components/content/TagSelector.spec.ts
```diff
@@ -108,6 +108,7 @@ describe("TagSelector.vue", () => {
     });
 
     it("can add tags to the passed Parent document", async () => {
+        //This test is not passing correctly, it is giving a false positive.
```
ivanslabbert, line 111: We need to look a bit more into this

cms/src/components/forms/LCombobox.vue
```diff
@@ -71,45 +69,71 @@ watch(showDropdown, () => {
...
+const toggleDropdown = () => {
+    showDropdown.value = !showDropdown.value;
+    nextTick(() => {
+        inputElement.value?.focus();
+    });
+};
 </script>
 
 <template>
+    {{ selectedOptions }}
```
ivanslabbert, line 84: remove test code

cms/src/components/forms/LCombobox.vue
```diff
@@ -165,7 +187,7 @@ const selectedLabels = computed(() => {
                 ]"
                 @click="
                     () => {
-                        if (!option.selected) {
+                        if (option.selected !== true) {
```
ivanslabbert, line 190: Not needed to have a comparitor here? `!` should work fine?

cms/src/components/forms/LCombobox.vue
```diff
@@ -178,14 +200,18 @@ const selectedLabels = computed(() => {
...
+            class="mt-3 flex flex-wrap gap-3"
+        >
             <LTag
                 v-for="option in selectedLabels"
                 :key="option.id"
                 @remove="
                     () => {
                         if (option.isRemovable !== false) {
-                            selectedOptions.splice(selectedOptions?.indexOf(option.value), 1);
+                            selectedOptions.splice(selectedOptions?.indexOf(option.id), 1);
```
ivanslabbert, line 214: you need to remove the same field as being added (e.g. in line 191). Using `id` can probably work, but needs to be implemented throughout in that case

cms/src/globalConfig.ts
```diff
@@ -41,9 +41,10 @@ export async function initLanguage() {
         );
         const defaultLang = languages.find((lang) => lang.default === 1);
 
-        cmsLanguageIdAsRef.value = preferredLang?._id || defaultLang?._id || "";
-    } else {
-        cmsLanguageIdAsRef.value = languages.filter((lang) => lang.default === 1)[0]._id;
+        cmsLanguageIdAsRef.value =
+            preferredLang?._id || defaultLang?._id || languages[0]?._id || "";
+
+        console.log("CMS languageID", cmsLanguageIdAsRef.value);
```
ivanslabbert, line 47: This should probably not have changed?

DirkDevelopmentFreelance:
> Unable to select groups: image

I was able to reproduce the bug for the group membership filter, I know why it didn't work for you, you have to just first clear the session storage or reset the query in the cms so that it gets the new groups field in the query options that was added, because before it was using the session storage of the query options before the groups field was added, so the LCombobox does not have an array to pass the groups to, because it was undefined, resulting in this warning and the errors that you got. That's why it worked on my side but not on yours, it also would not have worked for someone else if they tried to use it. And that is a problem, because let's say we merge this ticket, how will the users know to reset the query or something like that, and then the filter won't work for them? Should we add a triangle information icon next to the "reset query" button so that users click it so it can get the updated queryOptions with the required field for the groups? And in case we add more stuff to the query later?

cms/src/components/content/TagSelector.spec.ts
DirkDevelopmentFreelance: Yes, and also why it happens, so that we know for future case scenarios

ivanslabbert (approved):
Set default empty tag array to filter loaded from session storage if it is unset / undefined

PR #864, opened 2025-04-29
Map-user-memberOf-to-user-session-permissions

ivanslabbert:


api/src/dto/UserDto.ts
```diff
@@ -15,4 +15,12 @@ export class UserDto extends _contentBaseDto {
...
     @Expose()
     name: string;
+
+    /**
+     * External user ID for mapping to external systems
+     */
+    @IsOptional()
+    @IsString()
+    @Expose()
+    userId: string;
```
oslabbie, line 25: Make optional with ?

oslabbie (approved):

PR #866, opened 2025-04-30
Fix router not correctly redirecting to router paths

DirkDevelopmentFreelance:


app/src/router/index.ts
```diff
@@ -6,6 +6,7 @@ import SingleContent from "@/pages/SingleContent.vue";
 import ExplorePage from "@/pages/ExplorePage.vue";
 import BookmarksPage from "@/pages/BookmarksPage.vue";
 import VideoPage from "@/pages/VideoPage.vue";
+import { db, DocType, useDexieLiveQuery, type RedirectDto } from "luminary-shared";
```
ivanslabbert, line 9: unused import

app/src/router/index.ts
```diff
@@ -81,4 +82,24 @@ const router = createRouter({
     ],
 });
 
+router.beforeEach(async (to) => {
```
ivanslabbert, line 85: We can probably remove routing of redirects from SingleContent?

DirkDevelopmentFreelance: Seems to still be working with it not in SingleContent

PR #868, opened 2025-04-30
Fix API reprocessing image when saving in EditContent.vue

MrDirkelz:


cms/src/components/content/EditContent.vue
```diff
@@ -240,6 +240,21 @@ const isDirty = computed(
 
 const isValid = ref(true);
 
+const saveBtnText = ref("Save");
+
+const refreshParent = () => {
+    saveBtnText.value = "Saving";
+    // Use a timeout to wait for the API to process the new data, and then refresh the parent document
+    // This prevents things like the image from reprocessing in the API
+    setTimeout(() => {
```
ivanslabbert, line 249: Using a timer here will be unpredictable - we do not know how long it will take to get an API response

A better approach could be to

1. Use a useDexieLiveQuery to get live updates of the parent document from IndexedDB
2. When the image field updates in the database version of the parent document, patch the `editableParent` with the update image data

DirkDevelopmentFreelance:
Change mock express api port to be a random port to change it each time

cms/src/components/content/EditContent.spec.ts
```diff
@@ -256,36 +259,50 @@ describe("EditContent.vue", () => {
...
+
+            // Wait for the component to fetch data
+            await waitForExpect(() => {
+                expect(wrapper.find('input[name="title"]').exists()).toBe(true); // EditContentBasic is rendered
+                expect(wrapper.html()).toContain("Text content"); // EditContentText is rendered
+                expect(wrapper.html()).toContain("Video"); // EditContentVideo is rendered
+                expect(wrapper.find('button[data-test="save-button"]').exists()).toBe(true); // EditContentParentValidation is rendered
+            });
+        },
+        { timeout: 999999 },
```
ivanslabbert, line 304: remove timeout

ivanslabbert (commented):
Can you add a test to make sure the logic is working as expected? (Create new ticket?)

ivanslabbert (approved):

DirkDevelopmentFreelance:
Bad PR

PR #872, opened 2025-05-06
Remove duplicate change request functionality

DirkDevelopmentFreelance:
Notes:
Change request submission to Rest Endpoint is in ``RestApi.ts``
We want the change requests to only go to the rest endpoints
There is a SubscribeMessage for the change request submission in socketio.ts in the API, 
The shared library has a socket that also sends a change request submission

The change request in socket.io has a function "pushLocalChange", which emits a change request to the socket in the api
Is this duplicate functionality, or should this still be replaced with a changeRequest to the REST endpoint?


api/src/socketio.spec.ts
```diff
@@ -2,19 +2,19 @@ import { Socket, io } from "socket.io-client";
...
 import { DbService } from "./db/db.service";
 
 describe("Socketio", () => {
     const oldEnv = process.env;
     let server: Socketio;
     let client: Socket;
     let app: INestApplication;
-    let db: DbService;
+    //db may still be needed in the future, so keeping it here for now
+    let _db: DbService;
```
ivanslabbert, line 13: If not used anymore, remove. We can always look at the file history to see how it was done before

api/src/socketio.ts
ivanslabbert, line 1: Remove unused imports

shared/src/socket/socketio.spec.ts
ivanslabbert, line 1: The offline changes uploading logic (and tests) should move to it's own file (if it is still part of the shared library socket.io file). It has nothing to do with socket.io anymore, so it is best to separate it from the socket.io logic.

The tests in this file that has to do with offline changes uploading should still test the same functionality, but against the rest API instead of using the previous Socket.io message

shared/src/socket/socketio.ts
```diff
@@ -154,7 +155,14 @@ class SocketIO {
...
 
-        this.socket.emit("changeRequest", localChange);
+        const res = await getRest().changeRequest({
+            id: localChange.id,
+            doc: localChange.doc,
+        } as ChangeRequestQuery);
+
+        if (res) {
+            this.handleAck(res as ChangeReqAckDto);
+        }
```
ivanslabbert, line 165: Move to separate file (does not belong with Socket.io anymore)

ivanslabbert (commented):
- [x] check if image uploading is stil working correctly

api/.eslintrc.js
```diff
@@ -18,5 +18,12 @@ module.exports = {
         "@typescript-eslint/explicit-function-return-type": "off",
         "@typescript-eslint/explicit-module-boundary-types": "off",
         "@typescript-eslint/no-explicit-any": "off",
+        "@typescript-eslint/no-unused-vars": [
+            "warn",
+            {
+                argsIgnorePattern: "^_",
+                varsIgnorePattern: "^_",
+            },
+        ],
```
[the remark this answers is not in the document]
DirkDevelopmentFreelance: Now that I think of it, it may be better to not do this, as otherwise we can use this pattern and leave in a lot of unused variables

api/src/db/db.service.spec.ts
```diff
@@ -244,7 +244,7 @@ describe("DbService", () => {
             // if we got past this point without an exception, the test was successful
             res = true;
             expect(res).toBe(true);
-        }, 10000);
+        }, 20000);
```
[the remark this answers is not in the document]
DirkDevelopmentFreelance: Reverted

api/src/socketio.spec.ts
DirkDevelopmentFreelance: Removed

shared/src/socket/socketio.ts
DirkDevelopmentFreelance: image
Not sure where this should go, as don't we need the local changes to push to api when offline and going online again?
image

ivanslabbert (commented):
We can improve the logic as follows:

- [x] Use a DexieLiveQuery to monitor the localChanges table (instead of polling it at regular intervals)
- [x] It should not be needed to have a retry timer incase the API did not respond with an ACK message (as it was done with Socket.io) as the REST API will respond with either a success or failed ACK. In case the connection as broken, the isConnected monitoring should terminate the pushing process, and start afresh when connected again.

ivanslabbert (commented):
Use this architecture for uploading any objects that contains binary or non JSON stringify-able data:
image

shared/src/util/LFormData/LFormData.ts
```diff
@@ -0,0 +1,108 @@
...
+                        const valueKey = `${fileKey}-${k}`;
+                        if (
+                            typeof v === "string" ||
+                            typeof v === "boolean" ||
+                            typeof v === "number"
+                        ) {
+                            super.append(valueKey, String(v));
+                        } else if (k === "fileData" && this.isBinary(v)) {
+                            const blob = new Blob([v as BlobPart], {
+                                type: "image/octet-stream",
```
DirkDevelopmentFreelance, line 93:
```suggestion
                                type: "image/octet-stream",
```
```suggestion
                                type: "application/octet-stream",
```

api/src/endpoints/changeRequest.controller.ts
```diff
@@ -1,20 +1,73 @@
...
+            const apiVersion = body["changeRequestApiVersion"];
+            await validateApiVersion(apiVersion);
+
+            const changeReqId = JSON.parse(body["changeRequestId"]);
+            const parsedDoc = JSON.parse(body["changeRequestDoc-JSON"]);
+
+            //Only parent documents (Posts and Tags) can have files uploaded,
+            //Child documents only have a reference to the parent document's fileCollection field
+            //without this check it could lead to unexpected behavior or critical errors
+            if ((parsedDoc && parsedDoc.type == DocType.Post) || parsedDoc.type == DocType.Tag) {
```
ivanslabbert, line 41: The DTO checking should do this, but please test and make sure it is either stripping off unsupported fields or rejecting the DTO.

shared/src/rest/syncLocalChanges.spec.ts
```diff
@@ -0,0 +1,268 @@
...
+                expect.arrayContaining([
+                    ["changeRequestId", "1234"],
+                    [
+                        "changeRequestDoc-JSON",
+                        JSON.stringify({ _id: "test-doc", type: "post", updatedTimeUtc: 1234 }),
+                    ],
+                ]),
+            );
+        });
+    });
```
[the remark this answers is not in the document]
DirkDevelopmentFreelance: Fixed it

shared/src/rest/syncLocalChanges.spec.ts
```diff
@@ -0,0 +1,269 @@
+import "fake-indexeddb/auto";
+import { afterAll, afterEach, beforeAll, beforeEach, describe, expect, it, vi } from "vitest";
+import { ref } from "vue";
+import { AckStatus, ChangeReqDto, DocType, LocalChangeDto } from "../types";
+import { db, initDatabase } from "../db/database";
+import { getSocket, isConnected } from "../socket/socketio";
```
ivanslabbert, line 6: Remove unused imports

shared/src/rest/syncLocalChanges.ts
```diff
@@ -0,0 +1,70 @@
...
+        },
+        { immediate: true },
+    );
+
+    watch(isConnected, (connected) => {
+        if (connected) {
+            processChangeReqLock.value = false;
+        } else {
+            processChangeReqLock.value = true;
+        }
```
ivanslabbert, line 68: The else clause might be superfluous

ivanslabbert (approved):
See comment before merging.
Please update the PR title to match the functionality change

shared/src/rest/syncLocalChanges.ts
DirkDevelopmentFreelance: The else ensures that on disconnect, the lock is always set to true, preventing any syncing until you are reconnected and the lock is reset to false, this ensures when the user comes back online everything is indeed correctly set, and prevents bugs

PR #876, opened 2025-05-06
Exclude expired content when content only flag is set in API

DirkDevelopmentFreelance:
To ensure expired documents are not synced, we can use the content-only field and only include documents that are not expired. Not sure if I already did this in a previous PR that is in review?

### DIFF
--- a/api/src/db/db.service.ts
+++ b/api/src/db/db.service.ts
@@ -589,6 +589,11 @@ export class DbService extends EventEmitter {
                             { type: { $in: [DocType.Content] } },
                             { memberOf: { $in: groups } },
                             { parentType: docType },
+                            {
+                                $expr: {
+                                    $lt: ["$publishDate", "$expiryDate"],
+                                },
+                            },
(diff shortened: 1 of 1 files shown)


DirkDevelopmentFreelance (commented):
A test is failing

api/src/db/designDocs/parentType-status-expiryDate-language-index.json
ivanslabbert, line 1: Very good work, but unfortunately CouchDB can in many cases not use indexes if there are $or clauses in the query. In some cases it can make some use of it (if all the branches of the $or clause queries data in the exact same format). So for now we can remove this index, and as a separate task go through all the queries to rewrite them to be indexing friendly.

DirkDevelopmentFreelance: Seems without this file the data is not being sent at all on non-cms(contentOnly) clients

PR #889, opened 2025-05-15
Add lastLoggedIn field to UserDto

DirkDevelopmentFreelance:


api/src/dto/UserDto.ts
```diff
@@ -23,4 +23,9 @@ export class UserDto extends _contentBaseDto {
     @IsString()
     @Expose()
     userId?: string;
+
+    @IsOptional()
+    @IsNumber()
+    @Expose()
+    lastLoggedIn?: string;
```
ivanslabbert, line 30:
```suggestion
    lastLogin?: string;
```

shared/src/types/dto.ts
```diff
@@ -77,6 +77,7 @@ export type UserDto = ContentBaseDto & {
     type: DocType.User;
     email: string;
     name: string;
+    lastLoggedIn?: number;
```
ivanslabbert, line 80:
```suggestion
    lastLogin?: number;
```

cms/src/components/users/UserTable.vue
ivanslabbert, line 1: Remove last updated time, as this will not have a practical function (being continuously updated when the API updates the user docs with the last login times)

ivanslabbert (commented):
Please also implement in API (processJwt.js)

api/src/jwt/processJwt.ts
```diff
@@ -123,12 +123,13 @@ export async function processJwt(
...
-            const updated = { ...d, userId, email };
+            const updated = { ...d, userId, email, lastLogin: Date.now() };
             if (name) updated.name = name;
             if (
                 updated.name !== d.name ||
                 updated.userId !== d.userId ||
-                updated.email !== d.email
+                updated.email !== d.email ||
+                updated.lastLogin !== d.lastLogin
             ) {
```
ivanslabbert, line 133: This `if` statement will now always be true, so it can be removed

cms/src/components/users/UserTable.vue
```diff
@@ -84,12 +84,12 @@ onBeforeUnmount(() => {
...
                             ></th>
 
-                            <!-- updated -->
+                            <!-- Last Logged In -->
                             <th
                                 class="group py-3.5 pl-4 pr-3 text-left text-sm font-semibold text-zinc-900 sm:pl-3"
                                 @click="false"
                             >
-                                <div class="flex items-center gap-2">Last updated</div>
+                                <div class="flex items-center gap-2">Last Logged In</div>
```
ivanslabbert, line 92:
```suggestion
                                <div class="flex items-center gap-2">Last logged in</div>
```

ivanslabbert (approved):
Approved with comments

DirkDevelopmentFreelance:
This is not working anymore after I did the change where I removed the "if" statement, or maybe I implemented it in the wrong area.

ivanslabbert:
write a test to ensure the last logged in time is updated on the user document

DirkDevelopmentFreelance:
> write a test to ensure the last logged in time is updated on the user document

Can't see how a test can be written for this, as this logic depends on what comes from Auth0, because for example, why it didn't work, was because my stuff was not setup correctly on auth0, so the code works fine, so if there is a way we can write a test with auth0 then perhaps that could work. This could be an e2e test maybe?

PR #895, opened 2025-05-20
856 cms implement apilivesync for group editor

belljohan3:
1. List of bugs to fix: 

- [ ] Create button is working only of first click
- [ ] When adding a new ACL list to a group, it is empty and doesn't make the group dirty
- [ ] The duplicate button of the group is working but it is not reactive
- [ ] GroupTable is not reactive

2. Others

- [ ] Fix tests in EditGroup


cms/src/components/groups/GroupTable.vue
ivanslabbert, line 1: Check if still used, and remove if not

ivanslabbert (approved):

PR #903, opened 2025-05-30
899 app videojs audio mode

johan-bell:


app/src/components/content/VideoPlayer.vue
```diff
@@ -75,6 +76,74 @@ function setAudioTrackLanguage(languageCode: string | null) {
...
+                const absoluteTrackUri = new URL(track.uri, manifestDir).toString();
+
+                // Add an EXT-X-MEDIA tag for the audio track to the playlist.
+                lines.push(
+                    `#EXT-X-MEDIA:TYPE=AUDIO,GROUP-ID="${group}",NAME="${name}",LANGUAGE="${track.language}",DEFAULT=${track.default ? "YES" : "NO"},AUTOSELECT=${track.autoselect ? "YES" : "NO"},URI="${absoluteTrackUri}"`,
+                );
+
+                // Add an EXT-X-STREAM-INF tag for the audio track to the playlist.
+                lines.push(
+                    `#EXT-X-STREAM-INF:BANDWIDTH=128000,CODECS="mp4a.40.2",AUDIO="${group}"`,
```
ivanslabbert, line 134: I do not think we should hard-code the bandwidth and codec, as the stream might have different settings than what we are hard-coding. If it cannot be detected from the source M3U8 playlist, we should rather exclude the Bandwidth and Codec setting.

ivanslabbert (approved):
Approved with comments

belljohan3:
# 🎯 Should the `BANDWIDTH` in `#EXT-X-STREAM-INF` Match the Original?

## 🔸 For Video Streams

Yes — it should be as close as possible to the **actual stream bitrate**. HLS players use `BANDWIDTH` to:

- Choose the initial stream
- Adapt to network conditions
- Avoid unnecessary rebuffering

💡 If the value is too **low**, the player might select it too early.  
💡 If it’s too **high**, the player might avoid it even if it could play smoothly.

---

## 🔸 For Audio-Only Streams (like you're generating)

No — it doesn't have to be **exact**, but it **should be realistic**.

### Why?
- Original manifests typically **don’t provide per-audio-track** `#EXT-X-STREAM-INF`, so there's no reference.
- Most HLS players **only require a sensible value**, not an exact match.

### ✅ Typical Audio Bitrates

| Codec       | Channels | Bitrate (approx) |
|-------------|----------|------------------|
| AAC-LC (mono)   | 1        | 64,000 – 96,000 bps  |
| AAC-LC (stereo) | 2        | 128,000 – 192,000 bps |

---

## ✅ Best Practice for You

Since you're dynamically generating the playlist:

- Use `BANDWIDTH` values like:
  - **Mono**: `96000`
  - **Stereo**: `128000` or `160000`
- You can also **measure segment size** from the `.m3u8` chunk playlist to estimate bitrate more precisely (optional).

---

## ✅ Conclusion

> Use **realistic, approximate values** for audio-only playlists.  
> Exact match **isn’t required**, unless you're replacing an existing master playlist and need tight control.

belljohan3:
Hey @ivanslabbert @DirkDevelopmentFreelance 

I used base64-encode here because Videojs doesn't support HLS via blob because it is also videojs expect a direct HTTP(S) URL that the player or native decoder can request as a standalone resource.

app/src/components/content/VideoPlayer.vue
```diff
@@ -75,6 +76,98 @@ function setAudioTrackLanguage(languageCode: string | null) {
...
+                const relativeTrackUri = track.uri.split("?")[0];
+
+                // Find the matching playlist for this audio group
+                const matched = playlists.find(
+                    (p) => p.attributes?.AUDIO === group && p.uri.includes(relativeTrackUri),
+                );
+
+                // Infer bandwidth based on group name
+                const isStereo = group.toLowerCase().includes("stereo");
+                const isMono = group.toLowerCase().includes("mono");
```
ivanslabbert, line 146: Use the channels track attribute instead of parsing the group names. If the channels attribute is not set, assume the track is stereo

app/src/components/content/VideoPlayer.vue
```diff
@@ -75,6 +76,143 @@ function setAudioTrackLanguage(languageCode: string | null) {
     }
 }
 
+/**
+ * Extracts and builds an audio master playlist from a given HLS manifest URL.
+ *
+ * @param {string} originalUrl - The URL of the original HLS manifest file.
+ * @returns {Promise<string>} - A Promise that resolves to the generated audio master playlist as a string.
+ */
+async function extractAndBuildAudioMaster(originalUrl: string): Promise<string> {
```
ivanslabbert, line 85: Extract this function into a separate file. The file can be in the same folder as VideoPlayer.vue

Issue #856, opened 2025-04-24
CMS: Implement ApiLiveSync for group editor


ivanslabbert:
Refactor group editor to make use of the ApiLiveSync class instead of polling the API for updates

DirkDevelopmentFreelance:
I just need help with the EditGroup.spec test, as it expects the query prop, and no matter what I put there it gives me type errors

PR #906, opened 2025-06-03
779 cms api add search filter on useroverview

johan-bell:


cms/src/components/content/query.ts
```diff
@@ -25,6 +29,75 @@ export type ContentOverviewQueryOptions = {
...
+        result.sort((a, b) => {
+            const valA = a[options.orderBy!];
+            const valB = b[options.orderBy!];
+            if (valA === valB) return 0;
+            if (options.orderDirection === "desc") return valA < valB ? 1 : -1;
+            return valA > valB ? 1 : -1;
+        });
+
+        return result;
+    });
```
ivanslabbert, line 83: We should rather implement this in the API, especially for use cases where there will be many users.

Putting this on hold until the updated ApiLiveQuery is complete (#856)

PR #923, opened 2025-06-09
912 app add continue reading query to homepage

### LINKED ISSUE #912: APP: Add "Continue reading" query to homepage
Continue reading should be calculated using some logic to determine if the user is actually reading, and to which line he/she has read. Could work on a "exposure time" per word concept, or scroll position and time spent, etc.

See Google AI search result for some inspiration (search string: "analytics determine how far a user has read an article")

Image

johan-bell:


ivanslabbert:
The read tracking should more accurately detect if a user is actually reading, or if a user is just quickly scrolling through an article to prevent us from storing "continue reading" articles that the user has not actually ready, resulting in unreliable data.

PR #927, opened 2025-06-10
Revert tailwind to v3

johan-bell:


ivanslabbert:
We should rather manually revert the changes.

PR #944, opened 2025-07-01
Add gzip compression settings to nginx configuration in app and cms

DirkDevelopmentFreelance:


cms/nginx.conf
```diff
@@ -6,4 +6,29 @@ server {
         index  index.html index.htm;
         try_files $uri $uri/ /index.html;
     }
+
+    # Enable gzip compression
+    gzip on;
```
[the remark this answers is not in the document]
DirkDevelopmentFreelance: For our use case this might be a bit extreme, our nginx.config is sufficient.

app/nginx.conf
```diff
@@ -27,4 +27,28 @@ server {
...
+        application/xml
+        application/xml+rss
+        text/javascript
+        image/svg+xml
+        image/bmp
+        image/tiff
+        image/webp
+        image/gif
+        image/png
+        image/jpeg
```
ivanslabbert, line 48: Remove binary file formats as it gives very little compression compared to the time it takes to compress. It will probably make the responses slower for most users.

cms/nginx.conf
```diff
@@ -6,4 +6,28 @@ server {
...
+        application/xml
+        application/xml+rss
+        text/javascript
+        image/svg+xml
+        image/bmp
+        image/tiff
+        image/webp
+        image/gif
+        image/png
+        image/jpeg
```
ivanslabbert, line 27: Remove binary file formats as it gives very little compression compared to the time it takes to compress. It will probably make the responses slower for most users.

ivanslabbert (approved):
Approved with two comments

app/README.md
```diff
@@ -65,4 +65,10 @@ docker build -t luminary-app .
 docker run --rm -it -p 8080:80 luminary-app
 ```
 
+`gzip` functionality is enabled by default, disable it as shown:
```
ivanslabbert, line 68: Add a note that this is available as a docker .env parameter

ivanslabbert (approved):
Approved, but implement for CMS as well

PR #945, opened 2025-07-01
Fix image re-processing in the CMS after save

MrDirkelz:


cms/src/components/content/EditContent.vue
```diff
@@ -75,10 +78,21 @@ const editableParent = ref<ContentParentDto>({
 });
 const isLoading = computed(() => editableParent.value == undefined);
 const existingParent = ref<ContentParentDto>(); // Previous version of the parent document for dirty check
+const liveParent = useDexieLiveQuery(() => db.docs.where("_id").equals(parentId).first(), {
```
ivanslabbert, line 81: Should be able to use a "get" which might be more efficient

ivanslabbert (approved):

PR #947, opened 2025-07-01
CMS: Route to default language on ContentCard title click

MrDirkelz:


ivanslabbert (approved):
Approved, but please add a test for this?

PR #948, opened 2025-07-01
933 cms make edit content mobile friendly

belljohan3:


belljohan3:
- [ ] Fix all the margins and gap to match the same
- [ ] Add one more EditContentText in side and hide it on dsktop to fix the scrolling
- [ ] add the back button on desktop
- [ ] go through everything
- [ ] fix margin in SEO's inputs

cms/src/components/content/EditContentVideo.spec.ts
```diff
@@ -20,19 +20,6 @@ describe("EditContentVideo.vue", () => {
...
-            props: {
-                disabled: false,
-                content: content.value,
-            },
-        });
-
-        const videoContent = wrapper.find('div[data-test="videoContent"]');
-        expect(videoContent.exists()).toBe(false);
-    });
-
```
DirkDevelopmentFreelance, line 35: Is there another test that was created for this removed test? Think we don't want to remove tests if the functionality has not changed

cms/src/components/content/EditContentVideo.spec.ts
```diff
@@ -20,19 +20,6 @@ describe("EditContentVideo.vue", () => {
...
     it("display video field, when is defined", async () => {
```
DirkDevelopmentFreelance, line 23: This was probably not added in this ticket, but just something that I noticed here
```suggestion
    it("displays the video field, when it is defined", async () => {
```

cms/src/components/content/EditContentVideo.spec.ts
belljohan3: The funcionnality changed. We keep collapsed when a video is defined and non collapsed when not defined

DirkDevelopmentFreelance:
It seems there are some conflicts that need to be fixed.

belljohan3:
- [ ] Remove shadow on the cards
- [ ] add shadow bottom on the topbar
- [ ] fix the margin of images
- [ ] make the EditContentText fill the rest of the space and make the input text scrollable

DirkDevelopmentFreelance:
> @belljohan3 @ivanslabbert. From trying to work on this myself, I realised our structure for edit content makes it hard to make the content card stick below the translations card. I do have a solution that we can implement for it, but it is a bit complicated, but will make our edit content page, or even just the overall app easier to maintain, we can use a intersection observer to check if the EditText component is overlapping with the language selector on mobile and just "lock" the position of the EditText component there. But implementing a component that is able to observe "collision" between components makes it easier to place elements where we want them via javascript. I can do this in a seperate ticket as I have worked with intersecting elements before via hobby game development https://vueuse.org/core/useIntersectionObserver/#directive-usage

I could use this for making the language selector collapse when reaching the top

DirkDevelopmentFreelance:
- [ ] Make text editor fill rest of screen when scrolling past text editor

PR #949, opened 2025-07-02
Implement a network changes notification to editContent.vue

MrDirkelz:


DirkDevelopmentFreelance:
There is a bug that  I am not sure how to fix, will look for ideas from others during review

PR #951, opened 2025-07-03
Add Playwright testing framework and initial test cases for the CMS

MrDirkelz:


DirkDevelopmentFreelance:
Status:

1. Local tests have been set up and are working
2. Specific Auth0 instance has been set up for end to end tests

Todo:
- [ ] Set up github env variables set specifically for end to end tests.
- [ ] Ensure timings are correct for github actions and how slow it is.
- [ ] Ensure API is running correctly, alongside user access has to be set to super-admin for the e2e test user

DirkDevelopmentFreelance:
Hi @ivanslabbert, like last time I got the end to end tests to work locally, and I am trying to get the app to also use the new auth0 end-to-end environment that was created. But I am struggling to get everything working in github actions. I have also tried running it in codespaces but I can't figure out how to get it to work in github actions. I was wondering if you could maybe help me on that part? As it is mostly not related to the tests

I also tried looking at how someone else did it, but I just can't seem to crack the github actions side

Here is a reference to somebody else's PR for the end to end testing, but I don't see that they are using it like we want to use it
https://github.com/onetimesecret/onetimesecret/pull/710/files#diff-c5d401e588c34e31052f1b0c43179b5bc943d595859afbe9f21ecc35fa84620a

johan-bell:
@DirkDevelopmentFreelance We can pair-program on it if it is needed.

PR #960, opened 2025-07-07
952 app continue section improvements

belljohan3:


app/src/globalConfig.ts
```diff
@@ -237,11 +237,11 @@ export const setMediaProgress = (
         _mediaProgress.splice(index, 1);
     }
 
-    _mediaProgress.push({ mediaId, contentId, progress, duration });
+    _mediaProgress.unshift({ mediaId, contentId, progress, duration });
 
     // Only keep the last 10 progress entries
     while (_mediaProgress.length > 10) {
-        _mediaProgress.shift();
+        _mediaProgress.length = 10;
```
ivanslabbert, line 244: If setting the length like this, rather use an if instead of a while.
A better approach could be to use splice to remove all entries with index >= 10

ivanslabbert (approved):
Approved with 1 comment

PR #963, opened 2025-07-08
Removed triangle with exclamation and replaced with proper loading icon

Rapotukganakga:


cms/src/components/LoadingBar.vue
```diff
@@ -0,0 +1,20 @@
+<template>
+    <div class="relative h-3 w-1/2 overflow-hidden rounded-full border border-[#736d6d]">
+        <div class="animate-loading-bar absolute left-0 top-0 h-full bg-[#e2d003]"></div>
```
ivanslabbert, line 3: Check if we can use standard colours from the Tailwind colour palette?

The border colour should be a tailwind zink colour, probably border-zinc-300 / border-zinc-400 or so.

cms/src/components/users/UserTable.vue
```diff
@@ -106,7 +106,7 @@ onBeforeUnmount(() => {
                     </tbody>
                 </table>
                 <div class="flex h-32 w-full items-center justify-center gap-2" v-if="isLoading">
-                    <ExclamationTriangleIcon class="h-6 w-6 text-zinc-500" />
+                    <ArrowPathIcon class="h-6 w-6 text-zinc-500" />
```
ivanslabbert, line 109: Should we rather use the loading bar here as well?

ivanslabbert (commented):
We should remove the loadingspinner (make sure it is not in use somewhere else or replace it with the loadingBar

PR #966, opened 2025-07-09
Fix LastLogin field tracking

DirkDevelopmentFreelance:


api/src/jwt/processJwt.ts
```diff
@@ -120,10 +121,13 @@ export async function processJwt(
 
     const userDocs = (await db.getUserByIdOrEmail(email, userId)).docs as UserDto[];
 
-    // Update user details in the database (if changed) if userId is set
-    if (userId) {
+    // Update user details in the database if either userId or email is set
+    if (userId || email) {
```
ivanslabbert, line 125: This gives a potential security risk where the userid can be updated by someone signing in with only email.

Revert the `if` and add an
```ts
else if (email) {
   // update only timestamp
}
```

ivanslabbert (approved):

DirkDevelopmentFreelance:
> Please add a test before merging?

The spec file seems to already have the same tests  I would write for the functionality, the functionality has just been extended a bit to be more accurate, so the existing tests are still relevant to the changes made.

PR #978, opened 2025-07-23
Fix Audio Downloading in the Background when navigating away from a Live video

DirkDevelopmentFreelance:


app/src/components/content/VideoPlayer.vue
```diff
@@ -279,9 +279,24 @@ onMounted(() => {
...
-    player?.off("play", playerPlayEventHandler);
-    player?.off(["useractive", "userinactive"], playerUserActiveEventHandler);
+
+    if (player) {
+        // Pause the player first
+        player.pause();
+
+        // Reset the source to stop any ongoing downloads
+        // This prevents live videos from continuing to download audio in the background
+        player.src({ type: "application/x-mpegURL", src: "" });
```
ivanslabbert, line 289: Test if this is needed. Calling `player.dispose` should probably be sufficient.

ivanslabbert (approved):

app/src/components/content/VideoPlayer.vue
DirkDevelopmentFreelance: It seems it is

Issue #988, opened 2025-08-06
APP: Translation banner showing after selecting language with the quick language selector


ivanslabbert:
Translation banner showing after selecting language with the quick language selector. This should not be happening, as it was by user choice.

Pressing on the banner does nothing (same issue as with privacy policy banner?)

![Image](https://github.com/user-attachments/assets/3a16b3f6-1ccd-4b83-a93c-7657391ddbf2)

PR #990, opened 2025-08-06
APP: Fix notification banner bugs - Fix for #988

DirkDevelopmentFreelance:
Just used a function to mark a language switch from the dropdown correctly
Also added a global composable for the PrivacyPolicyModal to ensure the show model is not overwriting each other, so that there can only be one PrivacyPolicyModal instance at a time


DirkDevelopmentFreelance:
> The issue is probably due to the change we made moving away from a fixed top-bar to a base page where the top bar is included in the base page.
> 
> The current fix feels like a workaround instead of adressing the root of the problem

The top bar being in the base page was the problem, as the privacy policy modal is created *inside* of the profile menu, hence each page now gets its own "unique" profile menu/privacy policy modal that only exists *inside* the profile menu. So when trying to open it with the banner, it is trying to open the privacy policy modal inside of the profile menu where it was first mounted. 

I therefore stand by the approach of moving the privacy policy modal to the `App.vue` component, as this makes it global, and there can then only be one instance of this modal and it avoids conflicts.

ivanslabbert (approved):

PR #993, opened 2025-08-06
Implement privacy policy acceptance flow in authentication process

MrDirkelz:


DirkDevelopmentFreelance:
We decided that we need to think more in depth about the user experience in the stand up meeting on Aug 7 2025

belljohan3:
1. Not showing Necessary only button if the user is logged in
2. add i18n for string

belljohan3:
> 1. Not showing Necessary only button if the user is logged in
> 2. add i18n for string

This is done✅ in #[ffdf6db](https://github.com/bccsa/luminary/pull/993/commits/ffdf6dbfb87aaaea4684605c68b2c151a12554f1)

DirkDevelopmentFreelance:
visually looks good but it is not popping the notification that says that the privacy policy changed after having accepted, being logged in, and it changed. The privacy policy also takes a long while to actually display

DirkDevelopmentFreelance (commented):
visually looks good but it is not popping the notification that says that the privacy policy changed after having accepted, being logged in, and it changed. The privacy policy also takes a long while to actually display

belljohan3:
Thanks for the input @DirkDevelopmentFreelance i still have to fix that. Please move it to change requested if possible

THISmann (commented):
in the general the code is clear but for security issues , i suggest to implement more sanitize for input , to prevent again XSS attack

PR #999, opened 2025-08-08
Fix sorting logic in contentByTag to order by newest publish date

MrDirkelz:


app/src/components/contentByTag.ts
```diff
@@ -35,20 +35,24 @@ export const contentByTag = (
...
                     Object.assign(result.tagged.value[index], {
-                        newestContentDate: sorted[sorted.length - 1].publishDate || 0,
+                        newestContentDate: sorted[0].publishDate || 0,
                         content: sorted,
                     });
                 } else {
                     result.tagged.value.push({
                         tag,
-                        newestContentDate: sorted[sorted.length - 1].publishDate || 0,
+                        newestContentDate: sorted[0].publishDate || 0,
```
ivanslabbert, line 55: We need to also look at the pinned condition here to get the newest date in the collection, which is used for sorting the categories

PR #1006, opened 2025-08-15
Refactor router to use dynamic imports

DirkDevelopmentFreelance:
This improves initial load time by only loading components when they are needed.

Also adds v-memo to content tiles to prevent unnecessary re-renders.

### DIFF
--- a/app/src/router/index.ts
+++ b/app/src/router/index.ts
@@ -1,11 +1,4 @@
 import { createRouter, createWebHistory } from "vue-router";
-import NotFoundPage from "@/pages/NotFoundPage.vue";
-import HomePage from "@/pages/HomePage.vue";
-import SettingsPage from "@/pages/SettingsPage.vue";
-import SingleContent from "@/pages/SingleContent.vue";
-import ExplorePage from "@/pages/ExplorePage.vue";
-import BookmarksPage from "@/pages/BookmarksPage.vue";
-import VideoPage from "@/pages/VideoPage.vue";
 
 const router = createRouter({
     history: createWebHistory(import.meta.env.BASE_URL),
@@ -19,7 +12,7 @@ const router = createRouter({
     routes: [
         {
             path: "/",
-            component: HomePage,
+            component: () => import("@/pages/HomePage.vue"),
             name: "home",
             meta: {
                 title: "title.home",
@@ -28,7 +21,7 @@ const router = createRouter({
         },
         {
             path: "/explore",
-            component: ExplorePage,
+            component: () => import("@/pages/ExplorePage.vue"),
             name: "explore",
             meta: {
                 title: "title.explore",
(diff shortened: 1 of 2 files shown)


app/package.json
```diff
@@ -40,31 +41,34 @@
...
         "doiuse": "^6.0.5",
         "eslint": "^8.49.0",
         "eslint-plugin-vue": "^9.17.0",
         "fake-indexeddb": "^5.0.2",
         "jsdom": "^23.0.1",
         "npm-run-all2": "^6.1.1",
+        "postcss": "^8.5.4",
         "prettier": "^3.0.3",
         "prettier-plugin-tailwindcss": "^0.5.11",
+        "rollup-plugin-visualizer": "^6.0.3",
```
ivanslabbert, line 71: Add documentation on how to use the visualizer, or remove from project

app/src/main.ts
```diff
@@ -21,12 +20,32 @@ if (import.meta.env.VITE_FAV_ICON) {
...
+            const Sentry = await import("@sentry/vue");
+            if (message) {
+                Sentry.captureMessage(message);
+            }
+            Sentry.captureException(error);
+        } catch (e) {
+            console.error("Failed to load Sentry:", e);
+        }
+    }
+}
```
ivanslabbert, line 35: Only import once, and keep a reference to the imported library
```suggestion
function reportError(error: any, message?: string) {
    if (import.meta.env.PROD) {
        if (!Sentry) {
               console.log("Sentry not loaded yet");
               return;
            }
            if (message) {
                Sentry.captureMessage(message);
            }
            Sentry.captureException(error);
    }
}
```

app/src/main.ts
```diff
@@ -52,13 +71,13 @@ async function Startup() {
         ],
     }).catch((err) => {
         console.error(err);
-        Sentry.captureException(err);
+        reportError(err);
```
ivanslabbert, line 74: If we want to not use the `reportError` function, this could be:
```suggestion
        Sentry?.captureException(err);
```

app/src/router/index.ts
```diff
@@ -19,7 +12,7 @@ const router = createRouter({
     routes: [
         {
             path: "/",
-            component: HomePage,
+            component: () => import("@/pages/HomePage.vue"),
```
ivanslabbert, line 15: Lazy loading the homepage might not add any benefit

app/src/router/index.ts
ivanslabbert, line 1: After having loaded the homepage, we ideally want to load all the rest of the pages in the background, so that when the user navigates, the other pages' chunks are already downloaded, giving a fast user experience without slowing down the initial loading time of the homepage.

DirkDevelopmentFreelance:
> Check if lodash can be tree-shakeable in the shared library, as we still see a big lodash chunk when using the visualizer.
Lodash  was removed from the shared library, I explained it wrong last time, basicly lodash now only runs through the app, and goes down to the shared library to make the shared library smaller in turn and only install it at the "top level project". So I will try to force it to tree shake unnecessary imports in the *app*
image

ivanslabbert (approved):

DirkDevelopmentFreelance:
We may be able to remove RxJs here as well, just need to check, but it can be done in this PR or another new one

PR #1022, opened 2025-08-22
Add close button to LCombobox modal popup

DirkDevelopmentFreelance:


cms/src/components/common/LDialog.vue
ivanslabbert, line 1: Can we rather extend LModal in LDialog to keep the look and feel the same, and prevent code duplication?

ivanslabbert (approved):

PR #1023, opened 2025-08-22
CMS: Fix text editor bugs

MrDirkelz:


cms/src/components/editor/RichTextEditor.vue
```diff
@@ -56,16 +56,17 @@ const editor = useEditor({
...
-                .replace(/&nbsp;/gi, " ") // Clean non breaking spaces
-                // Clean heading tags
-                .replace(/<h([1-5])([^>]*)>/gi, (match, level, attrs) => {
+                .replace(/&nbsp;/g, " ") // Clean non breaking spaces
+                .replace(/<h([1-6])([^>]*)>/gi, (match, level, attrs) => {
                     const newLevel = Math.min(parseInt(level) + 1, 6);
                     return `<h${newLevel}${attrs}>`;
                 })
-                .replace(/<\/h([1-5])>/gi, (match, level) => {
+                .replace(/<\/h([1-6])>/gi, (match, level) => {
```
ivanslabbert, line 69: We do not support H6. All headings above H5 should be converted to normal text, i.e. stripping out the H(x) tags.

cms/src/components/editor/RichTextEditor.vue
```diff
@@ -330,8 +331,20 @@ div[data-tiptap-editor] > div {
...
     box-sizing: border-box;
     flex: 1;
     overflow-y: auto;
 }
+
+.ProseMirror > :first-child {
+    margin-top: 0;
+}
+
+.ProseMirror h5 {
```
ivanslabbert, line 345: Add a comment explaining why custom CSS is used

ivanslabbert (commented):
Add conditional logic to only bump up the heading level if an H1 is present in the pasted text.

cms/src/util/formatPastedHtml.ts
ivanslabbert, line 1: Add a test for this function, and add a test for RichTextEditor to check that formatPastedHtml.ts is called.

ivanslabbert (approved):

PR #1036, opened 2025-08-27
Move close button to top right corner and add zoom icons buttons

belljohan3:


belljohan3:
Could not make it to work with the [mask](https://tailwindcss.com/docs/mask-image) from tailwind. It seems like it works with image. I will merge it like this for now. @ivanslabbert

PR #1073, opened 2025-09-05
APP: Remove you are not logged in message / no content on Homepage

MrDirkelz:
We now have public content, which should display even when not logged in. So all the logic here can be removed


belljohan3 (dismissed):
Can remove the tests as well as it is not valid anymore ?

ivanslabbert (approved):

PR #1088, opened 2025-09-09
Implement useDexieLiveQueryAsEditable for editable live queries with Dexie

ivanslabbert:


DirkDevelopmentFreelance (commented):
Just one comment of something that could cause a bug in the change requests

PR #1107, opened 2025-09-15
APP: Fix empty image bug

DirkDevelopmentFreelance:


app/src/components/images/LImageProvider.vue
```diff
@@ -312,7 +312,7 @@ const modalSrcset = computed(() => {
         ]"
         alt=""
         data-test="image-element2"
-        loading="lazy"
+        loading="eager"
```
belljohan3, line 315: Is there a specific reason why you changed this ?

belljohan3 (changes requested):
It looks nice for me. Just make to fix the conflicts we have in `package` and `package-lock` files

PR #1109, opened 2025-09-15
APP: Add PWA support 

### LINKED ISSUE #1108: APP: Add PWA support with vite plugin
A small prompt gave me this reply on whether or not it is better to pair capacitor and the vite plugin

ChatGPT said:

Yes — using vite-plugin-pwa can simplify things quite a bit when you want to pair it with Capacitor.

Here’s why:

1. Capacitor by itself

Capacitor just wraps your web app into a native container.

It doesn’t handle offline caching, asset updates, or service workers.

If you want PWA-like behavior (offline support, background updates, etc.), you’d normally have to set up your own service worker and cache strategies.

2. With vite-plugin-pwa

It generates and injects a service worker during your Vite build.

Handles things like:

asset caching (via Workbox)

update strategies (autoUpdate / prompt)

manifest injection (so your PWA metadata is consistent).

You don’t have to manually write/maintain a service worker unless you want custom logic.

3. Together with Capacitor

Capacitor doesn’t care whether your web app is a plain SPA or a PWA.

If your app already has the vite-plugin-pwa service worker bundled:

On the web → it behaves like a PWA (offline, caching, auto-updates).

Inside Capacitor → it still works, but usually you’ll disable some parts (like the install prompt, since you’re running inside a native shell).

This gives you one unified build pipeline → Vite builds → plugin generates service worker → Capacitor wraps it → deploy to stores and web.

⚡️ So yes:
Using vite-plugin-pwa makes your Capacitor build simpler because you don’t have to hand-roll service worker code or manage updates. You just let the plugin handle the PWA side, and Capacitor takes care of the native wrapper.

MrDirkelz:


app/src/components/images/LImageProvider.spec.ts
```diff
@@ -45,7 +45,7 @@ describe("LImageProvider", () => {
         expect(img2.attributes("srcset")).toContain("classic-100.webp");
     });
 
-    it("renders fallback image when no main images are available", async () => {
+    it.skip("renders fallback image when no main images are available", async () => {
```
johan-bell: Consider make a ticket for skipped test ?

app/vite.config.ts
```diff
@@ -13,6 +13,84 @@ const env = loadEnv("", process.cwd());
...
+                        },
+                    },
+                    // Same-origin API (GET only)
+                    {
+                        urlPattern: ({ url, request }) =>
+                            request.method === "GET" && url.pathname.startsWith("/api/"),
+                        handler: "NetworkFirst",
+                        options: {
+                            cacheName: "api",
+                            expiration: { maxEntries: 200, maxAgeSeconds: 60 * 60 * 24 * 7 },
```
johan-bell: Add a comment here to specify how much time the `maxAgeSeconds` is ?

DirkDevelopmentFreelance:
> @DirkDevelopmentFreelance looks like this is has been merged ? Could you fix the confilcts ?

This is still in progress. Not ready for review.

PR #1110, opened 2025-09-15
audio player experience

johan-bell:


cms/src/components/media/MediaEditorThumbnail.vue
```diff
@@ -0,0 +1,282 @@
...
+    display: none;
+}
+audio::-webkit-media-controls-time-remaining-display {
+    display: none;
+}
+
+/* Play button css */
+audio::-webkit-media-controls-play-button {
+    margin-right: 10px;
+}
```
ivanslabbert: remove

ivanslabbert (commented):
Create a separate ticket to create a S3 service that can be used

DirkDevelopmentFreelance (changes requested):
I am only able to upload on the cms, the content in the app gets no audio,

Also  "translations" should be "text" now I think. But the cms visuals look nice, just couldn't test the app, as no audios are being displayed on the content

PR #1122, opened 2025-09-16
APP : Add LoadingBar component to the App as replacement for LoadingSpinner

THISmann:
…pinner

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -4,8 +4,5 @@
   "tailwindCSS.experimental.classRegex": [
     ["cva\\(([^)]*)\\)", "[\"'`]([^\"'`]*).*?[\"'`]"],
     ["cx\\(([^)]*)\\)", "(?:'|\"|`)([^']*)(?:'|\"|`)"]
-  ],
-  "[vue]": {
-    "editor.defaultFormatter": "Vue.volar"
-  }
+  ]
 }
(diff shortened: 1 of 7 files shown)


johan-bell (changes requested):
Nice job @THISmann but remember to rebase to main

THISmann (commented):
nothing

MrDirkelz (changes requested):
https://github.com/user-attachments/assets/e09a4855-2ad8-4ea4-9b51-9d68d8186315

This is some really great work @THISmann, it looks very "sleek". 

I just have some comments, but the code and visuals mostly look good.

 Maybe the light mode should be zinc color? As right now in light mode it looks very contrasted compared to the white/zinc background. Also we usually try to avoid gradients and usually just stick with a solid color, as it could impact performance, but this is perhaps a very small component that won't really impact performance that much. But not sure if we want it to look and feel like the cms loading bar as well so it is more clear on the progress of the loading?

But great work! Think it looks nice. We could even look into having "ghost" elements in the future. But that is something to think about later.

MrDirkelz (changes requested):
Some tests are still failing

johan-bell (changes requested):
Nice job, But i highly recommand installing prettier and eslint to avoid useless formating. @THISmann

PR #1123, opened 2025-09-16
Add socket events for ISR-server

MrDirkelz:


api/src/socketio.ts
```diff
@@ -93,8 +97,16 @@ export class Socketio implements OnGatewayInit {
...
+        this.server = server;
         // Handle authentication
         server.use(async (socket, next) => {
+            if (socket.handshake.query.isrHandshakeToken === "random-test-token-for-now") {
+                this.isrSocketId = socket.id;
+                socket.data.userDetails = { groups: [], accessMap: {} } as JwtUserDetails;
+                next();
+                return;
+            }
+
```
DirkDevelopmentFreelance, line 109: This is to still maintain security, @THISmann, we will be adding another socket server that will be retrieving data from these socket events, this data will be used for SSG, is there any security concerns with this approach?

ivanslabbert:
Closing as we will not make changes to the API for the ISR render service - the render service should interact with the API as a normal (Web) client.

PR #1133, opened 2025-09-18
fix: visuals bugs in EditContent

belljohan3:


DirkDevelopmentFreelance (changes requested):
1. Maybe the Tranlsations card should by default show the plus icon and not show the card expander button? As when you click on it all that happens are the colors changes and that may be confusing to users, and we want the plus icon to then also be easily accessable

2. The "show publish date" button still looks the same and emphasised.

cms/src/components/common/LCard.vue
```diff
@@ -83,7 +83,7 @@ function collapse() {
             </div>
         </div>
 
-        <slot name="persistent" />
+        <slot name="persistent" :collapsed="collapsed" />
```
DirkDevelopmentFreelance, line 86: There is a simpler solution where I think you won't have to pass this down, but I think I will do that in a seperate ticket then you can just revert the translation card collapsing functionality

ivanslabbert (approved):

PR #1138, opened 2025-09-22
CMS: Fix sidebar "dropdowns" state changing on route changes

DirkDevelopmentFreelance:


cms/src/globalConfig.ts
```diff
@@ -21,6 +21,8 @@ window.addEventListener("resize", () => {
     windowWidth.value = window.innerWidth;
 });
 export const isSmallScreen = computed(() => windowWidth.value < 1500);
+export const showPostsInSidebar = ref(false);
+export const showTagsInSidebar = ref(false);
```
belljohan3, line 25: I am not sure this should be there as we won't use that in other place. I suggest to move this to Sidebar component itselft ? What do you think @DirkDevelopmentFreelance

DirkDevelopmentFreelance: It can't be in the sidebar component itself, because the sidebar mounts everytime you route. It will then always reset back to false wherever you route

ivanslabbert:
@belljohan3 please approve, rebase and merge?

belljohan3 (approved):

PR #1149, opened 2025-09-29
Add screen reader text for language selection and profile menu items

MrDirkelz:
Enhance accessibility by adding screen reader text for language selection and profile menu items. This also makes it easy for the static site generator to crawl the languages dynamically


johan-bell (approved):

ivanslabbert (changes requested):
We should rather not use funcitonality (like screen reader accessibility) for a different purpose that what it is designed for.

We can however add a special URL query param to force the app language to a given language, which would be a cleaner approach.

PR #1154, opened 2025-10-02
Potential fix for code scanning alert no. 5: Client-side URL redirect

belljohan3:
Potential fix for [https://github.com/bccsa/luminary/security/code-scanning/5](https://github.com/bccsa/luminary/security/code-scanning/5)

The best way to address this problem is to ensure that only trusted, intended URLs can be used as the value for `MATOMO_SERVER`. The most secure approach is to maintain a whitelist of allowed URLs or hostnames, and only proceed to construct the script URL if the extracted value matches one of the allowed entries. In this specific scenario, a simple approach would be to define, in the service worker script, a list or set of allowed Matomo server hostnames or base URLs, only allowing import from one of those. If `matomo_server` does not match an entry (by exact string or hostname comparison), abort or throw an error, rather than blindly using whatever the user provides.

This change would be implemented in app/src/analytics/service-worker.js: extract the value, check if it is included in your set/list, and only then compose the script URL and call `importScripts()`. If the value is not valid, log an error or take other appropriate action.

---

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._

### DIFF
--- a/app/src/analytics/service-worker.js
+++ b/app/src/analytics/service-worker.js
@@ -1,4 +1,42 @@
-const MATOMO_SERVER = new URL(location.href).searchParams.get("matomo_server");
+// Define a whitelist of allowed Matomo server hosts (e.g., by hostname only, or full base URLs)
+const ALLOWED_MATOMO_SERVERS = [
+  "https://matomo.example.com",
+  // Add other allowed servers as needed
+];
+
+const matomoServerUrlRaw = new URL(location.href).searchParams.get("matomo_server");
+let MATOMO_SERVER = null;
+
+try {
+  if (matomoServerUrlRaw) {
+    // Normalize the provided server URL (strip trailing slashes, enforce HTTPS)
+    const url = new URL(matomoServerUrlRaw);
+    // Optionally enforce HTTPS only:
+    if (url.protocol !== "https:") {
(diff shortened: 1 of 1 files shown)


belljohan3:
I wondering if we should set the env for allowed servers. What do you think @ivanslabbert @THISmann @DirkDevelopmentFreelance ?

.github/workflows/app-deploy-staging.yml
```diff
@@ -45,6 +45,7 @@ jobs:
           echo VITE_PRIVACY_POLICY_ID=="${{ vars.VITE_PRIVACY_POLICY_ID }}" >> .env
           echo VITE_ANALYTICS_HOST="${{ secrets.VITE_ANALYTICS_HOST }}" >> .env
           echo VITE_ANALYTICS_SITEID="${{ secrets.VITE_ANALYTICS_SITEID }}" >> .env
+          echo VITE_ALLOWED_MATOMO_SERVERS="${{ secrets.VITE_ALLOWED_MATOMO_SERVERS }}" >> .env
```
johan-bell: May be this should be a vars instead of a secret ?

johan-bell (commented):
Nice accomplishment @DirkDevelopmentFreelance

ivanslabbert (approved):
Approved, but test with a real Matomo instance before merging

PR #1170, opened 2025-10-17
App - add Highlight support

MrDirkelz:


shared/stats.html
ivanslabbert: Delete before merging

PR #1175, opened 2025-10-20
Add automate-luminary script and documentation for project setup and management

MrDirkelz:
This will add a automation script to our project for a better and faster development experience


DirkDevelopmentFreelance:
> Hey @DirkDevelopmentFreelance
> 
> I see a test failing. Could you also look at the copilot review first before we digg into it later ?

I already checked Copilot’s suggestions, but I think we shouldn’t rely on it completely — it might miss things like subtle visual changes (when it is a visual PR) or project-specific details that only we’d catch. Maybe you’ll notice something I overlooked. 😊

ivanslabbert (approved):
Just check copilot comments?

PR #1180, opened 2025-10-21
Add stats.html to .gitignore files across multiple directories

MrDirkelz:


johan-bell (changes requested):
Nice @DirkDevelopmentFreelance 
But as far as i remeber, we only need to add `stats.html` in `.gitignore` of the shared folder. We don't need it in the CMS, APP folder, also only if there is specific reason ?

PR #1190, opened 2025-10-30
SideBar and MobileSideBar moved from the BasePage to the App.vue

ChrisTouo:


ivanslabbert (commented):
I believe with the new approach it should be possible to do the sidebar purely in CSS, without needing javascript functions or variables / Vue props?

PR #1194, opened 2025-10-31
SideBar and MobileSideBar moved from the BasePage to the App.vue

MrDirkelz:
This is a new branch that @ChrisTouo made. Some weird thing happend where all his commits were removed after a rebase.

### DIFF
--- a/cms/src/components/navigation/SideBar.vue
+++ b/cms/src/components/navigation/SideBar.vue
@@ -87,8 +87,7 @@ const toggleOpen = (item: NavigationEntry) => {
 
 <template>
     <div
-        @scroll.stop
-        class="flex max-h-screen grow flex-col gap-y-5 overflow-y-auto border-r border-zinc-200 bg-zinc-100 px-6 pb-4"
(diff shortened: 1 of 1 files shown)


cms/src/components/BasePage.vue
```diff
@@ -17,6 +15,7 @@ type Props = {
     backLinkText?: string;
     backLinkParams?: Record<string, string | undefined>;
     isFullWidth?: boolean;
+    onOpenMobileSidebar?: () => void;
```
DirkDevelopmentFreelance: Will take a look at this implementation in more detail. It works, but think we could simplify the Sidebar and mobile sidebar into one single component and move showSidebar from `App.vue` to the `BasePage.vue`. Which would remove the need for this, unfortunately 0 CSS is not possible here, as this is just used for the button. No other javascript is used. And the function is just to set as:
```ts 
showSidebar = true
```

DirkDevelopmentFreelance: Actually - This is the cleanest approach in my opinion. @johan-bell , what do you think?

PR #1197, opened 2025-11-03
ISR: Add shouldHidePrivacyPolicy logic to conditionally display privacy policy modal

MrDirkelz:
This is the cleanest approach to hide the privacy policy in the static site generator. Because if we should decide to change how the privacy policy works in the future, it will break the check in the static site generator/isr-system


cms/.env.example
```diff
@@ -11,6 +11,8 @@ VITE_AUTH0_DOMAIN=your.auth0.domain
 VITE_AUTH0_AUDIENCE=https://your.audience/api
 VITE_AUTH0_CLIENT_ID=FILL_IN
 
+VITE_SHOULD_HIDE_PRIVACY_POLICY=false
```
ivanslabbert:
```suggestion
VITE_HIDE_PRIVACY_POLICY=false
```

ivanslabbert (approved):

PR #1198, opened 2025-11-03
Fix 404 Page not displaying

### LINKED ISSUE #1196: APP: Routing to home page when opening slug that is not in IndexedDb
To recreate:

1. Open an incognito / inprivate browser window
2. Navigate to a path that contains a publically available slug

Result:
The app routes to the home page and not to the post

This was working before, so we need to add tests to fix this

MrDirkelz:


app/src/pages/SingleContent.spec.ts
ivanslabbert: Add a test that will ensure that a slug renders correctly when not in IndexedDb but available in the online API

ivanslabbert (commented):
Momentarily shows 404 error before loading content from slug. We need to wait until the API confirmed that the slug is not available before showing an 404. When offline however, we show the 404 if the slug is not in IndexedDb

DirkDevelopmentFreelance:
This was too complex of a solution

PR #1208, opened 2025-11-10
Refactor syncLocalChanges tests to simplify formData assertions

MrDirkelz:


api/src/endpoints/changeRequest.controller.ts
```diff
@@ -33,36 +35,43 @@ export class ChangeRequestController {
     ) {
         const token = authHeader?.replace("Bearer ", "") ?? "";
 
-        if (files?.length || typeof body["changeRequestDoc-JSON"] === "string") {
-            const apiVersion = body["changeRequestApiVersion"];
+        // Check if this is a multipart form data request (with potential binary data)
+        // LFormData uses __json suffix instead of -JSON
```
ivanslabbert: Check comment?

ivanslabbert (approved):

api/src/endpoints/changeRequest.controller.ts
```diff
@@ -71,6 +80,8 @@ export class ChangeRequestController {
 
         // If it is just a JSON object (not multipart), validate it correctly
         await validateApiVersion(body.apiVersion);
-        return this.changeRequestService.changeRequest(body, token);
+        // Clean prototype pollution from the body before processing
+        const cleanedBody = removeDangerousKeys(body);
```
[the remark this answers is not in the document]
DirkDevelopmentFreelance: Just realised this needs to be here because this is for the part where it is just json, I made this be comepletely generic so that this endpoint can still be used in other places where we don't want the complexity of multipart.

api/src/changeRequests/documentProcessing/processImageDto.ts
```diff
@@ -293,10 +292,12 @@ async function processImageUpload(
...
-            warnings.push(
-                "No bucket specified for image upload. Each post/tag must specify a target bucket with proper credentials.",
-            );
-            return { success: false, warnings };
+            return {
+                success: false,
+                warnings: [
+                    "No bucket specified for image upload. Each post/tag must specify a target bucket with proper credentials.",
+                ],
+            };
```
ivanslabbert: Check naming. Bucket is only used when configuring storage. Perhaps refer to "Storage not specified..."

api/src/changeRequests/documentProcessing/processImageDto.ts
```diff
@@ -408,7 +417,7 @@ async function resizeAndUploadImage(
         const imageFile = new ImageFileDto();
         imageFile.width = resized.info.width;
         imageFile.height = resized.info.height;
-        imageFile.filename = uuidv4();
+        imageFile.filename = uuidv4() + ".webp";
```
ivanslabbert: This should not be needed. Make sure you set the mime type when uploading.

PR #1218, opened 2025-11-21
refactor: Remove Headless UI components from SideBar and simplify navigation logic

ChrisTouo:


cms/src/components/navigation/SideBar.vue
```diff
@@ -98,75 +99,70 @@ const toggleOpen = (item: NavigationEntry) => {
...
-                        >
-                            <li v-for="subItem in item.children" :key="subItem.name">
-                                <DisclosureButton
-                                    :as="RouterLink"
-                                    :to="subItem.to"
-                                    active-class="bg-zinc-200 text-zinc-900"
-                                    class="block rounded-md py-2 pl-9 pr-2 text-sm font-medium leading-6 text-zinc-700 hover:bg-zinc-200"
-                                    @click="$emit('close')"
+        <nav class="flex flex-1 flex-col">
+            <ul role="list" class="flex flex-1 flex-col justify-between gap-y-7">
```
DirkDevelopmentFreelance: Is `gap-y-7` necessary? From what I saw visually it didn't really seem to affect the sidebar?

DirkDevelopmentFreelance (commented):
This is good work. 

Just a few comments
1. In the shared image, the sidebar is lacking the padding it used to have. @ivanslabbert mentioned we should perhaps remove this, not sure if he still feels this way, otherwise we should just re-add the padding. 
2. Shouldn't the logo perhaps be centered horizontally?
3. The list has a big margin to the logo, and it looks a bit unaligned with the topbar to me. 
Image

4. Maybe there needs to be a bit of bottom padding below the user?
Image

PR #1238, opened 2025-12-03
refactor: Update comments for clarity and improve layout structure in EditContent component

MrDirkelz:


cms/src/components/content/EditContent.vue
ivanslabbert: EditContentParent got removed?

PR #1239, opened 2025-12-03
refactor: Update Language components to use DisplayCard for consistency and improve slot naming conventions

ChrisTouo:


cms/src/components/languages/LanguageOverview.vue
```diff
@@ -35,7 +36,10 @@ const createNew = () => {
                 />
             </div>
         </template>
-
-        <LanguageTable />
```
ivanslabbert: Delete the LanguageTable vue component file if not used anymore

PR #1241, opened 2025-12-03
test: Enhance RichTextEditor tests to cover external prop changes, content clearing, and invalid JSON handling

MrDirkelz:


cms/src/components/content/EditContent.vue
```diff
@@ -618,90 +618,51 @@ const isLocalChange = db.isLocalChangeAsRef(parentId);
...
-                                    :disablePublish="!canPublish"
-                                />
-                            </div>
-                        </div>
                     </div>
                 </div>
             </div>
-            <!-- main content | This is for desktop-->
-            <div class="hidden w-full scrollbar-hide lg:block">
+            <!-- main content | Single editor instance for both mobile and desktop -->
```
ivanslabbert: Comment stating that it is for both mobile and desktop is not needed?

PR #1251, opened 2025-12-05
Remove Headless UI components from OnlineIndicator.vue and replace with LDropdown

ChrisTouo:


cms/src/components/OnlineIndicator.vue
```diff
@@ -1,43 +1,34 @@
 <script setup lang="ts">
-import { Popover, PopoverButton, PopoverPanel } from "@headlessui/vue";
 import { SignalIcon, SignalSlashIcon } from "@heroicons/vue/20/solid";
+import { ref } from "vue";
+import LDropdown from "./common/LDropdown.vue";
```
ivanslabbert: Use "@" import

PR #1269, opened 2025-12-24
audio switching language track persistence

johan-bell:


ivanslabbert (commented):
Check if you can find the reason why there is a delay when switching to audio mode?
Also test that language settings are kept when switching to full-screen mode

app/src/components/content/VideoPlayer.vue
```diff
@@ -437,59 +447,81 @@ watch(audioMode, async (mode) => {
...
-        player?.play();
-
-        // Wait for tracks to be available
-        player?.on("loadedmetadata", () => {
-            const newTracks = (player as any).audioTracks?.();
-
-            if (!newTracks || !selectedTrackInfo) return;
+    const adjustedTime = Math.max(0, currentTime - (mode ? AUDIO_MODE_TIME_ADJUSTMENT : 0));
 
+    // Helper function to restore the selected audio track
```
ivanslabbert:
```suggestion
    // Helper function to restore the selected audio track when switching back to video mode
```

PR #1277, opened 2026-01-07
fix: Fix content not displaying by adding an immediate to sync watcher in the app

odalamkundika-hue:


MrDirkelz:
> Meeting with @odalamkundika-hue on Feb 4 Wednesday 2026 (notes for discussion)
> 
> When saving/creating documents, the `processContentDoc` function adds the parent doc's tags to the content document it is adding with no checks - it just does it
> 
> When duplicating, we want to remove the cached parent tags on the content documents
> 
> We tried doing thatt in the client code. But it is overriden from the `processContentDoc` function.
> 
> @ivanslabbert do you have a way we can approach this?

Single source of truth for cached docs in api
Api handles caching of fields in dto

MrDirkelz:
> Hey @MrDirkelz can you tell me what this PR is about ?

Check ticket description? Think it is added there

PR #1281, opened 2026-01-09
Create shared docker dev environment setup

### LINKED ISSUE #1273: Project: Create Docker Development environment for better development experience on various development platforms

MrDirkelz:


ivanslabbert (commented):
- [x] Create a separate top level folder for the dev docker(s).
- [ ] Do not use hidden directory
- [ ] Add documentation on how to debug the API (add breakpoints, step through code, etc.)
- [ ] Ensure that the .env.example file is such that it can be used "out the box" to start a new dev environment (just copy .env.example to .env)

Use the setup on your machine to test it properly.

DirkDevelopmentFreelance:
this is starting to feel unviable, especially for low end computers. The reason cpu and ram usage is high is because vscode is indexing the files from a server it installs to be able to connect to the docker instance. I have tried optimisations like it ignore node modules, etc

Maybe we should rather have 1 sub-project (for example app has it's comepletely own container) and then when they work on it, they just open the app itself in this container.

As well as the auth0 variables and jwt secret still has to be set, as well as the JWT mapping (so the idea with automating it is not valid here)

johan-bell (commented):
Just check comments.

PR #1288, opened 2026-01-14
Fix group editor visual bug

MrDirkelz:


cms/src/components/groups/EditGroup.vue
ivanslabbert: Check unused import `watch`

ivanslabbert (approved):
Approved with 1 comment

PR #1293, opened 2026-01-16
fix: LCombobox dropdown appearing inside bounds of container cutting of content

MrDirkelz:
This is a temporary fix for now. LDropdown.vue will be used here in #1210. so LDropdown will get an update later probably to improve it and make it work perfecly


ivanslabbert:
Approved, but please fix tests

PR #1302, opened 2026-01-23
feat: Add RedirectDisplayCard component for improved redirect visuali…

ChrisTouo:
…zation

### DIFF
--- a/cms/src/components/BasePage.vue
+++ b/cms/src/components/BasePage.vue
@@ -5,6 +5,7 @@ import { type Component } from "vue";
 import { RouterLink, useRouter, type RouteLocationRaw } from "vue-router";
 import TopBar from "./navigation/TopBar.vue";
 import { isSmallScreen } from "@/globalConfig";
+import { breakpointsTailwind, useBreakpoints } from "@vueuse/core";
 
 type Props = {
     title?: string;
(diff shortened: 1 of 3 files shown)


cms/src/components/modals/LModal.vue
```diff
@@ -53,8 +55,8 @@ function handleMouseUp(e: MouseEvent) {
                 <h2 class="shrink-0 px-5 pb-4 pt-5 text-lg font-semibold">{{ heading }}</h2>
 
                 <div :class="contentClasses">
-                    <div class="divide-y divide-zinc-200">
-                        <slot></slot>
+                    <div :class="noDivider ? '' : 'divide-y divide-zinc-200'">
```
DirkDevelopmentFreelance:
```suggestion
                    <div :class="{
                         'divide-y divide-zinc-200':  !noDivider
                    }">
```

Just check if this is correct syntax? But think it is

cms/src/components/redirects/CreateOrEditRedirectModal.vue
```diff
@@ -159,93 +190,103 @@ const deleteRedirect = () => {
...
+                groups.map((group: GroupDto) => ({
+                    id: group._id,
+                    label: group.name,
+                    value: group._id,
+                }))
+            "
+            :selectedOptions="editable.memberOf"
+            placeholder="Select groups that can access this redirect"
+            class="w-full"
+            :showIcon="true"
```
DirkDevelopmentFreelance:
```suggestion
            showIcon
```

In html just writing a boolean attribute like this is the same as writing (:)showIcon="true"

ChrisTouo:
> Hi @ChrisTouo did you fix all the changes ?

Yess

PR #1305, opened 2026-01-28
fix: Remove changerequest ids to fix conversion errors

MrDirkelz:
PR #1208 added a @Type(() => Number) decorator to the ChangeReqDto.id field to ensure proper type conversion from strings to numbers. This is standard practice when working with class-transformer for DTOs.

However, the validation code was calling plainToInstance() without the enableImplicitConversion: true option. Without this flag, the @Type() decorators are ignored, so:

The id remained a string (as it comes from JSON.parse)
The @IsNumber() validator rejected it
All change requests failed validation
What I fixed:
I added enableImplicitConversion: true to all three plainToInstance() calls in the validation pipeline:

The main ChangeReqDto transformation
The document type-specific DTO transformation (PostDto, TagDto, etc.)
The LanguageDto transformation in access validation

enableImplicitConversion is an option in the class-transformer library that controls whether type decorators like @Type() are applied during transformation.

Without enableImplicitConversion, the @Type(() => Number) decorator is skipped

### DIFF
--- a/api/src/main.ts
+++ b/api/src/main.ts
@@ -4,6 +4,7 @@ import { AppModule } from "./app.module";
 import { upsertDesignDocs, upsertSeedingDocs } from "./db/db.seedingFunctions";
 import { DbService } from "./db/db.service";
 import { PermissionSystem } from "./permissions/permissions.service";
+import { upgradeDbSchema } from "./db/db.upgrade";
 import { S3Service } from "./s3/s3.service";
 import { ValidationPipe } from "@nestjs/common";
 import compress from "@fastify/compress";
@@ -43,6 +44,9 @@ async function bootstrap() {
 
     // Initialize S3Service change feed listener for automatic credential updates
     S3Service.initializeChangeListener(dbService);
+
+    // Upgrade database schema if needed
+    await upgradeDbSchema(dbService);
 
     app.enableCors({
         origin: JSON.parse(process.env.CORS_ORIGIN),


shared/src/db/database.ts
```diff
@@ -694,19 +696,24 @@ class Database extends Dexie {
...
         if (ack.ack == "accepted" && ack.warnings && ack.warnings.length > 0) {
             changeReqWarnings.value = ack.warnings;
         }
 
-        await this.localChanges.delete(ack.id);
+        // Only delete from queue if change was accepted
+        const changes = await this.localChanges.toArray();
+        if (changes.length > 0 && changes[0].id) {
+            await this.localChanges.delete(changes[0].id);
+        }
```
ivanslabbert: See above comment. This code should also be reverted

shared/src/db/database.ts
```diff
@@ -685,19 +685,19 @@ class Database extends Dexie {
...
             if (ack.docs && Array.isArray(ack.docs)) {
                 // Replace our local copy(s) with the provided database version
                 await this.docs.bulkPut(ack.docs);
             } else {
                 // Otherwise attempt to delete the item, as it might have been a rejected create action
-                const change = await this.localChanges.get(ack.id);
-
-                if (change?.doc) {
-                    await this.docs.delete(change.doc._id);
+                if (localChange?.doc) {
```
ivanslabbert: check if "if" is necessary?

shared/src/db/database.ts
```diff
@@ -706,7 +706,9 @@ class Database extends Dexie {
             changeReqWarnings.value = ack.warnings;
         }
 
-        await this.localChanges.delete(ack.id);
+        if (localChange?.id) {
+            await this.localChanges.delete(localChange.id);
+        }
```
ivanslabbert: We removed id's, check if this is still needed? (unless it is needed for cleanup of old change requests)

PR #1337, opened 2026-02-16
feat: Integrate language handling for deleteCmd entries in syncBatch and trim functions

MrDirkelz:


ivanslabbert (commented):
Add an upgrade script to add the language to the delete commands where the content document has not been deleted (DeleteCmd's generated due to permission or publish status changes). Important to not update the UpdatedDateTimeUtc field, as this will move the delete commands forward in time-space, and cause unwanted deletions.

api/src/db/schemaUpgrade/v11.ts
```diff
@@ -0,0 +1,54 @@
...
+                    const contentResult = await db.getDoc(doc.docId);
+                    if (
+                        contentResult.docs.length > 0 &&
+                        contentResult.docs[0].type === DocType.Content
+                    ) {
+                        const contentDoc = contentResult.docs[0];
+                        if (contentDoc.language) {
+                            doc.language = contentDoc.language;
+                            // Use insertDoc directly to preserve the existing updatedTimeUtc
+                            await db.insertDoc(doc);
```
ivanslabbert: db.insertDoc is not handling document update conflicts (see upsertDoc, so you need to handle it manually. Or the error handling logic could perhaps be moved to insertDoc to prevent code duplication, and solve another issue where the API occasionaly rejects db doc updates (through upsertDoc) when upsertDoc is called repetitively for the same doc.

api/src/socketio.ts
```diff
@@ -32,6 +32,7 @@ type ClientDataReq = {
 type ClientConfig = {
     maxUploadFileSize: number;
     maxMediaUploadFileSize?: number;
+    accessMap: AccessMap;
```
ivanslabbert: Please explain why this is added? Not sure it is part of this PR's scope?

shared/src/rest/sync2/syncBatch.ts
```diff
@@ -46,6 +46,11 @@ export async function syncBatch(options: SyncOptions) {
     // Add docType selector for deleteCmd queries
     if (options.type === DocType.DeleteCmd && options.subType) {
         mangoQuery.selector.docType = options.subType;
+
+        // Filter DeleteCmds by language for content-based delete commands
+        if (options.languages && options.languages.length > 0) {
+            mangoQuery.selector.language = { $in: options.languages };
+        }
```
ivanslabbert: When changing the structure of the query selector, check that the validation function in the API for the specific query will still allow the updated query selector.

shared/src/rest/sync2/sync.spec.ts
```diff
@@ -23,9 +23,13 @@ vi.mock("./trim", () => ({
...
-vi.mock("./merge", () => ({
-    merge: vi.fn(),
-}));
+vi.mock("./merge", async () => {
+    const actual = await vi.importActual("./merge");
+    return {
+        ...actual,
+        merge: vi.fn(),
+    };
+});
```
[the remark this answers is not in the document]
MrDirkelz: It was mocked before? I just ensured it adds all correct methods, otherwise it is not fully mocked (note `...actual`)

Image

PR #1360, opened 2026-02-27
Fix a mistake made in Edit Group and create a test for that component

ChrisTouo:


cms/src/components/groups/EditGroup.spec.ts
```diff
@@ -502,6 +504,24 @@ describe("EditGroup", () => {
             expect(wrapper.find('[data-test="groupNameInput"]').exists()).toBe(true);
         });
 
+        it.only("should be able to add new groups", async () => {
```
ChrisTouo:
```suggestion
        it("should be able to add new groups", async () => {
```

PR #1371, opened 2026-03-10
feat: Implement redirect generation for multiple translations

johan-bell:


ivanslabbert (approved):
According to the test written it seems as if the functionality is now correctly implemented

PR #1372, opened 2026-03-10
feat: Exclude Tag and Post documents when contentOnly is true

johan-bell:


ivanslabbert:
When merging, set the commit/merge message to something like: add test to verify that xxxxxxx

PR #1373, opened 2026-03-10
remove statusChange delete commands when content is republished from draft/expired state

johan-bell:


api/src/db/db.upgrade.ts
ivanslabbert: We have a V13 upgrade script from another feature. Revert these changes

ivanslabbert (approved):
Approved with 1 comment

PR #1377, opened 2026-03-13
Multi openid auth support implementation

### LINKED ISSUE #1223: Generic OpenID Auth integration
> As per discussion with BCC IT, the recommended way is to use our own auth system, and link BCC logins to it (the same way Hidden Treasures publishers does it) Ory Kratos looks like a good solution for an auth backend, but needs more investigation.
> 
> Need to create a separate task for this, and decide what is part of Luminary and what should be kept separate. Probably the auth system should be kept separate from Luminary, and Luminary needs a way to be compatible with any OpenID Auth provider. 

 _Originally posted by @ivanslabbert in [#1043](https://github.com/bccsa/luminary/issues/1043#issuecomment-3579752580)_

MrDirkelz:


MrDirkelz (commented):
Some comments and changes to be made

MrDirkelz (commented):
Sidecar file so that not everything that has to be synced to the clients that it does not need. 

Helper text at the top of the right column for the auth provider form modal to make it easier for admins to understand

Better name  for global confCMS: remove assign to  group 

Imager

shared/src/db/database.ts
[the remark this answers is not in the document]
MrDirkelz: he problem: IndexedDB's structured clone algorithm

When you store data in IndexedDB via Dexie, the browser's structured clone algorithm serializes the object. This algorithm cannot handle Vue reactive Proxy objects — it throws a DataCloneError when it encounters them.

Why toRaw alone isn't enough

toRaw is shallow — it unwraps one level of Vue's reactive Proxy, but only on the top-level object. If options.doc has nested objects (e.g. doc.acl, doc.memberOf[], nested DTOs), those nested values remain wrapped in reactive Proxies. The structured clone algorithm then hits one of those nested Proxies and fails.

Why cloneDeep alone isn't enough

cloneDeep performs a deep copy, but it traverses the object through Vue's Proxy traps. The resulting deep copy may still contain Proxy-wrapped values for deeply reactive nested structures.

Why the combination works

toRaw first strips the outer Proxy, giving cloneDeep a cleaner starting point. cloneDeep then recursively copies every nested value, and because it's traversing through the already-raw top-level object, lodash's cloning logic produces plain objects/arrays/primitives throughout — no Proxies survive.

What triggered this: The upsert method is called with Vue reactive objects passed down from components or composables. Before this change, saving a document that had reactive nested fields (like acl maps or memberOf arrays touched by Vue reactivity) would silently fail or throw a DataCloneError in the browser.

.github/workflows/api-unit-tests.yml
MrDirkelz, line 1: Revert

api/src/auth/authIdentity.service.ts
MrDirkelz, line 1: Create a ticket to ensure that all singleton instances of cached values for most of our dtos are cleared on database disconnection

.github/workflows/api-unit-tests.yml
```diff
@@ -82,10 +82,10 @@ jobs:
...
+      # - name: Lint
+      #   run: npm run lint
+      #   working-directory: api
 
-      - name: Type Check
-        run: npx tsc
-        working-directory: api
+      # - name: Type Check
+      #   run: tsc
+      #   working-directory: api
```
ivanslabbert: This should not be done in this ticket. Please add separate ticket?

api/src/db/seedingDocs/lang-eng.json
```diff
@@ -121,6 +121,9 @@
         "openapp_warning.description": "We noticed you're in Telegram. Please click on Continue to see the content properly on the {appName} app.",
         "openapp_warning.button_continue": "Continue",
         "openapp_warning.button_cancel": "Cancel",
-        "singlecontent.listen": "Listen"
+        "singlecontent.listen": "Listen",
+        "auth.sign_in": "Sign in",
+        "auth.no_methods_available": "No sign-in methods available.",
+        "auth.contact_support": "Please contact support for assistance."
```
ivanslabbert: I think we can remove this message - people will not be able to contact support when not logged in

api/src/db/seedingDocs/lang-fra.json
```diff
@@ -121,6 +121,9 @@
         "openapp_warning.description": "Nous avons remarqué que vous êtes sur Telegram. Veuillez cliquer sur Continuer pour voir le contenu correctement dans l'application {appName}.",
         "openapp_warning.button_continue": "Continuer",
         "openapp_warning.button_cancel": "Annuler",
-        "singlecontent.listen": "Écouter"
+        "singlecontent.listen": "Écouter",
+        "auth.sign_in": "Se connecter",
+        "auth.no_methods_available": "Aucune méthode de connexion disponible.",
+        "auth.contact_support": "Veuillez contacter le support pour obtenir de l'aide."
```
ivanslabbert: See comment in lang-eng.json

api/src/endpoints/storageStatus.controller.ts
```diff
@@ -29,18 +28,15 @@ export class StorageStatusController {
     async getStorageStatus(
         @Query("bucketId") bucketId: string,
         @Query("apiVersion") apiVersion: string,
-        @Headers("Authorization") authHeader: string,
+        @Req() request: any,
```
ivanslabbert: Why are we using the "any" type?

api/src/socketio.ts
```diff
@@ -92,24 +93,37 @@ export class Socketio implements OnGatewayInit {
...
+                        groups: defaultGroups,
+                        accessMap: PermissionSystem.getAccessMap(defaultGroups),
+                    } as JwtUserDetails;
+                }
+            } else {
+                const defaultGroups = await this.authIdentityService.getDefaultGroups();
+                socket.data.userDetails = {
+                    groups: defaultGroups,
+                    accessMap: PermissionSystem.getAccessMap(defaultGroups),
+                } as JwtUserDetails;
```
ivanslabbert: Is this duplication of logic?

api/scripts/add-auth-provider.ts
ivanslabbert, line 1: move to a `scripts` folder?

cms/src/components/authProvider/LabelAppearance.vue
ivanslabbert, line 1: Do we need to prefix all the components with AuthProvider when they are in a AuthProvider folder? It makes it a bit hard to read / distinguish between files

ivanslabbert (commented):
check prettier settings. There are many changes that are formatting changes, causing unneccesary edits

api/src/auth/auth.guard.ts
```diff
@@ -1,26 +1,18 @@
...
     async canActivate(context: ExecutionContext): Promise<boolean> {
         const request = context.switchToHttp().getRequest<FastifyRequest>();
         const token = this.extractTokenFromHeader(request);
-        if (!token) {
-            throw new UnauthorizedException();
-        }
-        try {
-            const payload = await this.jwtService.verifyAsync(token, {
-                secret: process.env.JWT_SECRET,
-            });
```
ivanslabbert, line 18: We should still throw an exception when auth fails. Non-logged in users will get public access, but when auth is failing the user need to know that, and the user app can auto-logout the user or handle it in an appropriate manner. Perhaps add a note in the code on why we still throw an exception on failed auth

api/src/changeRequests/documentProcessing/processImageDto.spec.ts
```diff
@@ -615,7 +615,8 @@ describe("S3ImageHandler - File Type Validation", () => {
         // Should fail with file type warning (jpeg not allowed when only webp is allowed)
         // Note: Sharp detects "jpeg" format but the code normalizes it to "jpg" in the warning
         const fileTypeWarning = warnings.warnings.find(
-            (w) => w.includes("not allowed") && (w.includes("image/jpg") || w.includes("image/jpeg")),
+            (w) =>
+                w.includes("not allowed") && (w.includes("image/jpg") || w.includes("image/jpeg")),
```
ivanslabbert: formating change, please revert

api/src/db/db.upgrade.spec.ts
```diff
@@ -1,8 +1,23 @@
...
+    default: jest.fn().mockResolvedValue(undefined),
+}));
+jest.mock("./schemaUpgrade/v12", () => ({
+    __esModule: true,
+    default: jest.fn().mockResolvedValue(undefined),
+}));
+jest.mock("./schemaUpgrade/v13", () => ({
+    __esModule: true,
+    default: jest.fn().mockResolvedValue(undefined),
+}));
```
ivanslabbert: Formating change, please revert

api/src/db/seedingDocs/group-languages.json
ivanslabbert, line 1: check formatting? Arrays should not expand if only having one entry?

api/src/db/seedingDocs/group-private-content.json
ivanslabbert, line 1: check formatting? Arrays should not expand if only having one entry?

api/src/db/seedingDocs/group-private-editors.json
ivanslabbert, line 1: check formatting? Arrays should not expand if only having one entry?

api/src/db/seedingDocs/group-private-users.json
ivanslabbert, line 1: check formatting? Arrays should not expand if only having one entry?

api/src/db/seedingDocs/group-private-users.json
```diff
@@ -6,37 +6,82 @@
...
+        },
+        {
+            "type": "authProvider",
+            "groupId": "group-super-admins",
+            "permission": [
+                "view",
+                "edit",
+                "delete",
+                "assign"
+            ]
```
ivanslabbert: Not needed - revert

api/src/db/seedingDocs/group-private-editors.json
```diff
@@ -6,37 +6,80 @@
...
+        },
+        {
+            "type": "authProvider",
+            "groupId": "group-super-admins",
+            "permission": [
+                "view",
+                "edit",
+                "delete",
+                "assign"
+            ]
```
ivanslabbert: Not needed - revert

api/src/db/seedingDocs/group-private-content.json
```diff
@@ -6,42 +6,82 @@
...
+            "permission": [
+                "view"
+            ]
+        },
+        {
+            "type": "authProvider",
+            "groupId": "group-private-editors",
+            "permission": [
+                "view"
+            ]
```
ivanslabbert: Not needed - revert

api/src/db/seedingDocs/group-languages.json
```diff
@@ -6,42 +6,88 @@
...
+            "permission": [
+                "view"
+            ]
+        },
+        {
+            "type": "authProvider",
+            "groupId": "group-private-editors",
+            "permission": [
+                "view"
+            ]
```
ivanslabbert: Not needed - revert

api/src/db/seedingDocs/group-public-editors.json
```diff
@@ -6,32 +6,72 @@
...
+        },
+        {
+            "type": "authProvider",
+            "groupId": "group-super-admins",
+            "permission": [
+                "view",
+                "edit",
+                "delete",
+                "assign"
+            ]
```
ivanslabbert: Not needed - revert

api/src/db/seedingDocs/group-public-users.json
```diff
@@ -6,37 +6,81 @@
...
+        },
+        {
+            "type": "authProvider",
+            "groupId": "group-super-admins",
+            "permission": [
+                "view",
+                "edit",
+                "delete",
+                "assign"
+            ]
```
ivanslabbert: Not needed - revert

api/src/db/seedingDocs/group-super-admins.json
ivanslabbert, line 1: check formatting?

api/src/db/seedingDocs/group-public-content.json
```diff
@@ -6,77 +6,133 @@
...
+            "permission": [
+                "view"
+            ]
+        },
+        {
+            "type": "authProvider",
+            "groupId": "group-private-users",
+            "permission": [
+                "view"
+            ]
```
ivanslabbert: Not needed. All users will have group-public-users access assigned to them, and will then automatically give access to see auth providers

api/src/db/seedingDocs/group-public-content.json
ivanslabbert, line 1: check formatting

api/src/dto/AuthProviderConfigDto.ts
```diff
@@ -0,0 +1,109 @@
...
+    IsString,
+    IsNotEmpty,
+    IsOptional,
+    ValidateNested,
+    IsArray,
+    IsObject,
+} from "class-validator";
+import { Expose, Type } from "class-transformer";
+
+export class AuthProviderCondition {
```
ivanslabbert: Add JSDoc

api/src/dto/AuthProviderConfigDto.ts
```diff
@@ -0,0 +1,109 @@
...
+    value?: string | string[];
+
+    @IsOptional()
+    @IsArray()
+    @IsString({ each: true })
+    @Expose()
+    values?: string[];
+}
+
+export class AuthProviderGroupMapping {
```
ivanslabbert: add JSDoc

api/src/dto/DefaultPermissionsDto.ts
```diff
@@ -0,0 +1,29 @@
+import { DocType, Uuid } from "../enums";
+import { _baseDto } from "./_baseDto";
+import { IsArray, IsString } from "class-validator";
+import { Expose } from "class-transformer";
+
+export class DefaultPermissionsDto extends _baseDto {
```
ivanslabbert: Add JSDOc

api/src/dto/AuthProviderDto.ts
```diff
@@ -0,0 +1,88 @@
+import { DocType, Uuid } from "../enums";
+import { _baseDto } from "./_baseDto";
+
+import { IsString, IsNotEmpty, IsOptional, IsArray, IsObject, IsNumber } from "class-validator";
+import { Expose } from "class-transformer";
+
+export class AuthProviderDto extends _baseDto {
```
ivanslabbert: Add JSDoc

api/src/dto/UserDto.ts
```diff
@@ -16,9 +16,7 @@ export class UserDto extends _contentBaseDto {
     @Expose()
     name: string;
 
-    /**
-     * External user ID for mapping to external systems
-     */
+    /** @deprecated */
```
ivanslabbert, line 19: Remove? The setup script should move existing ID's to the new externalUserId field and delete the old userId field

api/src/endpoints/changeRequest.controller.ts
```diff
@@ -112,7 +107,10 @@ export class ChangeRequestController {
                 changeRequest.apiVersion = apiVersion;
             }
 
-            const result = await this.changeRequestService.changeRequest(changeRequest, token);
+            const result = await this.changeRequestService.changeRequest(
+                changeRequest,
+                (request as any).user,
```
ivanslabbert: why  "as any"?

api/src/endpoints/changeRequest.controller.ts
```diff
@@ -121,7 +119,10 @@ export class ChangeRequestController {
         await validateApiVersion(body.apiVersion);
         // Clean prototype pollution from the body before processing
         const cleanedBody = removeDangerousKeys(body);
-        const result = await this.changeRequestService.changeRequest(cleanedBody, token);
+        const result = await this.changeRequestService.changeRequest(
+            cleanedBody,
+            (request as any).user,
```
ivanslabbert: why "as any"?

api/src/endpoints/query.controller.ts
```diff
@@ -43,9 +47,6 @@ export class QueryController {
         }
 
         delete body.identifier;
-        return this.queryService.query(
-            body as MongoQueryDto,
-            auth !== undefined ? auth.replace("Bearer ", "") : "",
-        );
+        return this.queryService.query(body as MongoQueryDto, (request as any).user);
```
ivanslabbert: why "as any"?

api/src/endpoints/search.controller.ts
```diff
@@ -1,24 +1,21 @@
...
+    @UseGuards(AuthGuard)
+    async getDocs(@Headers("X-Query") query: string, @Req() request: FastifyRequest): Promise<any> {
         const queryObj = xQuery(query, SearchReqDto);
         await validateApiVersion(queryObj.apiVersion); // validate API version
 
-        return this.searchService.processReq(
-            queryObj,
-            auth !== undefined ? auth.replace("Bearer ", "") : "",
-        );
+        return this.searchService.processReq(queryObj, (request as any).user);
```
ivanslabbert: why "as any"?

api/src/socketio.spec.ts
```diff
@@ -48,7 +47,7 @@ describe("Socketio", () => {
     });
 
     afterEach(() => {
-        clearJwtMap();
+        // JWT state is now managed by AuthIdentityService
     });
```
ivanslabbert: Can be removed

api/src/socketio.ts
```diff
@@ -92,24 +93,24 @@ export class Socketio implements OnGatewayInit {
...
         server.use(async (socket, next) => {
-            // Get automatically assigned group access
-            const userDetails = await processJwt(socket.handshake.auth.token, this.db, this.logger);
+            const token = socket.handshake.auth?.token as string | undefined;
+            const providerId = socket.handshake.auth?.providerId as string | undefined;
 
-            if (socket.handshake.auth.token && !userDetails.jwtPayload) {
-                // Assume that the user's token is expired.
-                // Prompt the user to re-authenticate when an invalid token is provided.
+            const result = await this.authIdentityService.resolveOrDefault(token, providerId);
```
ivanslabbert: Rename `result` to a more appropriate name

api/src/socketio.ts
```diff
@@ -92,24 +93,24 @@ export class Socketio implements OnGatewayInit {
...
+            socket.data.userDetails = result.userDetails;
+
+            if (result.status === "error") {
+                this.logger.error("Socket auth failed for providerId=" + providerId, {
+                    error: result.error instanceof Error ? result.error.message : result.error,
+                });
                 socket.emit("apiAuthFailed");
-                // Disconnect the client to prevent further communication.
-                socket.disconnect(true);
-                return;
```
ivanslabbert, line 109: Keep this logic

api/src/socketio.ts
```diff
@@ -92,24 +93,24 @@ export class Socketio implements OnGatewayInit {
...
-            // Get automatically assigned group access
-            const userDetails = await processJwt(socket.handshake.auth.token, this.db, this.logger);
+            const token = socket.handshake.auth?.token as string | undefined;
+            const providerId = socket.handshake.auth?.providerId as string | undefined;
 
-            if (socket.handshake.auth.token && !userDetails.jwtPayload) {
-                // Assume that the user's token is expired.
-                // Prompt the user to re-authenticate when an invalid token is provided.
+            const result = await this.authIdentityService.resolveOrDefault(token, providerId);
+            socket.data.userDetails = result.userDetails;
```
ivanslabbert: Move to after validation is done (as per previous logic)

api/src/util/ftsIndexing.spec.ts
ivanslabbert, line 1: formatting changes - revert

api/src/util/ftsIndexing.ts
ivanslabbert, line 1: formatting changes - revert

api/src/util/patchFileData.spec.ts
ivanslabbert, line 1: formatting changes - revert

api/src/auth/authIdentity.service.ts
```diff
@@ -0,0 +1,386 @@
...
+            };
+        } catch (error) {
+            if (error instanceof UnauthorizedException) throw error;
+            this.logger.error("Unexpected error resolving identity", error);
+            throw new UnauthorizedException("Authentication failed");
+        }
+    }
+}
+// TODO
+// Transparency on Logo instead of opacity
```
ivanslabbert: Remove?

app/src/components/authProvider/AuthProviderSelectionModal.vue
ivanslabbert, line 1: add unit tests?

app/src/components/images/LImage.vue
ivanslabbert, line 1: update unit tests?

app/src/components/images/LImageProvider.vue
```diff
@@ -299,6 +326,18 @@ watch(
         draggable="false"
         data-test="image-element1"
     />
+    <!-- Icon mode: simple contained rendering, no fallback landscape -->
+    <img
+        v-else-if="isIcon && iconSrc"
+        :src="iconSrc"
+        class="h-full w-full object-contain"
+        alt=""
```
ivanslabbert: Check spec on usage of "alt" in images. We perhaps have to set it to something to be spec compliant?

app/src/components/navigation/PrivacyPolicyModal.vue
ivanslabbert, line 1: only formatting changes - revert

app/src/main.ts
```diff
@@ -61,11 +64,14 @@ async function Startup() {
...
-        Sentry?.captureMessage("API authentication failed, redirecting to login");
-        await auth.loginRedirect(oauth);
+    await setupAuth(app, router);
+    getSocket().connect(); // ensure socket connects for public users (no-op if auth already called reconnect())
+    await resolveProviderId();
+
+    getSocket().on("apiAuthFailed", () => {
+        console.error("API authentication failed, opening provider selection");
+        Sentry?.captureMessage("API authentication failed, opening provider selection");
+        openProviderModal();
```
ivanslabbert: We need to redirect the user directly to the login page for the selected provider if auth is failing. The user should at this point already have made a selection of which provider he wants to use, so we can make use of that provider (stored in localstorage?) to redirect the user directly to the correct login provider.

app/src/sync.ts
```diff
@@ -66,6 +66,18 @@ export function initLanguageSync() {
...
+            if (access[DocType.AuthProvider] && access[DocType.AuthProvider].length) {
+                sync({
+                    type: DocType.AuthProvider,
+                    memberOf: access[DocType.AuthProvider],
+                    limit: 100,
+                    cms: false,
+                }).catch((err) => {
+                    Sentry?.captureException(err);
+                });
+            }
```
ivanslabbert, line 79: Update the function name (initLanguageSync) to reflect the added Auth provider sync (e.g. initAuthLangSync)

cms/src/main.ts
```diff
@@ -80,14 +86,19 @@ async function Startup() {
...
     if (!isAuthBypassed) {
-        socket.on("apiAuthFailed", async () => {
-            console.error("API authentication failed, redirecting to login");
-            Sentry.captureMessage("API authentication failed, redirecting to login");
-            await auth.loginRedirect(oauth);
+        socket.on("apiAuthFailed", () => {
+            console.error("API authentication failed");
+            Sentry.captureMessage("API authentication failed");
+            auth.clearAuth0Cache();
+            auth.openProviderModal();
```
ivanslabbert: Since we should already know which auth provider the user selected (on auth failure), redirect the user directly to the selected auth provider's login

cms/src/sync.ts
```diff
@@ -70,6 +70,48 @@ export function initLanguageSync() {
...
+                    type: DocType.DefaultPermissions,
+                    memberOf: access[DocType.DefaultPermissions],
+                    limit: 100,
+                    cms: true,
+                    includeDeleteCmds: false,
+                }).catch((err) => {
+                    console.error("Error during global config sync:", err);
+                    Sentry?.captureException(err);
+                });
+            }
```
ivanslabbert: Should not be synced as this is only edited directly to the API using ApiLiveQuery

cms/src/sync.ts
```diff
@@ -70,6 +70,48 @@ export function initLanguageSync() {
...
+                    type: DocType.AuthProvider,
+                    memberOf: access[DocType.AuthProvider],
+                    limit: 100,
+                    cms: true,
+                    includeDeleteCmds: true,
+                }).catch((err) => {
+                    console.error("Error during auth provider sync:", err);
+                    Sentry?.captureException(err);
+                });
+            }
```
ivanslabbert, line 89: rename function initLanguageSync to initAuthLangSync

cms/src/sync.ts
```diff
@@ -70,6 +70,48 @@ export function initLanguageSync() {
...
+                    type: DocType.AuthProviderConfig,
+                    memberOf: access[DocType.AuthProviderConfig],
+                    limit: 100,
+                    cms: true,
+                    includeDeleteCmds: true,
+                }).catch((err) => {
+                    console.error("Error during auth provider config sync:", err);
+                    Sentry?.captureException(err);
+                });
+            }
```
ivanslabbert: Do not sync - not needed for offline editing
Auth config is edited directly against the API using ApiLiveQuery

docs/LUMINARY_MONOREPO_SUMMARY.md
ivanslabbert, line 1: remove

docs/datamodel.drawio.svg
ivanslabbert, line 1: Change relationship between AuthProviderDto and AuthProviderConfigDto to a 1:1 relationship, where the ID of the AuthProviderConfigDto is stored in the AuthProviderDto doc. This should simplify indexing and retrieval a bit.

Use a separate permission system doctype entry for the AuthProviderConfigDto to prevent this info being available to public users.

package-lock.json
ivanslabbert, line 1: delete

shared/src/config.ts
```diff
@@ -46,3 +46,9 @@ export function initConfig(newConfig: SharedConfig) {
 
     config = newConfig;
 }
+
+export function setTokenConfig(token: string) {
```
ivanslabbert: add jsdoc

shared/src/rest/http.ts
ivanslabbert, line 1: We should rather just expose a method to set a custom header on requests to the API. The App & CMS should use this to set the x-auth-provider-id, and should store that id after successful login. Auth logic should not be in the shared library, which should be unaware of auth flows (other than providing the auth token). Setting of the auth token can possibly also be simplified by using the same custom header function.

We also need a function to remove custom headers. For the time being we do not need once-off headers, so global header setting and removal is the only changes needed for now. These headers should however not persist between reloads - this will be the responsibiliy of the app / CMS clients.

shared/src/socket/socketio.ts
```diff
@@ -39,7 +40,13 @@ class SocketIO {
      */
     constructor(config: SharedConfig) {
         const token = config.token;
-        this.socket = io(config.apiUrl, token ? { auth: { token } } : undefined);
+
+        this.socket = io(
+            config.apiUrl,
+            token
+                ? { auth: { token, providerId: getActiveProviderId() } }
+                : { autoConnect: false },
```
ivanslabbert: We need to rethink this

shared/src/socket/socketio.ts
```diff
@@ -119,6 +126,24 @@ class SocketIO {
...
 
+    /**
+     * Disconnect and reconnect to the socket server
+     */
+    /**
+     * Update the authentication token
+     * @param token - New JWT access token
+     */
+    public setToken(token: string) {
+        this.socket.auth = { token, providerId: getActiveProviderId() };
```
ivanslabbert: Implement together with the custom headers. Check how custom headers is done in socket.io, and see if we can implement it in the same way as the proposed custom headers for auth

shared/src/util/createEditable/createEditable.ts
```diff
@@ -37,37 +37,22 @@ export function createEditable<T extends BaseDocumentDto>(
...
-        () => {
-            shadow_filtered.value = shadow.value
-                .map((item) => _applyFilter(item))
-                .filter((item) => item !== undefined);
-        },
-        { deep: true, immediate: true },
+    const shadow_filtered = computed(() =>
+        shadow.value
+            .map((item) => _applyFilter(item))
+            .filter((item) => item !== undefined),
```
ivanslabbert: Using computed values will not do deep checking, and break existing functionality

api/src/dto/UserDto.ts
MrDirkelz: The userId is still relevant. This is needed for backwards compatibility, as the externalUserId is from any new auth0 instance, the old auth0 instance will still have the sub for userId. The externalUserId is for any auth provider that isn't the one we already used. So my recommendation is that we keep this. But we can remove this if it is a requirement, but at runtime the script won't actually know what a user might have access to

api/src/changeRequests/documentProcessing/processAuthProviderConfigDto.ts
```diff
@@ -0,0 +1,11 @@
+import { AuthProviderConfigDto } from "../../dto/AuthProviderConfigDto";
+
+/**
+ * Process AuthProviderConfig DTO
+ * - Enforces singleton ID to prevent duplicate docs
+ * - Enforces memberOf to prevent privilege escalation
+ */
+export default async function processAuthProviderConfigDto(doc: AuthProviderConfigDto) {
+    doc._id = "authProviderConfig";
+    doc.memberOf = ["group-super-admins"];
```
ivanslabbert: Cannot hardcode memberof values. This should be assignable in the same manner as other docTypes

api/src/dto/DefaultPermissionsDto.ts
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    /**
+     * Group membership for sync and ACL.
+     * Always enforced to ["group-super-admins"] by processDefaultPermissionsDto.
+     */
+    @IsArray()
+    @IsString({ each: true })
+    @Expose()
+    public memberOf!: Uuid[];
```
ivanslabbert: Extend from _contentBaseDto which includes memberOf?
Also check for the other new DTO's added with this ticket

app/src/auth.ts
```diff
@@ -1,159 +1,323 @@
...
+import { setCustomHeader, removeCustomHeader, getSocket } from "luminary-shared";
+import { db, DocType } from "luminary-shared";
+import type { AuthProviderDto } from "luminary-shared";
 
-export type AuthPlugin = Auth0Plugin & {
-    logout: (retrying?: boolean) => Promise<void>;
-};
+const GUEST_DOMAIN = "guest.auth0.com";
+const GUEST_CLIENT_ID = "guest";
+const GUEST_AUDIENCE = "https://guest";
```
ivanslabbert: Remove

app/src/auth.ts
```diff
@@ -1,159 +1,323 @@
...
+    for (let i = 0; i < 100; i++) {
+        try {
+            await db.docs.count();
+            return;
+        } catch {
+            await new Promise((r) => setTimeout(r, 50));
+        }
+    }
+    throw new Error("auth: db.docs not ready");
+}
```
ivanslabbert: Remove?

app/src/auth.ts
```diff
@@ -1,159 +1,323 @@
...
+    }
+    return {
+        domain: GUEST_DOMAIN,
+        clientId: GUEST_CLIENT_ID,
+        useRefreshTokens: false,
+        cacheLocation: "localstorage" as const,
+        authorizationParams: {
+            audience: GUEST_AUDIENCE,
+            scope: "openid profile email",
+            redirect_uri: webOrigin,
```
ivanslabbert: Remove

app/src/auth.ts
```diff
@@ -1,159 +1,323 @@
...
-        let returnTo = web_origin;
-        if (!retrying) returnTo += "?loggedOut";
+    const { handled, url } = await handleRedirectCallbackIfPresent();
+    if (handled) {
+        const postFootprint = readAuth0NativeStorage();
+        if (postFootprint) {
+            const resolvedProvider = await getProviderByClientId(postFootprint.client_id);
+            if (resolvedProvider) {
+                activeProviderId.value = resolvedProvider._id;
+                setCustomHeader("x-auth-provider-id", resolvedProvider._id);
```
ivanslabbert: Would it be cleaner that the handleRedirectCallbackIfPresent function returns the auth provider ID, and the following logic (outside of auth logic) sets the ID? doing it in the current way is making it hard to find where the header is set (spagetti).

Also check that it is done in the same one for the CMS

app/src/auth.ts
```diff
@@ -1,159 +1,323 @@
...
+                getSocket().reconnect();
+            }
+        } catch {
+            // not authenticated
+        }
+        const path = url.pathname + (url.hash || "");
+        router.replace(path || "/").catch(() => {});
+    } else if (provider) {
+        activeProviderId.value = provider._id;
+        setCustomHeader("x-auth-provider-id", provider._id);
```
ivanslabbert: See above comment

app/src/auth.ts
```diff
@@ -1,159 +1,323 @@
...
+        cacheLocation: "localstorage",
+        authorizationParams: {
+            audience: provider.audience,
+            scope: "openid profile email offline_access",
+            redirect_uri: webOrigin,
+        },
+    });
+    await client.loginWithRedirect({
+        authorizationParams: opts?.prompt ? { prompt: opts.prompt } : undefined,
+    });
```
ivanslabbert, line 197: Check if in use? Looks like a workaround

app/src/auth.ts
```diff
@@ -1,159 +1,323 @@
...
+    activeProviderId.value = null;
+    removeCustomHeader("Authorization");
+    removeCustomHeader("x-auth-provider-id");
 
-    const usedConnection = localStorage.getItem("usedAuth0Connection");
-    const retryCount = parseInt(localStorage.getItem("auth0AuthFailedRetryCount") || "0");
+    try {
+        const socket = getSocket();
+        socket.setAuth("", null);
+        socket.reconnect();
```
ivanslabbert: It is not the auth system's responsibility to reset socket.io connections. See if this can be done a bit more elegantly?

app/src/auth.ts
```diff
@@ -1,159 +1,323 @@
...
-            authorizationParams: {
-                connection: usedConnection ? usedConnection : undefined,
-                redirect_uri: window.location.origin,
-            },
-        });
-        return;
+    const keysToRemove: string[] = [];
+    for (let i = 0; i < localStorage.length; i++) {
+        const key = localStorage.key(i);
+        if (key?.startsWith("@@auth0spajs@@")) keysToRemove.push(key);
```
ivanslabbert, line 216: Add a note that this is specific to Auth0, and might need to be made more generic to support more Auth providers

docs/datamodel.drawio.svg
ivanslabbert, line 1: extend _contentBaseDto?

cms/src/auth.ts
ivanslabbert, line 1: See comments on app/src/auth.ts

api/src/changeRequests/aclValidation.ts
```diff
@@ -35,15 +35,36 @@ const availablePermissionsPerDocType = {
         AclPermission.Translate,
         AclPermission.Publish,
     ],
-    [DocType.User]: [AclPermission.View, AclPermission.Create, AclPermission.Edit, AclPermission.Delete],
-    [DocType.Redirect]: [AclPermission.View, AclPermission.Create, AclPermission.Edit, AclPermission.Delete],
+    [DocType.User]: [
+        AclPermission.View,
+        AclPermission.Create,
```
ivanslabbert: We do not use Create permissions. Also remove from the AclPermission.Create enum entry, and all other .Create settings

api/src/changeRequests/documentProcessing/processAuthProviderConfigDto.spec.ts
```diff
@@ -0,0 +1,169 @@
+import "reflect-metadata";
+import processAuthProviderConfigDto from "./processAuthProviderConfigDto";
+import { AuthProviderConfigDto } from "src/dto/AuthProviderConfigDto";
```
ivanslabbert: Change to relative path

api/src/changeRequests/documentProcessing/processAuthProviderConfigDto.ts
```diff
@@ -0,0 +1,30 @@
...
+    if (!doc.providers || typeof doc.providers !== "object") return;
+
+    for (const providerConfig of Object.values(doc.providers)) {
+        if (!providerConfig || !Array.isArray(providerConfig.groupMappings)) continue;
+        for (const mapping of providerConfig.groupMappings) {
+            if (!mapping || typeof mapping !== "object") continue;
+            const legacy = mapping as unknown as { groupId?: string; groupIds?: string[] };
+            if (Array.isArray(legacy.groupIds)) {
+                // New-shape doc — drop any stale legacy key that might have slipped through.
+                if ("groupId" in legacy) delete legacy.groupId;
```
ivanslabbert: can be removed?

api/src/db/seedingDocs/group-super-admins.json
```diff
@@ -42,6 +42,16 @@
...
+        },
+        {
+            "type": "authProvider",
+            "groupId": "group-super-admins",
+            "permission": ["view", "edit", "delete", "assign"]
+        },
+        {
+            "type": "defaultPermissions",
+            "groupId": "group-super-admins",
+            "permission": ["view", "edit"]
```
ivanslabbert: Also add self-assigned auth-provider-config access

api/src/dto/DefaultPermissionsDto.ts
```diff
@@ -0,0 +1,30 @@
...
+        Object.assign(this, init);
+    }
+
+    /**
+     * Group IDs automatically assigned to all users unauthenticated or authenticated.
+     */
+    @IsArray()
+    @IsString({ each: true })
+    @Expose()
+    public defaultGroups!: Uuid[];
```
ivanslabbert: Do we need the exclamation mark? IT should be mandatory in any case if it does not have a "?"? Please check

api/src/dto/AuthProviderDto.ts
ivanslabbert, line 1: check if "!" marks trailing the public property names are needed?

api/src/filters/allExceptions.filter.ts
ivanslabbert, line 1: Can be removed?

api/src/main.ts
```diff
@@ -46,10 +47,11 @@ export async function bootstrap() {
 
     app.enableCors({
         origin: JSON.parse(process.env.CORS_ORIGIN),
-        allowedHeaders: ["X-Query", "Authorization", "Content-Type"],
+        allowedHeaders: ["X-Query", "Authorization", "Content-Type", "x-auth-provider-id"],
     });
 
     app.useGlobalPipes(new ValidationPipe());
+    app.useGlobalFilters(new AllExceptionsFilter());
```
ivanslabbert, line 55: Check if this is needed?

api/src/main.ts
```diff
@@ -46,10 +47,11 @@ export async function bootstrap() {
 
     app.enableCors({
         origin: JSON.parse(process.env.CORS_ORIGIN),
-        allowedHeaders: ["X-Query", "Authorization", "Content-Type"],
+        allowedHeaders: ["X-Query", "Authorization", "Content-Type", "x-auth-provider-id"],
```
ivanslabbert, line 51: Add comment:
allowedHeaders might have to be made configurable. The shared client library supports injection of custom headers (use for x-auth-provider-id and the Authorization header), and could potentially be used for other custom headers as well.

app/src/util/waitUntilAuth0IsLoaded.ts
```diff
@@ -1,16 +1,17 @@
...
-    const { isLoading } = useAuth0();
-
     const fn = async () => {
         if (callback) await callback();
     };
 
-    if (!isLoading.value) {
-        return fn();
-    }
+    // Nothing to wait for if the Auth0 plugin was never installed at boot.
```
ivanslabbert:
```suggestion
    // Nothing to wait for if the Auth0 plugin was never installed at load.
```

api/src/filters/allExceptions.filter.ts
MrDirkelz: Silences ERR_STREAM_PREMATURE_CLOSE — without this, every time a client aborts a request (navigates away, closes tab, network drop), Nest logs an unhandled exception with a full stack trace. That's pure noise in production logs.

Guards against writing to dead sockets — if (reply.sent || reply.raw.writableEnded) return prevents the "Cannot write headers after they are sent" crash that Fastify throws when a response is already gone.

Consistent JSON error format — Nest's default exception filter can produce inconsistent response shapes depending on where the error originates (middleware vs controller vs pipe). This normalizes everything to { statusCode, message }.

api/src/main.ts
MrDirkelz: Silences ERR_STREAM_PREMATURE_CLOSE — without this, every time a client aborts a request (navigates away, closes tab, network drop), Nest logs an unhandled exception with a full stack trace. That's pure noise in production logs.

Guards against writing to dead sockets — if (reply.sent || reply.raw.writableEnded) return prevents the "Cannot write headers after they are sent" crash that Fastify throws when a response is already gone.

Consistent JSON error format — Nest's default exception filter can produce inconsistent response shapes depending on where the error originates (middleware vs controller vs pipe). This normalizes everything to { statusCode, message }.

api/src/changeRequests/documentProcessing/processAuthProviderConfigDto.spec.ts
MrDirkelz: This file has been removed

api/src/changeRequests/aclValidation.ts
MrDirkelz: Removed

api/src/changeRequests/documentProcessing/processAuthProviderConfigDto.ts
MrDirkelz: This file has been removed

api/src/changeRequests/documentProcessing/processAuthProviderDto.ts
```diff
@@ -0,0 +1,68 @@
...
+            const imageResult = await processImage(
+                { fileCollections: [] },
+                prevDoc.imageData,
+                db,
+                prevDoc.imageBucketId,
+            );
+            if (imageResult?.warnings?.length) {
+                warnings.push(...imageResult.warnings);
+            }
+        }
```
ivanslabbert, line 27: Extract to helper function that can be used in EditContent and here, and whereever we have image support in the future. Perhaps a function named deleteImage (similar to processImage but without needed prev docs)

api/src/changeRequests/documentProcessing/processAutoGroupMappingsDto.ts
```diff
@@ -0,0 +1,6 @@
+import { AutoGroupMappingsDto } from "../../dto/AutoGroupMappingsDto";
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+export default async function processAutoGroupMappingsDto(doc: AutoGroupMappingsDto) {
+    // No singleton enforcement — each mapping is its own document
+}
```
ivanslabbert: Remove?

api/src/changeRequests/validateChangeRequest.ts
```diff
@@ -14,6 +14,9 @@ import { validateChangeRequestAccess } from "./validateChangeRequestAccess";
 import { validateAcl } from "./aclValidation";
 import { RedirectDto } from "../dto/RedirectDto";
 import { StorageDto } from "../dto/StorageDto";
+import { AuthProviderDto } from "../dto/AuthProviderDto";
+import { AutoGroupMappingsDto } from "../dto/AutoGroupMappingsDto";
+import { DefaultPermissionsDto } from "../dto/DefaultPermissionsDto";
```
ivanslabbert: to be removed

api/src/db/seedingDocs/group-super-admins.json
```diff
@@ -42,6 +42,21 @@
...
```
ivanslabbert: Remove defaultPermissions - will be included in autoGroupMappings

api/src/dto/AuthProviderDto.ts
ivanslabbert, line 1: Check tha IsOptional flagged properties are marked with a "?" for typescript type checking

api/src/dto/AutoGroupMappingsDto.ts
ivanslabbert, line 1: Check tha IsOptional flagged properties are marked with a "?" for typescript type checking

api/src/enums.ts
```diff
@@ -22,6 +22,9 @@ export enum DocType {
     DeleteCmd = "deleteCmd",
     Storage = "storage",
     Crypto = "crypto",
+    AuthProvider = "authProvider",
+    AutoGroupMappings = "autoGroupMappings",
+    DefaultPermissions = "defaultPermissions",
```
ivanslabbert: to be removed

cms/src/components/authProvider/ClaimMappings.vue
ivanslabbert, line 1: Remove?

cms/src/components/authProvider/AutoGroupMappingConditions.vue
ivanslabbert, line 1: Check naming. Not sure what this component does looking at the name

ivanslabbert (approved):

PR #1385, opened 2026-03-18
feat: Add loading splash screen with LoadingBar component during app startup

### LINKED ISSUE #1014: APP: Add loading splash screen while Vue app is being loaded
Similar to how the CMS is using a loading splash screen.
Can use the new CMS loading bar instead of the spinner component

MrDirkelz:


ivanslabbert (commented):
update loading animation to be a "running slug", i.e. when the bar is filled from left to right, it starts emptying from left to right, and repeat.

app/src/globalConfig.ts
```diff
@@ -396,25 +396,44 @@ export const theme = computed<"system" | "dark" | "light">({
...
 );
 
 export const fallbackImageUrls = loadFallbackImageUrls();
+
+/**
+ * True while the app's Startup function is still running. Used to display the loading splash screen.
+ */
+export const isAppLoading = ref(
+    !new URLSearchParams(window.location.search).has("nosplash"),
+);
```
johan-bell, line 440: Isn't another to check that the app is loading ?

app/src/main.ts
```diff
@@ -73,12 +79,10 @@ async function Startup() {
...
 
     const i18n = await initI18n();
+    app.use(i18n);
     await loadPlugins();
 
-    app.use(createPinia());
-    app.use(router);
-    app.use(i18n);
-    app.mount("#app");
+    isAppLoading.value = false;
```
johan-bell, line 129: Please check that won't create sequencing issue ?

app/src/globalConfig.ts
```diff
@@ -396,25 +396,44 @@ export const theme = computed<"system" | "dark" | "light">({
...
+
+const applySystemTheme = () => {
+    if (systemThemeQuery.matches) {
+        document.documentElement.classList.add("dark");
+    } else {
+        document.documentElement.classList.remove("dark");
+    }
+    isDarkTheme.value = document.documentElement.classList.contains("dark");
+};
+
```
johan-bell, line 410: Why do we need this ? Is there a specific reason ?

johan-bell (commented):
Just left comments.

app/src/globalConfig.ts
MrDirkelz: The splash lives outside the scope of the loaded app (before vue was loaded)

app/src/globalConfig.ts
MrDirkelz: Image

Nope

app/src/main.ts
MrDirkelz: This flag can't cause sequencing issues, it's just a notifier to the splash screen that is set before we init to true, then here reset to false, the init function is mostly asyncronous, so this won't run contrary to anything

app/index.html
ChrisTouo, line 1: Check your prettier code ?

app/src/globalConfig.ts
```diff
@@ -396,25 +396,44 @@ export const theme = computed<"system" | "dark" | "light">({
  */
 export const isDarkTheme = ref(document.documentElement.classList.contains("dark"));
 
+const systemThemeQuery = window.matchMedia("(prefers-color-scheme: dark)");
```
ivanslabbert, line 400: Add a comment on why we need this

ivanslabbert (commented):
Add the URL query parameter needed to bypass the spash screen (needed for SSG). Remember to document it with the other URL query params

app/src/App.vue
```diff
@@ -133,8 +154,20 @@ onErrorCaptured((err) => {
...
-    <!-- Privacy Policy Modal for authentication flow -->
-    <PrivacyPolicyModal v-model:show="showPrivacyPolicyModal" @close="handleModalClose" />
-    <AuthProviderSelectionModal v-model:isVisible="showProviderSelectionModal" />
+    <!-- Modals depend on i18n, which isn't installed until splash finishes — keep them out of the tree during the loading phase. -->
+    <template v-if="!isAppLoading">
+        <!-- Privacy Policy Modal for authentication flow -->
+        <PrivacyPolicyModal
+            v-model:show="showPrivacyPolicyModal"
+            @close="handleModalClose"
+        />
```
johan-bell: `PrivacyPolicyModal` is rendered twice when !`isAppLoading`: once inside the main layout and again in the outer `<template v-if="!isAppLoading">`. That likely duplicates the modal and can break focus/overlay behavior. Keep a single instance—e.g. remove the one in the `v-else` block and keep the gated block (with `AuthProviderSelectionModal`), or the reverse—so only one tree owns it.

app/src/components/LoadingBar.vue
```diff
@@ -1,13 +1,20 @@
...
+    label?: string;
+};
+defineProps<Props>();
+</script>
+
 <template>
     <div class="flex w-full flex-col items-center gap-4">
         <div class="text-2xl font-semibold text-zinc-500 dark:text-slate-400 max-lg:text-xl">
-            Loading...
+            {{ label }}
```
johan-bell: `label` is optional, and the splash usage currently renders <LoadingBar /> without passing one, so this text can be blank during startup.
Could we set a default (e.g. "Loading...") or pass label explicitly from the splash callsite?

johan-bell (approved):
Left comments, but nice jobs

johan-bell (approved):
Left comments, but nice jobs

Issue #771, opened 2025-03-11
API: socket updates to exclude tag and post docs if flagged with contentOnly


ivanslabbert:

DirkDevelopmentFreelance:
This is implemented inside the db.service.ts ``search`` function. Not sure if it should also be implemented somewhere else

ivanslabbert:
> This is implemented inside the db.service.ts `search` function. Not sure if it should also be implemented somewhere else

Tested in the browser, and "Post" and "Tag" documents are still included.

PR #1396, opened 2026-03-19
771 api socket updates to exclude tag and post docs if flagged with contentonly

johan-bell:


johan-bell:
Closing: its issue #771 was closed in March; this is no longer needed.

PR #1404, opened 2026-03-22
refactor: Replace mangoIsPublished with mangoIsPublishedOrScheduled 

### LINKED ISSUE #1403: APP: Display scheduled content with `Coming soon`

johan-bell:
This is how it looks.

image


MrDirkelz (commented):
I tested this on dev, set a post as scheduled and saved with the "Show as coming soon" set as true, but it did nothing for me, not displaying at all. @johan-bell could you also attach an image of what it is supposed to look like? I also searched for the slug, as well as in the topic for it, but it is not appearing anywhere

MrDirkelz:
> > I tested this on dev, set a post as scheduled and saved with the "Show as coming soon" set as true, but it did nothing for me, not displaying at all. @johan-bell could you also attach an image of what it is supposed to look like? I also searched for the slug, as well as in the topic for it, but it is not appearing anywhere
> 
> There is a toogle in the blog settings. Did you click on that ?

Yes, but maybe you can just give a demo in the stand up meeting

api/src/changeRequests/documentProcessing/processContentDto.spec.ts
```diff
@@ -106,6 +106,7 @@ describe("processContentDto", () => {
             tagType: "topic",
             pinned: 1,
             publishDateVisible: true,
+    showComingSoon: false,
```
MrDirkelz: Just check the formatting in this file?

api/src/db/db.service.spec.ts
```diff
@@ -850,6 +850,7 @@ describe("DbService", () => {
                 postType: PostType.Blog,
                 tags: [],
                 publishDateVisible: true,
+    showComingSoon: false,
```
MrDirkelz: Just check formatting?

api/src/test/changeRequestDocuments.ts
```diff
@@ -33,6 +34,7 @@ export const changeRequest_tag = () => {
             image: randomString(),
             tags: [],
             publishDateVisible: false,
+    showComingSoon: false,
         },
```
MrDirkelz: Formatting?

MrDirkelz (commented):
Just a few comments regarding formatting

api/src/dto/_contentParentDto.ts
```diff
@@ -24,6 +24,10 @@ export class _contentParentDto extends _contentBaseDto {
     @Expose()
     publishDateVisible: boolean;
 
+    @IsBoolean()
+    @Expose()
+    showComingSoon: boolean;
```
ivanslabbert: This should be optional

app/src/util/mangoIsPublished.ts
```diff
@@ -52,6 +52,43 @@ export function mangoIsPublished(languageIds: Uuid[]): MangoSelector[] {
...
+        // Expiry date check: either doesn't exist, is null, or is in the future
+        { $or: [{ expiryDate: { $exists: false } }, { expiryDate: null }, { expiryDate: { $gte: now } }] },
+
+        // Either already published (publishDate in the past/absent) OR scheduled with the flag set
+        {
+            $or: [
+                { publishDate: { $exists: false } },
+                { publishDate: null },
+                { publishDate: { $lte: now } },
+                { $and: [{ publishDate: { $gt: now } }, { parentShowComingSoon: true }] },
```
ivanslabbert: This can be simplified to:
```suggestion
                { parentShowComingSoon: true },
```

app/src/util/mangoIsPublished.ts
```diff
@@ -52,6 +52,43 @@ export function mangoIsPublished(languageIds: Uuid[]): MangoSelector[] {
...
+        { status: PublishStatus.Published },
+
+        // Expiry date check: either doesn't exist, is null, or is in the future
+        { $or: [{ expiryDate: { $exists: false } }, { expiryDate: null }, { expiryDate: { $gte: now } }] },
+
+        // Either already published (publishDate in the past/absent) OR scheduled with the flag set
+        {
+            $or: [
+                { publishDate: { $exists: false } },
+                { publishDate: null },
```
ivanslabbert: This field should not be needed. It would show non-published docs, but as docs without a publish date is not synced to app clients it has no effect

MrDirkelz (changes requested):
# Code Review — PR #1404: Replace `mangoIsPublished` with `mangoIsPublishedOrScheduled`

**Overall impression:** The design is sound. The cascade chain (parent → `processPostTagDto` / `processContentDto` → `parentShowComingSoon` on the child `ContentDto`) follows the exact same pattern as `parentPublishDateVisible`, the Mango query logic is well-documented and well-tested, and the `ContentTile` overlay is clean. A few things need attention before this merges.

---

## 🚨 Critical

### 1. Silent false-positive test in `processPostTagDto.spec.ts`

`api/src/changeRequests/documentProcessing/processPostTagDto.spec.ts`, near line 564:

```ts
expect(contentDoc.docs[0].showComingSoon).toBeFalsy();
```

`contentDoc.docs[0]` is a **`ContentDto`**, which has `parentShowComingSoon` — not `showComingSoon`. `showComingSoon` lives on `ContentParentDto`. Because the field doesn't exist on `ContentDto`, this expression evaluates to `undefined`, which is always falsy. The assertion will pass even if the cascade is completely broken.

**Fix:**
```ts
expect(contentDoc.docs[0].parentShowComingSoon).toBeFalsy();
```

---

## ⚠️ Warning

### 2. `aria-disabled` on a `<div>` is a no-op for accessibility

`app/src/components/content/ContentTile.vue`

```html
<component
    :is="isComingSoon ? 'div' : 'RouterLink'"
    :aria-disabled="isComingSoon || undefined"
    ...
>
```

`aria-disabled` is only meaningful on elements that have an implicit or explicit interactive ARIA role (links, buttons, etc.). A plain `<div>` has no role, so screen-reader users get no signal that this tile is disabled. The `cursor-not-allowed` only helps sighted users.

**Suggested fix** — give the `div` a role so the `aria-disabled` means something:
```html
<div
    v-if="isComingSoon"
    role="img"
    :aria-label="`${props.content.title} — ${t('content.coming_soon')}`"
    class="ease-out-expo group cursor-not-allowed opacity-80 transition"
>
    <!-- tile body -->
</div>
<RouterLink v-else :to="{ name: 'content', params: { slug: props.content.slug } }" ...>
    <!-- tile body -->
</RouterLink>
```

An explicit `v-if / v-else` branch is also simpler to read than the `<component :is>` + `v-bind` conditional dance, which duplicates the entire prop-binding logic.

### 3. `isComingSoon` uses a non-reactive `Date.now()` snapshot

`app/src/components/content/ContentTile.vue`

```ts
const isComingSoon = computed(() => {
    ...
    publishDate > Date.now() &&   // captured at evaluation time, never re-checked
    ...
});
```

If a user loads the page before a content item's publish date and keeps the tab open until after that date, the tile will remain locked behind the "Coming soon" overlay indefinitely — the computed won't invalidate because none of its Vue-reactive dependencies change. The same issue exists in `mangoIsPublishedOrScheduled` where `const now = Date.now()` is captured once per query build.

This may be acceptable for a v1, but it should be a conscious decision. A simple mitigation is a short polling interval (e.g. every minute) that bumps a reactive ref:

```ts
// in a shared composable
const now = ref(Date.now());
setInterval(() => { now.value = Date.now(); }, 60_000);
```

Then replace `Date.now()` with `now.value` in both places.

### 4. "Show as Coming soon" toggle appears for Tags — semantics are unclear

`cms/src/components/content/EditContentParent.vue`

The toggle is rendered unconditionally for every parent doc type (Posts *and* Tags). It's not obvious to a CMS editor what "Show as Coming soon" means for a Category or Topic. The `showComingSoon` flag on a Tag will cascade down to all its child `ContentDto` records via `processPostTagDto.ts`, which means a tag with `showComingSoon: true` effectively marks all its scheduled children as "coming soon" — this is a very wide blast radius.

If this is intentional, add a tooltip or helper text in the UI explaining the Tag case. If it's unintentional, guard the toggle:

```html
<div v-if="docType !== DocType.Tag" ...>
    <FormLabel>Show as Coming soon</FormLabel>
    <LToggle v-model="parent.showComingSoon" ... />
</div>
```

---

## 💡 Nitpick / Suggestion

### 5. Indentation inconsistency in several spec fixtures

In `processContentDto.spec.ts` (and `validateChangeRequest.spec.ts`, `processPostTagDto.spec.ts`), several of the added `showComingSoon: false` lines are at the wrong indentation level — 4 spaces when the surrounding object properties use 8 or 12 spaces. This was likely a copy-paste artefact.

### 6. Double blank line in `EditContentParent.vue`

After the new "Show as Coming soon" `</div>`, there are two consecutive blank lines before the pinned-toggle `<div>`. One is enough.

### 7. First `ContentTile.spec.ts` test hardcodes the English translation string

```ts
expect(wrapper.text()).toContain("Coming soon");   // hardcoded
```

The second test already does this correctly:
```ts
expect(wrapper.text()).not.toContain(mockLanguageDtoEng.translations["content.coming_soon"]);
```

Make the first test consistent so a future translation key rename gets caught:
```ts
expect(wrapper.text()).toContain(mockLanguageDtoEng.translations["content.coming_soon"]);
```

### 8. Spec file naming is inconsistent with its source module

The new tests live in `app/src/util/mangoIsPublishedOrScheduled.spec.ts` but the function under test is exported from `mangoIsPublished.ts`. Either consolidate the specs into `mangoIsPublished.spec.ts`, or move the function into its own module so the names align.

---

cms/src/components/content/EditContentBasic.vue
ChrisTouo, line 1: Check your prettier code ?

ivanslabbert (commented):
Combine the functionality of mangoIsPublishedOrScheduled in mangoIsPublished with an optional property that can be passed to enable it

api/src/changeRequests/documentProcessing/processContentDto.ts
```diff
@@ -36,6 +36,7 @@ export default async function processContentDto(doc: ContentDto, db: DbService)
             doc.parentTaggedDocs = (parentDoc as TagDto).taggedDocs;
         }
         doc.parentPublishDateVisible = parentDoc.publishDateVisible;
+        doc.parentShowComingSoon = parentDoc.showComingSoon ?? false;
```
ivanslabbert:
```suggestion
        doc.parentShowComingSoon = parentDoc.showComingSoon;
```

Keep optional fields optional

app/src/util/mangoIsPublished.ts
```diff
@@ -1,16 +1,26 @@
 import { type Uuid, type MangoSelector, PublishStatus } from "luminary-shared";
 
+export type MangoIsPublishedOptions = {
+    /**
+     * When true, also matches scheduled content: `publishDate` in the future is allowed
+     * if `parentShowComingSoon === true` (for list/carousel "Coming soon" tiles).
+     * Default false: publishDate must be missing, null, or <= now.
+     */
+    includeScheduled?: boolean;
```
ivanslabbert, line 9: Make this by default true as we want to have it set in most cases. The calling function has to explicitly set it to false to disable the feature. This will remove code changes on most implementing files.

shared/src/types/dto.ts
```diff
@@ -114,6 +115,7 @@ export type ContentParentDto = ContentBaseDto & {
     imageData?: ImageDto;
     tags: Uuid[];
     publishDateVisible: boolean;
+    showComingSoon: boolean;
```
ivanslabbert:
```suggestion
    showComingSoon?: boolean;
```

app/src/components/HomePage/HomePageNewest.vue
```diff
@@ -1,4 +1,4 @@
-<script setup lang="ts">
+dd .<script setup lang="ts">
```
ivanslabbert: typo

cms/src/components/content/EditContent.vue
ivanslabbert, line 1: Only set to true if we want to set showComingSoon. Else do not set the property (as it adds weight)

ivanslabbert (approved):

PR #1423, opened 2026-03-26
feat: Sync expired documents so non-cms clients can remove expired content documents after being offline

### LINKED ISSUE #1395: API: Ensure content doc expiry is applied to offline clients when expiry date is changed on a published doc
> APP: Change to the same thinking used for syncing delete commands:
> 
> - when doing a full / initial sync (to the beginning of time), do not include any exired ContentDto docs in the query
> - when doing an update sync (from latest locally available updatedDatetimeUtc to max integer value of updatedDateTimeUtc), include expired ContentDto docs, to ensure the client gets the updated doc with expiry status set, and can do cleanup locally.
> 
> CMS: no change compared to main branch (use conditional logic in shared library sync query to always sync all expired docs) 

 _Originally posted by @ivanslabbert in [#1278](https://github.com/bccsa/luminary/issues/1278#issuecomment-4045087384)_

MrDirkelz:


shared/src/rest/sync2/syncBatch.ts
```diff
@@ -46,6 +52,10 @@ export async function syncBatch(options: SyncOptions) {
         identifier: "sync", // Identifier for the API query validation template
     };
 
+    if (isUpdateSync) {
```
ivanslabbert: Not needed to extract the logic - rather include the conditions directly here
```suggestion
    if (options.type === DocType.Content && !options.cms && !firstSync) {
```

I removed options.initialSync from the condition, as this may cause it to only include expired docs on the first chunk, and ignore it on subsequent chunks. (We cannot assume that the client will receive all new updates in one chunk - especially if the client has been offline for a long while.)

ivanslabbert (commented):
Check merge conflicts

johan-bell (approved):

PR #1434, opened 2026-03-27
Implement Delete functionality in Group editor

### LINKED ISSUE #1432: CMS: Add a delete functionnality for Groups

ChrisTouo:


api/src/db/db.service.ts
```diff
@@ -511,7 +497,7 @@ export class DbService extends EventEmitter {
         const p = options.prevDoc as unknown as _contentBaseDto;
 
         if (options.reason === DeleteReason.Deleted) {
-            cmd.memberOf = p.memberOf;
+            cmd.memberOf = options.doc.type === DocType.Group ? [options.doc._id] : p.memberOf;
```
ChrisTouo: Useless change, have to put it back as it was

MrDirkelz (dismissed):
Image

Please help me delete these by adding a delete button to these cards too? 😅😇

MrDirkelz (commented):
Please create a ticket that for some reason super admins can not be duplicated, works for other groups, but specifically not super admins

Image

cms/src/components/groups/EditGroup.vue
```diff
@@ -196,34 +239,19 @@ const duplicateGroup = async () => {
...
         title=""
         :open="openModal"
         @update:open="(val) => !val && emit('close')"
-        :primaryAction="saveChanges"
-        primaryButtonText="Save"
+        :primaryAction="
+            () => {
+                save(group._id);
+            }
+        "
```
MrDirkelz:
```suggestion
        :primaryAction="save(group._id)"
```

Just check if there was reason for using this notation to ensure this way does not break anything

cms/src/components/redirects/CreateOrEditRedirectModal.vue
```diff
@@ -204,7 +204,7 @@ const revertChanges = () => {
         @close="emit('close')"
         :primaryAction="
             () => {
-                save(), emit('close');
+                save();
             }
         "
```
MrDirkelz: See above comments

cms/src/components/users/CreateOrEditUser.vue
```diff
@@ -178,7 +179,7 @@ const saveDisabled = computed(() => {
         @close.stop="emit('close')"
         :primaryAction="
             () => {
-                save(), emit('close');
+                save();
             }
```
MrDirkelz: Check above comments

ivanslabbert (approved):
Swop delete and duplicate icons on the group editor ACL entry component and reduce the gap between them.
Create separate ticket to find a better looking outline icon that can be used for duplicate buttons throughout the CMS

cms/src/components/groups/EditGroup.vue
ChrisTouo: Get the following error when I write that way @MrDirkelz
Type 'Promise<any>' is missing the following properties from type 'Function'

PR #1457, opened 2026-04-01
fix: Treat "no docs found" as empty list

MrDirkelz:
Bug: When an initial sync finds no documents (e.g., language type had none at the time), the syncList stores blockStart: Number.MAX_SAFE_INTEGER. On the next re-initial-sync (triggered by
  connection/access changes), calcChunk computed blockEnd =
MAX_SAFE_INTEGER - 1000, producing the query {$gte: 9007199254739991, $lte: 9007199254740991} — a near-empty range that matches nothing,
   permanently blocking sync for that type.

  Fix: In calcChunk(), when the existing entry's blockStart is
Number.MAX_SAFE_INTEGER (the sentinel for "no docs found"), treat it the same as an empty list and set blockEnd = 0, so the full range is re-queried.


shared/src/rest/sync2/utils.ts
```diff
@@ -26,7 +26,10 @@ export function calcChunk(options: {
     else blockStart = list[0].blockEnd;
 
     if (options.initialSync)
-        blockEnd = list.length === 0 ? 0 : list[0].blockStart - syncTolerance; // add tolerance to push end into past of first synced blockStart giving an overlap to avoid missing documents
+        blockEnd =
+            list.length === 0 || list[0].blockStart >= Number.MAX_SAFE_INTEGER
```
ivanslabbert:
```suggestion
            list.length === 0 || list[0].blockStart == Number.MAX_SAFE_INTEGER
```
Add comment to explain why we need this (which is fixing the effect of another bug)

ivanslabbert: Perhaps a better way would rather be to have validity checking on the synclist, and delete invalid entries (e.g. entries with a blockstart or blockend value more than 1 hour into the future (compared to the browsers UTC timestamp)

PR #1482, opened 2026-04-07
API, CMS, APP, SHARED: Implement user friendly message notification for server errors

### LINKED ISSUE #1444: API, Clients: Implement system to give user friendly message for server side errors

MrDirkelz:


ivanslabbert (commented):
failing tests?

app/src/main.ts
```diff
@@ -113,6 +114,27 @@ async function Startup() {
...
+    // Show notification on server error (5xx), debounced to avoid flooding
+    let serverErrorTimeout: ReturnType<typeof setTimeout> | null = null;
+    watch(serverError, (error) => {
+        if (error) {
+            serverError.value = null;
+            if (serverErrorTimeout) return;
+            Sentry?.captureMessage(`Server error: ${error}`, "error");
+            useNotificationStore().addNotification({
+                title: "Server error",
+                description: error,
```
johan-bell: `useNotificationStore()` is invoked before Pinia is installed (`app.use(createPinia()`) happens later).
If `serverError` is set during startup, this watcher can run with no active Pinia and throw.
Can we either install Pinia before registering this watcher, or create/reuse the store from an explicit Pinia instance?

cms/src/main.ts
```diff
@@ -131,6 +137,27 @@ async function Startup() {
...
+    // Show notification on server error (5xx), debounced to avoid flooding
+    let serverErrorTimeout: ReturnType<typeof setTimeout> | null = null;
+    watch(serverError, (error) => {
+        if (error) {
+            serverError.value = null;
+            if (serverErrorTimeout) return;
+            Sentry.captureMessage(`Server error: ${error}`, "error");
+            useNotificationStore().addNotification({
+                title: "Server error",
+                description: error,
```
johan-bell: Same ordering issue here. check comment on `main.ts` of the app

cms/src/main.ts
johan-bell, line 176: This watcher also depends on `useNotificationStore()` before Pinia installation.
Even if rare, any early warning/error push during bootstrap can hit the same “no active Pinia” path.
Could we register these watchers only after `app.use(createPinia())`?

johan-bell (dismissed):
Move to changes requested. Please rebase to main

shared/src/rest/http.ts
```diff
@@ -16,13 +18,47 @@ export function removeCustomHeader(name: string) {
...
+async function handleResponse(res: Response) {
+    if (res.ok) {
+        return res.json().catch((err) => {
+            console.log(err.message);
+        });
+    }
+
+    // Server error (5xx) - show user-friendly message
+    if (res.status >= 500) {
+        let message = "Something went wrong on the server. Please try again in a minute.";
```
ivanslabbert: Implement i18n for client-side errors. Error messages should perhaps not be set in shared lib

ivanslabbert (approved):

PR #1493, opened 2026-04-13
CMS: Keep auto-selected storage buckets in lock-step to prevent fake dirty state

MrDirkelz:
MediaEditor and ImageEditor previously wrote the auto-selected storage
bucket onto the parent via a mount-time watchEffect, producing a phantom
dirty state on legacy docs (and on any doc with a single bucket
configured). This required a lock-step watcher in EditContent to keep
editableParent and existingParent in sync.

Introduce `effectiveMediaBucketId` / `effectiveImageBucketId` computed
refs that fall back to the auto-selected bucket for display and
validation without touching the parent. The parent is only written when
the user actually picks a bucket or uploads a file — at which point any
resulting dirty state is legitimate. The lock-step watcher in
EditContent is deleted.

### DIFF
--- a/cms/package.json
+++ b/cms/package.json
@@ -32,7 +32,7 @@
         "luminary-shared": "file:../shared",
         "luxon": "^3.4.4",
         "pinia": "^2.1.7",
-        "rte-vue": "^1.1.6",
+        "rte-vue": "^1.1.11",
         "socket.io-client": "^4.7.4",
         "tailwind-scrollbar-hide": "^4.0.0",
         "tailwindcss": "^3.4.17",
--- a/cms/package.json
+++ b/cms/package.json
@@ -32,7 +32,7 @@
         "luminary-shared": "file:../shared",
         "luxon": "^3.4.4",
         "pinia": "^2.1.7",
-        "rte-vue": "^1.1.11",
+        "rte-vue": "^1.1.13",
         "socket.io-client": "^4.7.4",
         "tailwind-scrollbar-hide": "^4.0.0",
         "tailwindcss": "^3.4.17",
--- a/cms/src/components/content/EditContent.spec.ts
+++ b/cms/src/components/content/EditContent.spec.ts
@@ -1170,4 +1170,119 @@ describe("EditContent.vue", () => {
             });
         });
     });
+
+    describe("dirty state on load", () => {
+        // Regression tests for a bug where MediaEditor / ImageEditor auto-selected the
+        // single available storage bucket on mount and wrote it only to editableParent,
+        // which made the diff against existingParent flag a phantom dirty state every
+        // time a legacy doc (no bucket IDs persisted) was opened.
+
+        const loadWithoutUserEdits = async (postOverride: Partial<PostDto> = {}) => {
+            await db.docs.clear();
+            await db.localChanges.clear();
+
+            const legacyPost: PostDto = {
+                ...mockData.mockPostDto,
+                ...postOverride,
+            };
+            delete (legacyPost as Partial<PostDto>).mediaBucketId;
+            delete (legacyPost as Partial<PostDto>).imageBucketId;
+
+            await db.docs.bulkPut([legacyPost]);
+            await db.docs.bulkPut([
+                mockData.mockEnglishContentDto,
+                mockData.mockFrenchContentDto,
+                mockData.mockSwahiliContentDto,
+            ]);
+            await db.docs.bulkPut([
+                mockData.mockLanguageDtoEng,
+                mockData.mockLanguageDtoFra,
+                mockData.mockLanguageDtoSwa,
+            ]);
+
+            const wrapper = mount(EditContent, {
+                props: {
+                    docType: DocType.Post,
+                    id: legacyPost._id,
+                    languageCode: "eng",
+                    tagOrPostType: PostType.Blog,
+                },
+            });
+
+            await waitForExpect(() => {
+                expect(wrapper.find('input[name="title"]').exists()).toBe(true);
+            });
+            // Let storage-bucket live query resolve and any post-load watchers settle.
+            await wait(50);
+            return wrapper;
+        };
+
+        it("is not dirty on load when the post has no bucket IDs and a single media bucket is auto-selected", async () => {
+            await db.docs.bulkPut([mockData.mockStorageDtoWithEncryptedCredentials]);
+
+            const wrapper = await loadWithoutUserEdits();
+
+            // Revert action is only rendered while isDirty is true — its absence proves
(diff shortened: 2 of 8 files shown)


cms/src/components/content/EditContent.vue
```diff
@@ -149,6 +150,34 @@ const loadData = (id: string, isNew: boolean) => {
...
+        if (imageBucketId && !existingParent.value?.imageBucketId) {
+            existingParent.value = { ...existingParent.value!, imageBucketId };
+            if (editableParent.value && !editableParent.value.imageBucketId) {
+                editableParent.value.imageBucketId = imageBucketId;
+            }
+        }
+    },
+    { immediate: true },
+);
+
```
ivanslabbert: I think we need to find a simpler solution here. Perhaps set the bucket only on new post / tag creation, and thereafter it is set for the post / tag.

ivanslabbert (commented):
failing tests?

cms/src/components/images/ImageEditor.vue
johan-bell, line 59: `acceptedmimeTypes` / `fileTypeDescription` use effectiveImageBucketId (lines 63–89), but `bucketBaseUrl` still only reads `parent.value?.imageBucketId`. For a legacy post with no ID and exactly one image bucket, the effective bucket is implied but `bucketBaseUrl` stays empty, so `existingImagesBucketUrl` / `newUploadsBucketUrl` (lines 44–53) can disagree with MIME rules. Consider resolving the bucket id the same way (e.g. `effectiveImageBucketId` or equivalent) when building `bucketBaseUrl`.

In others terms, The editor now treats “the only image bucket” as chosen in the UI without saving it to the post until the user does something. The code that builds the image URL still only looks at what’s saved on the post. So in edge cases, file-type rules and image URLs might not match. Worth double-checking that previews and links still look right for old posts with no bucket saved.

cms/src/components/content/EditContent.spec.ts
```diff
@@ -1169,4 +1169,117 @@ describe("EditContent.vue", () => {
...
+                    languageCode: "eng",
+                    tagOrPostType: PostType.Blog,
+                },
+            });
+
+            await waitForExpect(() => {
+                expect(wrapper.find('input[name="title"]').exists()).toBe(true);
+            });
+            // Let storage-bucket live query resolve and any post-load watchers settle.
+            await wait(50);
```
johan-bell, line 1214: Fixed wait(50) can flake in CI. Prefer waitForExpect on a stable signal instead of a magic delay ? But if it pass then it is correct.

cms/src/components/media/MediaEditor.vue
johan-bell, line 414: Only relevant if you ever change how multi-bucket selection works. Fine to skip unless someone asks.

cms/src/components/images/ImageEditor.vue
johan-bell, line 307: Only relevant if you ever change how multi-bucket selection works. Fine to skip unless someone asks.

cms/src/components/images/ImageEditor.vue
```diff
@@ -159,20 +162,26 @@ const handleFiles = (files: FileList | null) => {
...
+    // legitimate. Until this point the parent was untouched.
+    if (!effectiveImageBucketId.value) {
         failureMessage.value = "Please select a storage bucket before uploading images.";
         showFailureMessage.value = true;
         return;
     }
+    if (parent.value && !parent.value.imageBucketId) {
+        parent.value.imageBucketId = effectiveImageBucketId.value;
+        emit("bucketSelected", effectiveImageBucketId.value);
+    }
```
johan-bell, line 175: Before, picking the only bucket happened automatically on load and fired an event. Now that only happens when the user uploads or picks a bucket in the dropdown. 

If nothing else in the app was relying on that first automatic event, you’re fine; if something was, please mention it in the PR so no one gets surprised.
(Same idea in MediaEditor, around 268–270.)

cms/package.json
```diff
@@ -34,7 +34,7 @@
         "luminary-shared": "file:../shared",
         "luxon": "^3.4.4",
         "pinia": "^2.1.7",
-        "rte-vue": "^1.1.6",
+        "rte-vue": "^1.1.11",
```
johan-bell: Bumping the rich-text package is easy to miss next to the bucket changes. is it intended or can be removed ? Otherwise, note what the rte-vue update is for.

johan-bell (commented):
Looks good overall —  Maybe align image URL logic with the new “effective” bucket so previews stay consistent. Please fix failing tests and resolve conflicts as well

cms/src/components/media/MediaEditor.vue
MrDirkelz: Not sure what you mean here? Did you comment on the wrong code? As this was not modified? 😅🤔

cms/src/components/content/EditContent.spec.ts
MrDirkelz: It is just a quick wait for the watchers to come to a halt. There is no solid way to check that specifically in unit tests, and this is just a helper function to ensure the tests are actually acurate for dirty checking.

cms/package.json
MrDirkelz: Rte-vue was updated to remove now unneeded functionality.

cms/src/components/images/ImageEditor.vue
MrDirkelz: the automatic on-load bucketSelected emit is preserved (see the watchEffect block a few lines above), and no current consumer acts on the event anyway — so nothing to flag in the PR description. Happy to tighten the code comment around that watchEffect if it would make the intent more obvious to future reviewers.

But please pull this locally and test.

cms/src/components/images/ImageEditor.vue
johan-bell: MediaEditor no longer writes mediaBucketId to parent in watchEffect when a single bucket auto-resolves; only effectiveMediaBucketId + handleFiles persist. This block still mutates parent.imageBucketId on mount, which is the original “phantom dirty” pattern for legacy posts. Please remove this auto-select branch (or gate it so it never touches parent until upload / explicit selection) so Image matches Media and the EditContent dirty tests stay honest.

johan-bell, line 58: Now that “current bucket” can be effective (auto-select) without persisting to parent, double-check that bucketBaseUrl / newUploadsBucketUrl still resolve to the right public URL when only effectiveImageBucketId is set. If the UI shows the wrong bucket base until first upload, consider using effectiveImageBucketId here (with the same fallbacks as elsewhere).

johan-bell (approved):

PR #1509, opened 2026-04-14
fix: prevent app modal from closing on drag-out from input

### LINKED ISSUE #1506: APP: Modal clicking and dragging over input causes modal to close
When clicking and dragging from a modal to the backdrop it triggers the close event. See fix in cms for modals. "Mousedown" event should be used rather than "click"

MrDirkelz:
Collapse BaseModal's backdrop and flex-centering wrapper into a single div so @mousedown.self reliably ignores events originating inside the modal content, matching the CMS LModal structure.

## Summary
- Collapse BaseModal's two sibling `fixed inset-0` divs into a single backdrop + flex-centering wrapper, mirroring the CMS LModal structure
- Fixes the modal closing when a user mousedowns on an input (or other modal content) and drags outside the modal bounds before releasing

## Test plan
- [x] Open any app modal containing an input, click inside the input, drag the cursor outside the modal, release — modal stays open
- [x] Click directly on the backdrop — modal closes
- [X] Existing LModal.spec.ts tests pass

### DIFF
--- a/app/src/components/form/BaseModal.vue
+++ b/app/src/components/form/BaseModal.vue
@@ -7,19 +7,12 @@ const emit = defineEmits(["close"]);
 
 <template>
     <LTeleport v-if="isVisible">
-        <div class="relative z-[100]">
-            <div
-                class="fixed inset-0 bg-zinc-800 bg-opacity-50 backdrop-blur-sm dark:bg-slate-800 dark:bg-opacity-50"
-                @mousedown.self="emit('close')"
-                data-test="modal-backdrop"
-            ></div>
-            <div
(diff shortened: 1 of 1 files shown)


johan-bell (commented):
Do you show us a component where this fix actually applied ?

MrDirkelz:
> Do you show us a component where this fix actually applied ?

This is just the same fix as in the cms, when clicking and dragging over a modal it won't close now. Please refer to PR description as there is a test plan there that you can follow as well 😁

ChrisTouo (commented):
👍

ivanslabbert (approved):

PR #1511, opened 2026-04-15
Improve the sidebar logic

### LINKED ISSUE #1508: CMS: Simplify mobile sidebar logic
We should perhaps create a ticket to simplify this in a helper function later - otherwise if we start to add more pages this will be one massive if clause, and long if clauses create performance issues once they become bigger (mainly on nested if statements, but same can count for these types

_Originally posted by @MrDirkelz in https://github.com/bccsa/luminary/pull/1449#discussion_r3073814656_

ChrisTouo:


cms/src/components/BasePage.vue
```diff
@@ -33,6 +33,12 @@ const isEditContentPage = router.currentRoute.value.name === "edit";
 const isEditLanguagePage = router.currentRoute.value.name === "language";
 const breakpoints = useBreakpoints(breakpointsTailwind);
 const isMobileScreen = breakpoints.smaller("lg");
+
+const handleMobileSidebarToggle = () => {
+    if (isEditContentPage) router.push({ name: "overview" });
+    else if (isEditLanguagePage) router.push({ name: "languages" });
+    else props.onOpenMobileSidebar?.();
+};
 </script>
```
MrDirkelz, line 42: Claude reviewed code I agree with:

Non-reactive route detection (pre-existing, but now more prominent)

// BasePage.vue:32-33
const isEditContentPage = router.currentRoute.value.name === "edit";
const isEditLanguagePage = router.currentRoute.value.name === "language";
These are plain boolean values evaluated once at component setup. They are not computed refs. If BasePage stays mounted across a route change (e.g. via <keep-alive> or a layout that persists), handleMobileSidebarToggle will act on stale values and route to the wrong target.

This was already present in the template before, but now that it lives inside a named function in <script setup>, the risk is more visible. Even if the current routing architecture avoids the problem today, making the intent explicit costs nothing:

// Prefer computed so the values stay fresh if the route changes
const isEditContentPage = computed(() => router.currentRoute.value.name === "edit");
const isEditLanguagePage = computed(() => router.currentRoute.value.name === "language");
Then update the function to read .value:

const handleMobileSidebarToggle = () => {
    if (isEditContentPage.value) router.push({ name: "overview" });
    else if (isEditLanguagePage.value) router.push({ name: "languages" });
    else props.onOpenMobileSidebar?.();
};

MrDirkelz (dismissed):
3. Missing test coverage.

There are no new tests in this PR:

BasePage.spec.ts should cover handleMobileSidebarToggle: verify that clicking the chevron on an edit-content page calls router.push({ name: 'overview' }), on a language page calls router.push({ name: 'languages' }), and otherwise calls onOpenMobileSidebar.
MobileSideBar.spec.ts should verify that when SideBar emits close, the update:open event is emitted with false — essentially testing the bug that was just fixed.

johan-bell (approved):

PR #1513, opened 2026-04-16
CMS: enhance layout of ACL entry and group components for better responsiveness

### LINKED ISSUE #1510: CMS: Improve the layout responsiveness of Acl entry

ChrisTouo:


cms/src/components/groups/EditAclByGroup.vue
```diff
@@ -180,7 +180,7 @@ onMounted(() => {
             <div v-if="typesWithActivePermissions.length === 0" class="text-xs">
                 No active permissions, use the selector to add
             </div>
-            <div class="mb-3">
+            <div class="mb-3 grid grid-cols-[max-content_1fr_max-content] items-center">
```
MrDirkelz: items-center on the grid conflicts with the h-full cell design

The label and action cells (EditAclEntry.vue:53, EditAclEntry.vue:70) use h-full so their border-b spans the full row height. But items-center (align-items: center) on the grid causes those cells to be vertically centered within the row, not stretched. The h-full height resolves to the row track height regardless, so visually it may appear to work — but only because h-full effectively overrides the centering. This is confusing and fragile.

Recommendation: Remove items-center from the grid container and rely on the default align-items: stretch. Each cell that needs its content centered vertically already handles that itself via its own flex items-center classes. The middle permissions cell (which has no h-full) will stretch to fill the row height, which is the correct behaviour for a border-accented table.

johan-bell:
- [x] Do you think it is better when it bleeds to the edges on mobile ? 

image

- [x] I think we can move this button somewhere else as it does the same thing in each group that have access to the current group. It seems to be the same across all the group (check second and third images...)

image

image
image

- [x] May be it would more nicer here if we remove the paddming and marging here so that the text stay also aligned with the rest ? 

image

- [x] remove/disable the PlusIcon where there is no group available instead of showing the input field ? Unless there is a specific reason for that.

image
image

- [x] "**No active persmissions, use the selector to add**" can be centerered a bit bigger when there is no persmission ? 

image

ChrisTouo:
This is how it look now


image

ivanslabbert (approved):

PR #1521, opened 2026-04-17
CMS: change LSelect to have similar styling to LCombobox - 1515

### LINKED ISSUE #1515: CMS: Change LSelect to have simiar styling to LCombobox
Image
Currently uses native dropdown which does not fit the style of the CMS

johan-bell:


cms/src/components/common/LPaginator.vue
```diff
@@ -149,7 +149,7 @@ watch(openPageSizeSelect, (newValue) => {
                 />
                 <span v-if="variant === 'simple'" class="text-sm text-zinc-600">
                     Page <strong>{{ index + 1 }}</strong> of
-                    <strong>{{ paginatorPages.length + 1 }}</strong>
```
MrDirkelz, line 152: Do not think this is part of this ticket's scope? Or is this a fix for a echo affect coming from changing LSelect?

cms/src/components/content/ContentOverview/ContentOverview.spec.ts
```diff
@@ -454,30 +455,30 @@ describe("ContentOverview.vue", () => {
         //@ts-ignore as this code is valid
         wrapper.vm.selectedLanguage = "lang-eng";
 
-        const filterInputSelects = wrapper.findAll('[data-test="filter-select"]');
+        const filterSelects = wrapper.findAllComponents(LSelect);
```
MrDirkelz, line 459:
```suggestion
        const filterSelectInputs = wrapper.findAllComponents(LSelect);
```

MrDirkelz (approved):
Just 1 comment and a suggestion on naming, also changed the title of this PR so it displays nicer in the git commit history, but past that looks good to me.

PR #1551, opened 2026-04-21
perf(fts): speed up sync writes and search query latency

### LINKED ISSUE #1549: Shared: Increase FTS Search performance

MrDirkelz:
# FTS search perf comparison: branch `1549` vs `main`

Local IndexedDB, instrumented `shared/src/fts/ftsSearch.ts` with identical phase timers and ran the same queries against each algorithm. Numbers in ms.

Three variants measured:

1. **main** — the algorithm on `main`. No trigram inverted index, no candidate window, no in-memory caches. Scores every matched doc.
2. **branch+caches** — branch `1549` with `trigramPostings`, coarse-rank candidate window, plus in-memory `docCache` and `wordSetCache`.
3. **branch no-caches** — the final shipped version: same persistent-IDB optimizations as variant 2 but with both in-memory caches removed (per senior directive: no process-memory caching).

## Per-query totals

| query | main | branch+caches | branch no-caches |
|---|---:|---:|---:|
| "What is the benefits to self-denial" | 1188 | 465 | 767 |
| "How to overcome habitual sin" | 2501 | 407 | — |
| "How we can overcome sin" | 1054 | — | 2314 ⚠ |
| "14 Bible verses…overcome sin" | — | — | 1299 |

⚠ "How we can overcome sin" on no-caches hit a sync-contention spike: 2188 ms to load just 60 docs (36 ms/doc vs the normal 5–10 ms). Treat as variance, not algorithmic regression — rerun for a fair picture.

## Phase breakdown — *"What is the benefits to self-denial"*

15 query trigrams, 10 usable, ~442 matched docs in the corpus.

| metric | main (slow) | branch+caches | branch no-caches |
|---|---:|---:|---:|
| **total** | **1188** | **465** | **767** |
| candidates scored | 442 | 60 | 60 |
| corpusStats | 19 | 18 | 23 |
| trigramCount (main only) | 52 | — | — |
| trigramLookup | 15 | 16 | 8 |
| coarseRank | — | 2 | 2 |
| docsLoad | 527 | 343 | 639 |
| bm25 | 94 | 15 | 17 |
| wordMatch | 481 | 71 | 78 |
| sort | 1.3 | 0 | 0.4 |

## What's driving the win (preserved in no-caches version)

| change on the branch | impact | preserved without caches? |
|---|---|---|
| **Coarse-rank candidate window** (`Σ idf·tf` from postings → top `offset+limit+20` docs) | Dominant. Cuts 442 → 60 docs that get full BM25 + word-match. Multiplies through every later phase. | ✅ Yes |
| **Trigram inverted index** (`trigramPostings` table, single `bulkGet`) | Replaces main's two sequential per-trigram passes (`count()` + `primaryKeys()`). Saves the ~50 ms `trigramCount` phase outright. | ✅ Yes |
| **Per-store implicit transactions** instead of one big readonly tx | Smaller scopes get granted faster under sync write contention. | ✅ Yes |
| **Compound `[language+_id]` index** on docs | Wrong-language candidates miss at the IDB layer — no row read for docs we'd discard. | ✅ Yes |
| **Doc cache** (LRU 500, `ContentDto` by `_id`) | Skipped IDB hydration for overlapping candidate sets during typing/pagination. | ❌ Removed |
| **Word-set cache** (LRU 500, per-doc normalized field word sets) | Skipped repeat `stripHtml` + `normalizeText` work for docs seen in adjacent queries. | ❌ Removed |

## Per-doc scoring cost is identical, the win is in N

| | main | branch+caches | branch no-caches |
|---|---:|---:|---:|
| wordMatch ms / doc | 1.09 | 1.18 | 1.30 |
| bm25 ms / doc | 0.21 | 0.25 | 0.28 |

The branch isn't faster *per doc* — it just does the work on far fewer docs.

## Cost of removing the in-memory caches

| flow | regression |
|---|---|
| **Cold first search** | **None.** Both caches start empty anyway; numbers are within run-to-run variance of the cached version. |
| **Repeated overlapping queries** (interactive typing, pagination) | Theoretical ~30–200 ms slower per follow-up keystroke. Not measured in either trace set — would need a separate test with repeated queries to quantify. |
| **Sync-contention bursts** | `docCache` would have damped some of the `docsLoad` variance. Without it, slow-run spikes are larger. |

## Recall tradeoff (and why it's fine)

Coarse rank doesn't account for word-match field boosts (e.g. title at 3.0×), so in theory a low-coarse-score doc with a strong title match could be excluded by the candidate window.

In practice this isn't an issue because:

- A title match means the title's trigrams hit, which drives high coarse score anyway — the doc is already in the window.
- Empirically: title-match queries return the expected doc on the branch.
- **Pagination expands the window** — `slice(0, offset + limit + 20)` grows with offset, so as the user pages forward, more candidates become reachable. A doc that "should have been" in the top 20 by full BM25 but missed the coarse cut will surface within a page or two.

## What's still on the table

- **`docsLoad` is the remaining bottleneck** (~75% of branch latency, baseline ~80 ms uncontended, much worse during sync). Avoiding `db.docs` at search time entirely (precomputed display data in a smaller store) would be the next biggest win.
- **Web Worker** wouldn't reduce work but would stop main-thread sync from causing the 6× slow-run amplification.
- **Remove the `*fts` multi-entry index from `db.docs`** — its only runtime consumer is the fallback path. Removal cuts IDB storage materially and reduces sync write amplification. Backfill iterates docs directly, so it doesn't need the index either.

### DIFF
--- a/app/src/pages/HomePage.vue
+++ b/app/src/pages/HomePage.vue
@@ -2,6 +2,7 @@
 import IgnorePagePadding from "@/components/IgnorePagePadding.vue";
 import HomePagePinned from "@/components/HomePage/HomePagePinned.vue";
 import HomePageNewest from "@/components/HomePage/HomePageNewest.vue";
+import HomePageSearch from "@/components/HomePage/HomePageSearch.vue";
 import BasePage from "@/components/BasePage.vue";
 import ContinueWatching from "@/components/HomePage/ContinueWatching.vue";
 import ContinueListening from "@/components/HomePage/ContinueListening.vue";
@@ -10,6 +11,7 @@ import ContinueListening from "@/components/HomePage/ContinueListening.vue";
 <template>
     <BasePage>
         <IgnorePagePadding ignoreTop>
+            <HomePageSearch />
             <Suspense>
                 <HomePagePinned />
             </Suspense>
--- a/app/src/pages/HomePage.vue
+++ b/app/src/pages/HomePage.vue
@@ -2,7 +2,6 @@
 import IgnorePagePadding from "@/components/IgnorePagePadding.vue";
 import HomePagePinned from "@/components/HomePage/HomePagePinned.vue";
 import HomePageNewest from "@/components/HomePage/HomePageNewest.vue";
-import HomePageSearch from "@/components/HomePage/HomePageSearch.vue";
 import BasePage from "@/components/BasePage.vue";
 import ContinueWatching from "@/components/HomePage/ContinueWatching.vue";
 import ContinueListening from "@/components/HomePage/ContinueListening.vue";
--- a/shared/src/fts/ftsSearch.ts
+++ b/shared/src/fts/ftsSearch.ts
@@ -23,6 +23,40 @@ const FTS_FIELDS: FtsFieldConfig[] = [
     { name: "author", boost: 1.0 },
 ];
 
+// LRU cache of per-doc normalized word sets.
+// stripHtml + normalizeText dominate word-match scoring cost; they depend only on
+// the doc's field content, so cache across queries and invalidate on edit via
+// `updatedTimeUtc` in the key.
+const WORD_SET_CACHE_MAX = 500;
+const wordSetCache = new Map<string, Record<string, Set<string>>>();
+
+function getWordSets(
+    doc: ContentDto,
+    fields: FtsFieldConfig[],
+): Record<string, Set<string>> {
+    const key = `${doc._id}:${doc.updatedTimeUtc}`;
+    const cached = wordSetCache.get(key);
+    if (cached) {
+        wordSetCache.delete(key);
+        wordSetCache.set(key, cached);
+        return cached;
+    }
+    const sets: Record<string, Set<string>> = {};
+    const docRec = doc as Record<string, any>;
+    for (const field of fields) {
+        const value = docRec[field.name];
+        if (typeof value !== "string" || !value) continue;
+        const text = normalizeText(field.isHtml ? stripHtml(value) : value);
+        sets[field.name] = new Set(text.split(" "));
+    }
+    wordSetCache.set(key, sets);
+    if (wordSetCache.size > WORD_SET_CACHE_MAX) {
+        const oldest = wordSetCache.keys().next().value;
+        if (oldest !== undefined) wordSetCache.delete(oldest);
+    }
+    return sets;
+}
+
 /**
  * Compute a boost-weighted word match score across fields.
  * For each field, counts how many query words appear as full words,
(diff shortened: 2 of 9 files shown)


johan-bell (dismissed):
Do we really need to keep DEV flag ? Else LGTM

shared/src/fts/ftsSearch.ts
```diff
@@ -75,16 +104,23 @@ export async function ftsSearch(options: FtsSearchOptions): Promise<FtsSearchRes
...
 
-    // Step 1: Count docs per trigram, filter over-represented
+    // Step 1 (merged from former step1+step3): fetch primaryKeys per trigram in parallel.
+    // `df` is derived from `ids.length`, so the previous separate .count() pass is gone.
+    const trigramResults = await Promise.all(
+        trigrams.map(async (trigram) => {
+            const ids = (await db.docs
+                .where("fts")
+                .between(trigram + ":", trigram + ";", true, false)
+                .primaryKeys()) as string[];
```
johan-bell: We fetch `primaryKeys()` for every trigram up front. For very common trigrams this can materialize a large key set before we even know it is usable.

shared/src/fts/ftsSearch.ts
```diff
@@ -75,16 +104,23 @@ export async function ftsSearch(options: FtsSearchOptions): Promise<FtsSearchRes
...
-        const count = await db.docs
-            .where("fts")
-            .between(trigram + ":", trigram + ";", true, false)
-            .count();
-        if (count <= maxDocCount) {
-            usableTrigrams.push({ token: trigram, df: count });
-        }
+    const matchedDocIds = new Set<string>();
+    for (const { token, ids } of trigramResults) {
+        if (ids.length > maxDocCount) continue;
```
johan-bell: The maxDocCount guard happens after key materialization, so oversized trigram matches are still fully loaded and then discarded.

shared/src/fts/ftsSearch.ts
```diff
@@ -75,16 +104,23 @@ export async function ftsSearch(options: FtsSearchOptions): Promise<FtsSearchRes
...
+        usableTrigrams.push({ token, df: ids.length });
+        for (const id of ids) matchedDocIds.add(id);
```
johan-bell: Only usable trigrams should reach this accumulation path; consider a `count()` pre-pass (or capped scan) so we avoid loading IDs for trigrams that fail the threshold.

johan-bell (commented):
We now call primaryKeys() before applying maxDocCount. For high-frequency trigrams, this loads a large ID list that is immediately discarded by if (ids.length > maxDocCount) continue;. Could we do a count() pre-pass and only fetch keys for usable trigrams to avoid extra IDB work and memory pressure?

shared/src/fts/ftsSearch.ts
ivanslabbert, line 1: Investigate if we can fetch all the trigram matches from the DB in one query instead of iterating through the trigrams and doing a DB lookup for each trigram. That will potentially make the search a lot faster

app/src/pages/HomePage.vue
```diff
@@ -2,6 +2,7 @@
 import IgnorePagePadding from "@/components/IgnorePagePadding.vue";
 import HomePagePinned from "@/components/HomePage/HomePagePinned.vue";
 import HomePageNewest from "@/components/HomePage/HomePageNewest.vue";
+import HomePageSearch from "@/components/HomePage/HomePageSearch.vue";
```
ivanslabbert: remove

app/src/pages/HomePage.vue
```diff
@@ -10,6 +11,7 @@ import ContinueListening from "@/components/HomePage/ContinueListening.vue";
 <template>
     <BasePage>
         <IgnorePagePadding ignoreTop>
+            <HomePageSearch />
```
ivanslabbert, line 13: remove

PR #1557, opened 2026-04-23
fix(api): clear cached DTO state on DB disconnect

### LINKED ISSUE #1492: API: Ensure that all singleton instances of cached values for most of our dtos are cleared on database disconnection

MrDirkelz:
DbService now emits `disconnect` when the changes feed drops after a successful connect, and `reconnect` once the DB is reachable again. Services that hold in-memory DTO caches listen for these events:

- AuthIdentityService drops providerCache, autoGroupMappingsCache, defaultGroupsCache and jwksClients (all lazy — repopulate on next access).
- QueryService empties its languages cache on disconnect and re-fetches on reconnect.
- PermissionSystem clears the module-level groupMap on disconnect and re-hydrates from getGroups() on reconnect. The resumed changes feed does not replay events missed during downtime, so a full rebuild is needed to avoid serving stale permissions.
- S3Service.initializeChangeListener now also calls clearCache() on disconnect, and is finally wired into main.ts (it was previously only called from tests).

Tests added for the disconnect/reconnect contract on DbService and for each affected service, including lazy repopulation of AuthIdentityService and S3Service caches after a clear.

### DIFF
--- a/api/src/endpoints/query.service.ts
+++ b/api/src/endpoints/query.service.ts
@@ -35,6 +35,19 @@ export class QueryService {
             }
         });
 
+        // Drop the language cache on disconnect; any changes made while the feed
+        // was down would otherwise be missed by the resumed change stream.
+        this.db.on("disconnect", () => {
+            this.languages = [];
+        });
+        this.db.on("reconnect", () => {
+            this.loadLanguages();
+        });
+
+        this.loadLanguages();
+    }
+
+    private loadLanguages() {
         this.db
             .executeFindQuery({
                 selector: { type: DocType.Language },
@@ -42,7 +55,7 @@ export class QueryService {
                 use_index: "sync-language-index",
             })
             .then((res) => {
-                this.languages.push(...(res.docs as LanguageDto[]));
+                this.languages = res.docs as LanguageDto[];
             });
     }
 
--- a/api/src/endpoints/query.service.ts
+++ b/api/src/endpoints/query.service.ts
@@ -55,7 +55,17 @@ export class QueryService {
                 use_index: "sync-language-index",
             })
             .then((res) => {
-                this.languages = res.docs as LanguageDto[];
+                // Merge by _id rather than reassigning or pushing blindly: the
+                // `languageUpdate` change-stream listener may fire during the async
+                // window of this query (especially on reconnect, where the resumed
+                // feed can deliver events before this promise resolves). Reassigning
+                // would clobber those concurrent updates; pushing could create
+                // duplicates for docs the listener already inserted.
+                for (const doc of res.docs as LanguageDto[]) {
+                    const i = this.languages.findIndex((l) => l._id == doc._id);
+                    if (i >= 0) this.languages[i] = doc;
+                    else this.languages.push(doc);
+                }
             });
     }
 
--- a/api/src/endpoints/query.service.ts
+++ b/api/src/endpoints/query.service.ts
@@ -66,6 +66,9 @@ export class QueryService {
                     if (i >= 0) this.languages[i] = doc;
                     else this.languages.push(doc);
                 }
+            })
+            .catch((err) => {
+                this.logger.error("Failed to load languages cache", err);
             });
     }
 
--- a/api/src/permissions/permissions.service.ts
+++ b/api/src/permissions/permissions.service.ts
@@ -104,7 +104,7 @@ export type AccessMap = Map<Uuid, Map<DocType, Map<AclPermission, boolean>>>;
  * Accessed throughout this file via object-style property access, so it is
  * typed as a Record rather than a Map.
  */
-const groupMap: Record<Uuid, PermissionSystem> = {};
+const groupMap: Map<Uuid, PermissionSystem> = new Map<Uuid, PermissionSystem>();
 
 /**
  * Represents a permission system that uses a tree structure to organize groups and manage permissions.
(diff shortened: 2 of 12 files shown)


api/src/endpoints/query.service.ts
```diff
@@ -35,14 +35,27 @@ export class QueryService {
...
+    private loadLanguages() {
         this.db
             .executeFindQuery({
                 selector: { type: DocType.Language },
                 limit: Number.MAX_SAFE_INTEGER,
                 use_index: "sync-language-index",
             })
             .then((res) => {
-                this.languages.push(...(res.docs as LanguageDto[]));
+                this.languages = res.docs as LanguageDto[];
```
ivanslabbert: Just double check this? Not sure why the previous logic used `push`

api/src/permissions/permissions.service.ts
```diff
@@ -153,6 +153,22 @@ export class PermissionSystem extends EventEmitter {
             }
         });
 
+        // Drop the groupMap on disconnect — the change feed resumes without
+        // replaying events missed during downtime, so we must rehydrate from
+        // scratch on reconnect rather than risk serving stale permissions.
+        dbService.on("disconnect", () => {
+            initialized = false;
+            updateQueue = [];
+            PermissionSystem.clearGroupMap();
```
ivanslabbert: We need to handle this differently, as it can cause users connecting to an islanded API to have all data deleted.

On reconnection we can do a diff if possible, and only update changes to the group system

api/src/permissions/permissions.service.ts
```diff
@@ -153,6 +153,38 @@ export class PermissionSystem extends EventEmitter {
...
+        // for existing groups, so we just need to additionally remove groups
+        // that were deleted from the DB while we were disconnected. This keeps
+        // the cache continuously valid and avoids the brief "no permissions"
+        // window a full rebuild would create.
+        dbService.on("reconnect", async () => {
+            const dbGroups = await dbService.getGroups();
+
+            const fetchedIds = new Set(dbGroups.docs.map((d: GroupDto) => d._id));
+            const staleIds = Object.keys(groupMap).filter((id) => !fetchedIds.has(id));
+            if (staleIds.length > 0) PermissionSystem.removeGroups(staleIds);
```
johan-bell: **Small maintainability note**: **groupMap** is typed as **Map<Uuid, PermissionSystem>** but this line uses **Object.keys(groupMap)**, which relies on object-style property access used elsewhere in this file. Consider using a consistent representation (**Record** everywhere or real **Map** APIs like **Array.from(groupMap.keys())**) to avoid surprises during future refactors ??

api/src/endpoints/query.service.ts
```diff
@@ -35,14 +35,37 @@ export class QueryService {
...
+                // window of this query (especially on reconnect, where the resumed
+                // feed can deliver events before this promise resolves). Reassigning
+                // would clobber those concurrent updates; pushing could create
+                // duplicates for docs the listener already inserted.
+                for (const doc of res.docs as LanguageDto[]) {
+                    const i = this.languages.findIndex((l) => l._id == doc._id);
+                    if (i >= 0) this.languages[i] = doc;
+                    else this.languages.push(doc);
+                }
             });
```
johan-bell, line 72: **Suggestion**: consider adding a **.catch(...)** (or converting **loadLanguages()** to **async/await** with try/catch) so a failed **executeFindQuery** on reconnect is logged explicitly. Right now failures are effectively silent, which can make cache-rehydration issues harder to diagnose in production.

johan-bell (commented):
LGTM, just two small comments left.

api/src/endpoints/query.service.ts
ivanslabbert: Only catching at top level is a good design by default, where the top level caught errors are logged and reported back to the user (nestjs default)

api/src/permissions/permissions.service.ts
```diff
@@ -100,9 +100,11 @@ type TargetRouteTypeMap = Map<
 export type AccessMap = Map<Uuid, Map<DocType, Map<AclPermission, boolean>>>;
 
 /**
- * Global Group Map used for permission lookups
+ * Global Group Map used for permission lookups.
+ * Accessed throughout this file via object-style property access, so it is
+ * typed as a Record rather than a Map.
  */
-const groupMap: Map<Uuid, PermissionSystem> = new Map<Uuid, PermissionSystem>();
+const groupMap: Record<Uuid, PermissionSystem> = {};
```
ivanslabbert: Revert

api/src/permissions/permissions.service.ts
```diff
@@ -153,6 +155,38 @@ export class PermissionSystem extends EventEmitter {
...
+            const fetchedIds = new Set(dbGroups.docs.map((d: GroupDto) => d._id));
+            const staleIds = Object.keys(groupMap).filter((id) => !fetchedIds.has(id));
+            if (staleIds.length > 0) PermissionSystem.removeGroups(staleIds);
+
+            PermissionSystem.upsertGroups(dbGroups.docs);
+            PermissionSystem.upsertGroups(updateQueue);
+            updateQueue = [];
+            initialized = true;
+        });
+
```
ivanslabbert: Re-think this. Could be we should use timestamps (updatedTimeUtc) to determine which changes happened after the API got disconnected from the database server

api/src/permissions/permissions.service.ts
```diff
@@ -144,7 +146,12 @@ export class PermissionSystem extends EventEmitter {
         let initialized = false;
         let updateQueue: any[] = [];
 
-        // Read group changes from the database change feed and update the permission system
+        // Read group changes from the database change feed and update the permission system.
+        // DbService resumes the feed from its last seq cursor on reconnect, so updates made
+        // while disconnected flow through here once we're back online — no snapshot refetch
+        // needed. The in-memory groupMap is intentionally kept intact across disconnects:
+        // clearing it would deny every request and cause clients syncing via this API to
+        // purge their local data.
```
ivanslabbert: Not sure we should describe  in such detail what the dbservice does here. It becomes difficult to maintain if we would make changes in the dbservice

ivanslabbert (approved):

PR #1562, opened 2026-04-24
fix: LCombobox inline tag removal and auto-group mapping discard - 1555

### LINKED ISSUE #1555: CMS: Auto Group Mapping bugs
- Unable to remove inline tags from LCombobox input -> Seems to be event propogation from the dropdown opening.
- Sometimes data not being correctly discarded when making a change and then trying to discard it.

MrDirkelz:
- LDropdown trigger used @click.capture.stop which swallowed clicks to nested interactive elements (e.g. LTag X button). Switched to @click.stop and added .stop on LTag's remove button.
- CreateOrEditAutoGroupMappingModal cloned props.mapping via spread but overrode groupIds/conditions with the original reactive arrays, so edits mutated the parent mapping and survived discard. Now reads defaults off the deep-cloned value in both the watcher and revert().

### DIFF
--- a/cms/src/components/autoGroupMappings/CreateOrEditAutoGroupMappingModal.vue
+++ b/cms/src/components/autoGroupMappings/CreateOrEditAutoGroupMappingModal.vue
@@ -64,13 +64,16 @@ watch(
     () => props.mapping,
     (existing) => {
         if (existing) {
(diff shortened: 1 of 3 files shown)


johan-bell (approved):
Please consider rebasing to main before merging.

ivanslabbert (approved):

PR #1569, opened 2026-04-27
1568 cms api duplicate image on post duplication

### LINKED ISSUE #1567: CMS / API: Prompt to ask user to duplicate image on post duplication
**Is your feature request related to a problem? Please describe.**
Some posts have an image that should be re-used on duplicated posts.

**Describe the solution you'd like**
User should be prompted to consent that image is duplicated.

**Describe alternatives you've considered**

**Additional context**
Images should not be re-encoded, but perhaps all renditions should be pointed to, and upon saving copied to new S3 image files

### LINKED ISSUE #1568: CMS, API: duplicate image on post duplication
Prompt user if he/she wants to duplicate image.
Need to think how image duplication will work without re-encoding image, while keeping offline editing support etc.

johan-bell:


MrDirkelz (commented):
Quality seems to be kept when duplicating locally 

Some comments: 

- I think we have an LRadio component in the cms you should be able to use here and just make it have a variant that still looks the same but just acts as a checkbox? 
Image

- Images sometimes fails to duplicate when duplicating even when the duplicate image checkbox is ticked.
Image

johan-bell:
@MrDirkelz :

- The `LRadio` suggestion doesn’t fit this case. `LRadio` is a true `type="radio"` component in cms, not a checkbox-style control, so swapping the “Duplicate image” checkbox to that component would be semantically 
wrong unless we build a dedicated checkbox variant/component.

Quality is preserved in the current duplication flow.

- In cms, duplication sends existing imageData.fileCollections plus duplicateFrom.
- In api, processImageDto handles that through duplicateImageFilesWithoutReencoding(...).
- That function copies original file bytes from storage and uploads them as-is (same content type), only generating new filenames.
- It does not run Sharp/webp conversion in that path, so no recompression or quality loss is introduced by duplication itself.

MrDirkelz:
> @MrDirkelz :
> 
> - The `LRadio` suggestion doesn’t fit this case. `LRadio` is a true `type="radio"` component in cms, not a checkbox-style control, so swapping the “Duplicate image” checkbox to that component would be semantically 
> wrong unless we build a dedicated checkbox variant/component.
> 
> Quality is preserved in the current duplication flow.
> 
> - In cms, duplication sends existing imageData.fileCollections plus duplicateFrom.
> - In api, processImageDto handles that through duplicateImageFilesWithoutReencoding(...).
> - That function copies original file bytes from storage and uploads them as-is (same content type), only generating new filenames.
> - It does not run Sharp/webp conversion in that path, so no recompression or quality loss is introduced by duplication itself.

Think I menioned using a variant in my review for that checkbox

api/src/dto/ImageDto.ts
```diff
@@ -19,4 +20,10 @@ export class ImageDto {
     @Type(() => ImageUploadDto) // This throws an exception on validation failure, so we need to catch the error on validation. The message is less user-friendly but at least the validator fails and will protect our data.
     @Expose()
     uploadData?: ImageUploadDto[];
+
+    @IsOptional()
+    @ValidateNested()
+    @Type(() => ImageDuplicateFromDto)
+    @Expose()
+    duplicateFrom?: ImageDuplicateFromDto;
```
ivanslabbert: Change to boolean "duplicate"
When set, processImageDto does the binary duplication, and replaces the file collection with the new file references (to the duplicated images)

If duplication would fail for some or other reason, the original fileCollections should be removed

cms/src/components/content/EditContent.vue
ivanslabbert, line 1: If the "duplicate" flag is set on the image field, do not remove the existing image. The API (processImageDto) will take care of the image duplication.

api/src/changeRequests/documentProcessing/processImageDto.ts
```diff
@@ -257,13 +280,69 @@ export async function processImage(
...
+
+        delete image.duplicate;
     } catch (error) {
         warnings.push(`Image processing failed: ${error.message}`);
     }
 
     return { migrationFailed, warnings };
 }
 
+async function duplicateImageFilesWithoutReencoding(
```
ivanslabbert, line 292: check if this function can use the S3 CP command instead of downloading and uploading the assets

MrDirkelz:
image
https://github.com/minio/minio-js/blob/master/examples/copy-object.mjs

api/src/changeRequests/documentProcessing/processImageDto.ts
johan-bell: Good call — `copyObject` would work here when both buckets share the same endpoint, but since each bucket can have its own S3 endpoint we'd also need a fallback for the cross-server case, which adds complexity. On top of that, this is an infrequent editor-triggered operation on already-small processed WebPs, so the value it brings is limited.

ivanslabbert (approved):

PR #1573, opened 2026-04-28
Feat: Hide privacy policy with query param

### LINKED ISSUE #1572: APP: Add query param to hide privacy policy modal for SSG purposes

MrDirkelz:
Allow hiding the privacy policy modal and banner by appending `?noprivacypolicy=true` to the URL. This is useful for testing or specific scenarios where the policy should not be displayed.


johan-bell (commented):
Tests failing. Isn't better to have an env variable instead of a URL param ?

ivanslabbert (approved):

PR #1574, opened 2026-04-28
Refactor: Replace LoadingSpinner with LoadingBar

### LINKED ISSUE #1514: CMS: Review loading state on all overview pages and use loading bar (uniform UI design)

MrDirkelz:
The `LoadingSpinner` component has been replaced with a more visually engaging `LoadingBar` component across various views. This change aims to provide a more consistent and modern loading experience for the user. Additionally, the `LoadingSpinner.vue` file has been removed.

https://github.com/user-attachments/assets/a2caaa7c-06d6-47df-ac8e-efbfc101dbc1


MrDirkelz (commented):
Good effort on the visual redesign — the animated loading bar is a nice UX upgrade over the spinner. The `BasePage` centralisation of loading state is the right architectural direction. That said, there are a few issues that need attention before merge.

---

## 🚨 Critical — Potential stuck loading state in `LanguageOverview` and `RedirectOverview`

Both files use this pattern:

```ts
const isLoading = ref(true);
const stopLoadingWatcher = watch(languages, () => {
    isLoading.value = false;
    stopLoadingWatcher();
});
```

`watch` is **lazy by default** — it only fires when the ref *changes* after it is registered. If `db.whereTypeAsRef` is backed by a local cache (e.g. IndexedDB), the ref may already hold its final value at the time the component mounts. In that case the watcher never fires, `isLoading` stays `true` forever, and the user sees an infinite loading bar instead of their data.

**Fix option 1 — `{ once: true }` (Vue 3.4+, cleanest):**
```ts
const isLoading = ref(true);
watch(languages, () => { isLoading.value = false; }, { once: true });
```

**Fix option 2 — `{ immediate: true }` with a guard:**
```ts
const isLoading = ref(true);
watch(languages, () => { isLoading.value = false; }, { immediate: true, once: true });
```

**Fix option 3 — align with the other overviews.** `AutoGroupMappingOverview` and `UserOverview` derive loading state from the query object itself (`mappingQuery.isLoading.value`). If `db.whereTypeAsRef` has an equivalent loading flag, use it for consistency.

---

## ⚠️ Warning — `LoadingBar` lacks a proper ARIA live region

```html
<div class="flex w-full flex-col items-center gap-0"
     aria-label="Loading...">
```

`aria-label` on a plain `<div>` does nothing useful for screen readers — they don't announce it. For a loading indicator the accessible pattern is `role="status"` (implicitly `aria-live="polite"`), which causes screen readers to announce the content when it appears in the DOM.

```html
<div role="status" aria-label="Loading..." class="flex w-full flex-col items-center">
```

---

## ⚠️ Warning — Three different loading-state patterns in one PR

Compare the four overview files touched in this PR:

| File | Pattern |
|---|---|
| `AutoGroupMappingOverview` | `computed` from `query.isLoading` ✅ |
| `UserOverview` | `computed` from `query.isLoading` ✅ |
| `LanguageOverview` | `ref(true)` + manual `watch` ⚠️ |
| `RedirectOverview` | `ref(true)` + manual `watch` ⚠️ |

Additionally, `AuthProviderOverview` and `StorageOverview` still use inline `v-if="isLoading && !items.length"` directly in the template rather than going through `BasePage`'s new `loading` prop. Mixing three approaches in a single PR makes the codebase harder to reason about. Pick one pattern and apply it consistently — the `BasePage` loading prop approach is the right one.

---

## 💡 Nitpick — Redundant `gap-0` in `LoadingBar.vue`

`gap-0` is Tailwind's default and adds no value:

```html
<!-- remove gap-0 -->
<div class="flex w-full flex-col items-center">
```

---

## 💡 Nitpick — Empty `class=""` attribute

```html
<span class="">Loading</span>
```

Remove the empty attribute entirely.

---

## 💡 Nitpick — Whitespace hack in `LoadingBar.vue`

The closing-tag trick to suppress inter-dot spacing is fragile and surprising to future readers:

```html
<span ...>.</span
><span ...>.</span
><span ...>.</span>
```

A flex wrapper removes whitespace entirely and is self-documenting:

```html
<div class="flex items-end">
    <span class="animate-bob-1 inline-block text-lg" aria-hidden="true">.</span>
    <span class="animate-bob-2 inline-block text-lg" aria-hidden="true">.</span>
    <span class="animate-bob-3 inline-block text-lg" aria-hidden="true">.</span>
</div>
```

---

## 💡 Nitpick — Overly verbose null-coalescing in `UserOverview`

```ts
// before
() => apiLiveQuery.isLoading.value && !(users.value?.length ?? 0),

// after — equivalent and simpler (both undefined and 0 are falsy when negated)
() => apiLiveQuery.isLoading.value && !users.value?.length,
```

---

The critical issue (potential stuck loading state) needs resolving before merge; the rest are relatively small. The overall direction is solid — once these are addressed this will be a clear improvement.

---

johan-bell (approved):

MrDirkelz (commented):
## Code Review — Refactor: Replace LoadingSpinner with LoadingBar

> **Note:** This PR is already merged — these findings are for learning and to flag any follow-up work needed.

The visual improvement is a nice upgrade, and most of the changes are clean and consistent. The `BasePage`-level loading integration is well-structured. There are a few issues worth calling out, one of which is a production bug.

---

### 🚨 Critical — Infinite loading state for empty or fast-loading datasets

**Files:** `LanguageOverview.vue`, `RedirectOverview.vue`

```js
const isLoading = ref(true);
const stopLoadingWatcher = watch(languages, () => {
    isLoading.value = false;
    stopLoadingWatcher();
});
```

This pattern has two failure modes:

1. **Empty dataset**: `isLoading` starts as `true`. The watcher only fires when the reactive ref *changes*. If there are zero languages/redirects in the database, the ref starts as `[]` and may never receive an update, so `isLoading` stays `true` forever. Users will see an infinite loading bar and the empty-state UI (`v-else-if="!languages.length"`) will never appear.

2. **Synchronous/cached population**: This is an offline-first app backed by a local database. If `db.whereTypeAsRef` populates the ref synchronously (e.g., from IndexedDB cache before Vue's next tick), the ref's value is set before the watcher is registered and the callback never fires.

Compare with the correct patterns elsewhere in this PR:
- `UserOverview.vue` derives loading from `apiLiveQuery.isLoading.value` — a proper loading state from the query object.
- `AutoGroupMappingOverview.vue` does the same with `mappingQuery.isLoading.value`.

**Recommended fix:** Check whether `db.whereTypeAsRef` (or the underlying query) exposes an `isLoading` companion ref — if it does, use that directly instead of a manual sentinel. If not, the cleanest approach is to change the default from `[]` to `null` so you can distinguish "not yet loaded" from "loaded empty":

```ts
// db.whereTypeAsRef returns null until the first query resolves
const languages = db.whereTypeAsRef<LanguageDto[] | null>(DocType.Language, null);
const isLoading = computed(() => languages.value === null);

// Then in the template, use languages.value ?? [] where an array is needed
```

If changing the API default isn't feasible, at minimum add `{ immediate: true }` to the watcher **and** initialise `isLoading` based on whether the ref already has data on mount:

```js
const isLoading = ref(languages.value.length === 0);
watch(languages, () => { isLoading.value = false; }, { immediate: true });
```

This is still imperfect (it can't distinguish "empty database" from "not loaded yet"), but eliminates the synchronous-cache failure.

---

### ⚠️ Warning — `aria-label` on a roleless `<div>` has no effect

**File:** `LoadingBar.vue`

```html
<div class="text-sm text-zinc-500 dark:text-slate-400" aria-label="Loading...">
```

`aria-label` is only meaningful on elements with an explicit or implicit ARIA role. On a plain `<div>`, screen readers ignore it entirely. For the loading state to be announced, the outer container needs `role="status"` and `aria-live="polite"`:

```html
<div
    class="flex w-full flex-col items-center"
    role="status"
    aria-live="polite"
    aria-label="Loading"
>
```

The visible text "Loading" inside the component is then sufficient for the accessible name — you can drop the `aria-label` on the inner `<div>` entirely.

---

### ⚠️ Warning — Mixed dark-mode design tokens

**File:** `LoadingBar.vue`

Light mode uses `zinc-*` tokens throughout, but dark mode introduces `slate-*` (`dark:text-slate-400`, `dark:bg-slate-700`, `dark:bg-slate-400`). Other dark-mode classes in the codebase use `dark:bg-zinc-*` and `dark:text-zinc-*`. Unless the design system intentionally maps these differently for dark mode, this will produce a slightly off colour compared to everything around it. Verify and align:

```html
<!-- Before -->
class="... dark:text-slate-400"
<!-- After -->
class="... dark:text-zinc-400"
```

---

### 💡 Nitpick — Redundant `gap-0` and empty `class` attribute

**File:** `LoadingBar.vue`

```html
<!-- gap-0 is the flex default — remove it -->
<div class="flex w-full flex-col items-center gap-0">

<!-- class="" is a no-op — remove it -->
<span class="">Loading</span>
```

---

### 💡 Nitpick — Use `watchOnce` from VueUse to clean up the stop-watcher pattern

**Files:** `LanguageOverview.vue`, `RedirectOverview.vue`

The manual stop-watcher pattern works, but `@vueuse/core` (already a project dependency) has `watchOnce` which is semantically cleaner and removes the boilerplate:

```js
import { watchOnce } from "@vueuse/core";

const isLoading = ref(true);
watchOnce(languages, () => { isLoading.value = false; });
```

Note: this is a style improvement only — it doesn't fix the empty-data bug described above.

---

**Summary:** The main thing needing a follow-up fix is the loading-state logic in `LanguageOverview.vue` and `RedirectOverview.vue`. Both are susceptible to showing an infinite loading bar when the dataset is empty. The accessibility and token-consistency issues in `LoadingBar.vue` are smaller but worth cleaning up in a follow-up pass.

---

PR #1577, opened 2026-04-29
Implement an estimated reading time calculation

### LINKED ISSUE #1496: API & APP: Estimated reading time calculation

ChrisTouo:


api/src/changeRequests/documentProcessing/processContentDto.ts
```diff
@@ -4,7 +4,7 @@ import { PostDto } from "../../dto/PostDto";
 import { TagDto } from "../../dto/TagDto";
 import { DbService } from "../../db/db.service";
 import { DocType, PublishStatus, Uuid } from "../../enums";
-import { computeFtsData } from "../../util/ftsIndexing";
+import { computeFtsData, wordCounter } from "../../util/ftsIndexing";
```
MrDirkelz: Do not think wordcounter should be part of ftsIndexing?

api/src/changeRequests/documentProcessing/processContentDto.ts
```diff
@@ -66,4 +66,13 @@ export default async function processContentDto(doc: ContentDto, db: DbService)
         doc.fts = ftsData.fts;
         doc.ftsTokenCount = ftsData.ftsTokenCount;
     }
+
+    const wordCount = wordCounter(doc.text || "");
+    doc.wordCounter = wordCount;
```
MrDirkelz:
```suggestion
    doc.wordCount = wordCount;
```

api/src/dto/ContentDto.ts
```diff
@@ -178,4 +178,14 @@ export class ContentDto extends _contentBaseDto {
     @IsString()
     @Expose()
     copyright?: string;
+
+    @IsOptional() // Optional as it is set upon change request processing
+    @IsNumber()
+    @Expose()
+    wordCounter?: number;
```
MrDirkelz:
```suggestion
    wordCount?: number;
```

api/src/dto/LanguageDto.ts
```diff
@@ -24,4 +24,9 @@ export class LanguageDto extends _contentBaseDto {
     @Expose()
     @IsNotEmpty()
     translations: Record<string, string>;
+
+    @IsOptional()
+    @IsInt()
+    @Expose()
+    readingSpeed?: number;
```
MrDirkelz:
```suggestion
    averageReadingSpeed?: number;
```

Nitpick: readingSpeed does not describe clearly enough

Also perhaps add JSDoc here to describe the readingSpeed? Since we use it to set the average reading time for users in specific languages

api/src/util/ftsIndexing.ts
```diff
@@ -153,6 +153,16 @@ export function stripHtml(html: string): string {
...
+/**
+ * Calculate the estimated reading time for a given text (in minutes).
+ * Uses a standard reading speed of 200 words per minute.
+ */
+export function wordCounter(text: string): number {
+    const strippedText = stripHtml(text).trim();
+    if (!strippedText) return 0;
+    return strippedText.split(/\s+/).length;
+}
+
```
MrDirkelz: THink this should not be part of the `ftsIndexing.ts` file

app/src/pages/SingleContent/SingleContent.vue
```diff
@@ -721,6 +726,14 @@ const playAudio = () => {
                             >
                                 By {{ content.author }}
                             </div>
+                            <div
+                                v-if="readingTime"
+                                class="flex items-center justify-center gap-1 text-center text-xs text-zinc-500 dark:text-slate-300"
+                            >
+                                <EyeIcon class="h-4 w-4" />
+                                {{ readingTime ? `${readingTime} min` : "-" }}
```
MrDirkelz: The reading time computed value already returns "-" so it should either be done only there or only here

cms/src/components/languages/EditLanguage.vue
```diff
@@ -421,6 +422,16 @@ const contentActions = computed(() => {
                     :disabled="!canEditOrCreate"
                 />
 
+                <LInput
+                    inputType="number"
+                    label="Reading Speed (words per minute)"
```
MrDirkelz:
```suggestion
                    label="Average reading speed (words per minute)"
```

Maybe in a description or in the placeholder also add that this is for the average reading speed for people speaking this language as first language perhaps?

shared/src/types/dto.ts
```diff
@@ -108,6 +109,8 @@ export type ContentDto = ContentBaseDto & {
     fts?: string[];
     ftsTokenCount?: number;
     copyright?: string;
+    wordCounter?: number;
+    readingTime?: number;
```
MrDirkelz: See above comments on new dto fields

api/src/changeRequests/documentProcessing/processContentDto.ts
```diff
@@ -66,4 +66,13 @@ export default async function processContentDto(doc: ContentDto, db: DbService)
         doc.fts = ftsData.fts;
         doc.ftsTokenCount = ftsData.ftsTokenCount;
     }
+
+    const wordCount = wordsCount(doc.text || "");
```
MrDirkelz: Do we need the default value? If the text is empty wouldn't the wordcount just be 0?

cms/src/components/forms/LInput.vue
```diff
@@ -59,6 +59,15 @@ const autoResize = () => {
     }
 };
 
+const handleInput = (e: Event) => {
+    const target = e.target as HTMLInputElement;
+    let val: any = target.value;
+    if (target.type === "number" && val !== "") {
+        val = Number(val);
+    }
```
MrDirkelz: This can be simplified by using Number.parseInt and Number.NaN for validation

app/src/pages/SingleContent/SingleContent.vue
```diff
@@ -564,6 +564,11 @@ const playAudio = () => {
         addToMediaQueue(content.value);
     }
 };
+
+const readingTime = computed(() => {
```
MrDirkelz, line 603: This computed can be removed

app/src/pages/SingleContent/SingleContent.vue
```diff
@@ -724,6 +729,14 @@ const playAudio = () => {
                             >
                                 By {{ content.author }}
                             </div>
+                            <div
+                                v-if="readingTime"
```
MrDirkelz, line 786:
```suggestion
                                v-if="content.readingTime"
```

app/src/pages/SingleContent/SingleContent.vue
```diff
@@ -724,6 +729,14 @@ const playAudio = () => {
...
+                                class="flex items-center justify-center gap-1 text-center text-xs text-zinc-500 dark:text-slate-300"
+                            >
+                                <EyeIcon class="h-4 w-4" />
+                                {{ readingTime }}
```
MrDirkelz:
```suggestion
                                {{ content.readingTime }}
```

MrDirkelz (commented):
Nice! Maybe some tests for the word counting function could also be helpful as well as for its visual display?

api/src/changeRequests/documentProcessing/processContentDto.ts
```diff
@@ -66,4 +66,13 @@ export default async function processContentDto(doc: ContentDto, db: DbService)
...
     }
+
+    const wordCount = wordsCount(doc.text);
+    doc.wordCount = wordCount;
+
+    // Calculate reading time based on language reading speed
+    const langQuery = await db.getDoc(doc.language);
+    const langDoc = langQuery.docs[0] as any;
+    const averageReadingSpeed = langDoc?.averageReadingSpeed || 200;
+    doc.readingTime = Math.ceil(wordCount / averageReadingSpeed);
```
MrDirkelz: 🚨 **Critical — `||` silently masks a valid `0` and division by zero produces `Infinity`**

Two bugs on this line:

1. `langDoc?.averageReadingSpeed || 200` uses `||`, which falls back to `200` for *any* falsy value — including an explicitly stored `0`. Switch to nullish coalescing so only `null`/`undefined` triggers the default:
   ```ts
   const averageReadingSpeed = langDoc?.averageReadingSpeed ?? 200;
   ```

2. Even with that fix, if a user somehow saves `averageReadingSpeed: 0`, `Math.ceil(wordCount / 0)` evaluates to `Infinity`, which gets persisted to the database. Add `@Min(1)` to `LanguageDto.averageReadingSpeed` (see comment there) to block this at the validation layer.

Also on line 70: `wordsCount(doc.text)` — `doc.text` may be `undefined` at runtime even if typed as `string`. Mirror what the v14 migration does: `wordsCount(doc.text || "")`.

---

api/src/dto/LanguageDto.ts
```diff
@@ -24,4 +27,9 @@ export class LanguageDto extends _contentBaseDto {
     @Expose()
     @IsNotEmpty()
     translations: Record<string, string>;
+
+    @IsOptional()
```
MrDirkelz, line 28: 🚨 **Critical — Missing `@Min(1)` allows zero, which causes division by zero**

`@IsInt()` accepts `0`, but `Math.ceil(wordCount / 0)` = `Infinity`, which would be persisted to the database. Add a minimum bound:

```ts
@IsOptional()
@IsInt()
@Min(1)
@Expose()
averageReadingSpeed?: number;
```

You'll also need to import `Min` from `class-validator`.

---

api/src/db/schemaUpgrade/v14.ts
```diff
@@ -0,0 +1,65 @@
...
+            let languagesUpdated = 0;
+            let languagesWithoutSpeed = 0;
+            let contentsWithoutLanguage = 0;
+
+            const readingSpeeds = new Map<string, number>();
+
+            await db.processAllDocs([DocType.Language], async (doc: any) => {
+                if (!doc) return;
+                if (!doc.averageReadingSpeed) {
+                    doc.averageReadingSpeed = 200;
```
MrDirkelz, line 20: 🚨 **Critical — `!doc.averageReadingSpeed` treats `0` as "not set", silently overwriting it**

`!0` is `true`, so any language that had `averageReadingSpeed: 0` explicitly stored will be overwritten to `200` by this migration. Use a null-check instead:

```ts
if (doc.averageReadingSpeed == null) {
    doc.averageReadingSpeed = 200;
    await db.insertDoc(doc);
    languagesWithoutSpeed++;
}
```

Similarly on line 34, prefer `??` over `||`:
```ts
const speed = readingSpeeds.get(doc.language) ?? 200;
```

---

api/src/db/schemaUpgrade/v14.ts
```diff
@@ -0,0 +1,65 @@
...
+                    await db.insertDoc(doc);
+                    languagesWithoutSpeed++;
+                }
+                readingSpeeds.set(doc._id, doc.averageReadingSpeed);
+                languagesUpdated++;
+            });
```
MrDirkelz, line 26: ⚠️ **`languagesUpdated` counts all languages processed, not those actually updated**

`languagesUpdated++` runs for every language doc, whether or not it was modified. The log message on line 46 then reports this as the number of languages "updated", which is misleading. Rename to `languagesProcessed`, or only increment when `insertDoc` is called (i.e. inside the `if` block above).

---

api/src/util/ftsIndexing.ts
```diff
@@ -153,6 +153,15 @@ export function stripHtml(html: string): string {
     return result.join("");
 }
 
+/**
+ * Calculate the estimated reading time for a given text (in minutes).
+ */
+export function wordsCount(text: string): number {
```
MrDirkelz: 💡 **Wrong JSDoc — says "reading time" but function returns a word count**

```ts
/**
 * Count the number of words in a given HTML string.
 */
export function wordsCount(text: string): number {
```

Also worth noting: `split(/\s+/)` is inaccurate for languages that don't use spaces between words (Chinese, Japanese, Korean). A 500-character Chinese passage with no spaces counts as a single word. If the platform serves content in those languages this will silently produce very wrong reading times. At minimum, document the limitation in a comment.

---

cms/src/components/forms/LInput.vue
```diff
@@ -39,7 +39,7 @@ const props = withDefaults(defineProps<Props>(), {
 });
 
 const emit = defineEmits<{
-    (event: "update:modelValue", value: string): void;
+    (event: "update:modelValue", value: any): void;
```
MrDirkelz: ⚠️ **Type regression — `any` weakens safety of a globally-shared component**

`LInput` is used everywhere in the CMS. Changing the emit type to `any` silently breaks all parent components that relied on `string`, and erases the type guarantee. Use a proper union instead:

```ts
const emit = defineEmits<{
    (event: "update:modelValue", value: string | number | null): void;
}>();
```

The `handleInput` function can then return `string | number | null` and the compiler will catch misuse at call sites.

---

app/src/pages/SingleContent/SingleContent.vue
```diff
@@ -724,6 +724,14 @@ const playAudio = () => {
...
                                 By {{ content.author }}
                             </div>
+                            <div
+                                v-if="content.readingTime"
+                                class="flex items-center justify-center gap-1 text-center text-xs text-zinc-500 dark:text-slate-300"
+                            >
+                                <EyeIcon class="h-4 w-4" />
+                                {{ content.readingTime }}
+                            </div>
+
```
MrDirkelz, line 792: 💡 **Two UX issues: wrong icon and missing unit**

1. `EyeIcon` (👁) means "view/visibility" — it will confuse users. A `ClockIcon` or `BookOpenIcon` from the same Heroicons package is semantically correct for reading time.

2. `{{ content.readingTime }}` displays a bare number (e.g. `3`) with no unit. A reader has no idea if that means minutes, seconds, or pages. Show the unit explicitly:

```html
<ClockIcon class="h-4 w-4" />
{{ content.readingTime }} min read
```

(Or use an i18n key if the app has localisation for UI strings.)

---

cms/src/components/languages/EditLanguage.vue
```diff
@@ -421,6 +422,16 @@ const contentActions = computed(() => {
...
                 />
 
+                <LInput
+                    type="number"
+                    label="Average reading Speed (words per minute)"
+                    name="averageReadingSpeed"
+                    v-model="editable.averageReadingSpeed"
+                    class="mb-4 w-full"
+                    placeholder="Enter the average reading speed for people speaking this language( 200 wpm is the default)"
+                    :disabled="!canEditOrCreate"
```
MrDirkelz, line 432: 💡 **Label capitalisation and placeholder formatting**

Two small copy issues:
- `"Average reading Speed"` → `"Average reading speed"` (only first word capitalised)
- `"...language( 200 wpm is the default)"` → `"...language (200 wpm is the default)"` (missing space before parenthesis)

---

MrDirkelz (changes requested):
## Code Review — Estimated Reading Time (#1577)

Good foundational work — the feature is reasonably scoped and the migration strategy (pre-loading all language speeds into a `Map` before iterating content) is solid. There are a few issues that need to be addressed before this merges, including one that will store `Infinity` in the database if a user ever saves a reading speed of `0`.

See inline comments for specifics. Summary below.

---

### 🚨 Critical

1. **Division by zero → `Infinity` stored in DB.** `LanguageDto.averageReadingSpeed` has no `@Min(1)` guard. A zero value passes validation, and `Math.ceil(wordCount / 0)` = `Infinity`. Add `@Min(1)` to the DTO.
2. **`!doc.averageReadingSpeed` treats `0` as "missing"** in the v14 migration, silently overwriting an explicit zero with 200. Use `doc.averageReadingSpeed == null` instead.
3. **`||` should be `??` (nullish coalescing)** in both `processContentDto.ts` and `v14.ts`. Using `||` causes any falsy value (including a valid `0`) to fall back to 200, hiding the actual stored value.

### ⚠️ Warning

4. **`doc.text` is not guarded against `undefined`** in `processContentDto`. The v14 migration correctly uses `doc.text || ""` but the live path does not — calling `wordsCount(undefined)` passes undefined to a function typed as `(text: string)`.
5. **Extra DB round-trip per content save** — `db.getDoc(doc.language)` is called on every content save with no caching. For bulk operations this compounds quickly.
6. **`split(/\s+/)` word counting is wrong for CJK languages** (Chinese, Japanese, Korean) that don't use spaces between words. A 500-character Chinese text would count as a single word. At minimum, document this limitation.
7. **Type regression in `LInput.vue`** — emitting `value: any` instead of `value: string | number | null` weakens type safety across a globally-shared component.
8. **`languagesUpdated` counter is misleading** in v14 — it counts all languages *processed*, not those actually *updated*.

### 💡 Nitpick / Suggestion

9. **Wrong JSDoc on `wordsCount`** — says "reading time" but the function returns a word count.
10. **`EyeIcon` is semantically wrong** for reading time — a `ClockIcon` or `BookOpenIcon` would be clearer.
11. **Missing unit in UI** — `{{ content.readingTime }}` displays a bare number. Show `{{ content.readingTime }} min read` instead.
12. **Label capitalisation** — `"Average reading Speed"` → `"Average reading speed"`.
13. **Placeholder missing space** — `"...language( 200 wpm"` → `"...language (200 wpm"`.

---

cms/src/components/forms/LInput.vue
MrDirkelz: Revert this any

cms/src/components/forms/LInput.vue
```diff
@@ -59,6 +59,16 @@ const autoResize = () => {
     }
 };
 
+const handleInput = (e: Event) => {
+    const target = e.target as HTMLInputElement;
+    let val: any = target.value;
+    if (target.type === "number" && val !== "") {
+        const parsed = parseInt(val, 10);
```
MrDirkelz: your using Number.NaN so this seems like a duplicate conditional check

cms/src/components/languages/EditLanguage.vue
```diff
@@ -421,6 +422,16 @@ const contentActions = computed(() => {
                     :disabled="!canEditOrCreate"
                 />
 
+                <LInput
+                    type="number"
+                    label="Average reading Speed (words per minute)"
```
MrDirkelz:
```suggestion
                    label="Average reading speed (words per minute)"
```

cms/src/components/languages/EditLanguage.vue
```diff
@@ -421,6 +422,16 @@ const contentActions = computed(() => {
...
+                    name="averageReadingSpeed"
+                    v-model="editable.averageReadingSpeed"
+                    class="mb-4 w-full"
+                    placeholder="Enter the average reading speed for people speaking this language( 200 wpm is the default)"
```
MrDirkelz:
```suggestion
                    placeholder="Enter the average reading speed for people speaking this language ( 200 wpm is the default)"
```

MrDirkelz (changes requested):

api/src/util/ftsIndexing.ts
ivanslabbert: this should be in ftsIndexing.ts, but not as a separate function. It should be an additional output from the fts function that is calculated during the same FTS content scan. The title and author should however be excluded from the word count, while the summary should still be included.

api/src/changeRequests/documentProcessing/processContentDto.ts
```diff
@@ -66,4 +66,13 @@ export default async function processContentDto(doc: ContentDto, db: DbService)
...
     }
+
+    const wordCount = ftsData.wordCount;
+    doc.wordCount = wordCount;
+
+    // Calculate reading time based on language reading speed
+    const langQuery = await db.getDoc(doc.language);
+    const langDoc = langQuery.docs[0] as any;
+    const averageReadingSpeed = langDoc?.averageReadingSpeed || 200;
+    doc.readingTime = Math.ceil(wordCount / averageReadingSpeed);
```
ivanslabbert: We need to set the wordcount on the doc instead of the reading time. Reading time should be calculated on demand in the app by dividing the word count of the doc by the words per minute value from the language doc.

api/src/dto/ContentDto.ts
```diff
@@ -178,4 +178,14 @@ export class ContentDto extends _contentBaseDto {
...
+
+    @IsOptional() // Optional as it is set upon change request processing
+    @IsNumber()
+    @Expose()
+    wordCount?: number;
+
+    @IsOptional()
+    @IsNumber()
+    @Expose()
+    readingTime?: number;
```
ivanslabbert: remove

api/src/dto/LanguageDto.ts
```diff
@@ -1,9 +1,12 @@
...
+import { IsInt, IsNotEmpty, IsOptional, IsString, Min } from "class-validator";
 import { _contentBaseDto } from "./_contentBaseDto";
 import { Expose } from "class-transformer";
 
 /**
- * Database structured Language object
+ * Database structured Language object.
+ * * @property {number} averageReadingSpeed - The average reading speed in words per minute
+ * for this language, used to calculate reading time for content.
+ * If not provided, a default of 200 wpm is assumed.
```
ivanslabbert: Remove cmment

api/src/util/ftsIndexing.ts
ivanslabbert, line 224: We can perhaps do the wordcount on this text const, which will avoid having to re-process the text again

api/src/util/ftsIndexing.ts
```diff
@@ -230,9 +231,13 @@ export function computeFtsData(doc: Record<string, any>): FtsData | undefined {
...
-        ([token, tf]) => token + ":" + tf,
-    );
+    const fts: string[] = Array.from(aggregatedTf.entries()).map(([token, tf]) => token + ":" + tf);
 
-    return { fts, ftsTokenCount: totalTokenCount };
+    // Compute word count for reading time estimation (strip HTML and split on whitespace)
+    const summary = doc.summary || "";
+    const text = stripHtml(doc.text || "");
+    const combinedContent = `${summary} ${text}`.trim();
+    const wordCount = combinedContent ? combinedContent.split(/\s+/).length : 0;
```
ivanslabbert: See comment above

shared/src/types/dto.ts
```diff
@@ -108,6 +109,8 @@ export type ContentDto = ContentBaseDto & {
     fts?: string[];
     ftsTokenCount?: number;
     copyright?: string;
+    wordCount?: number;
+    readingTime?: number;
```
ivanslabbert: Remove

api/src/util/ftsIndexing.ts
```diff
@@ -230,9 +235,7 @@ export function computeFtsData(doc: Record<string, any>): FtsData | undefined {
 
     if (aggregatedTf.size === 0) return undefined;
 
-    const fts: string[] = Array.from(aggregatedTf.entries()).map(
-        ([token, tf]) => token + ":" + tf,
-    );
+    const fts: string[] = Array.from(aggregatedTf.entries()).map(([token, tf]) => token + ":" + tf);
```
MrDirkelz: Would it be possible to revert this formatting. Prior version is easier to read

app/src/pages/SingleContent/SingleContent.vue
```diff
@@ -564,6 +577,17 @@ const playAudio = () => {
         addToMediaQueue(content.value);
     }
 };
+
+const readingTime = computed(() => {
+    if (!content.value) return "";
+    const wordCount = content.value.wordCount!;
+
+    const readingSpeed = currentLanguageData.value?.averageReadingSpeed
+        ? currentLanguageData.value.averageReadingSpeed
```
MrDirkelz: See above comment on `currentLanguageData.value` usage risks

MrDirkelz (commented):
Please check merge conflicts and comments.

app/src/pages/SingleContent/SingleContent.vue
```diff
@@ -282,6 +286,15 @@ const tags = useDexieLiveQueryWithDeps(
...
 
+const currentLanguageData = useDexieLiveQueryWithDeps(
+    [content],
+    ([content]) => {
+        if (!content?.language) return Promise.resolve(undefined);
+        return db.docs.get(content.language) as unknown as Promise<LanguageDto | undefined>;
+    },
+    { initialValue: undefined as LanguageDto | undefined },
+);
+
```
ivanslabbert: I think we are already getting the language docs somewhere else in SingleContent?

MrDirkelz (dismissed):
# Code Review: Estimated Reading Time Calculation

Great first feature, @ChrisTouo! The overall approach is sound — storing `wordCount` on the document and computing reading time from language speed is the right mental model. There are a few bugs that need to be fixed before this merges, plus some design inconsistencies worth addressing.

---

## 🚨 Critical

### 1. `processContentDto.ts` — TypeError when `computeFtsData` returns `undefined`

`computeFtsData` returns `FtsData | undefined` (it returns `undefined` when the doc has no indexable trigrams, e.g. very short content or content with only 1–2 character words). The new line is placed **outside** the `if (ftsData)` guard, so it will throw a runtime `TypeError` for those documents:

```typescript
const ftsData = computeFtsData(doc);
if (ftsData) {
    doc.fts = ftsData.fts;
    doc.ftsTokenCount = ftsData.ftsTokenCount;
}

// 💥 ftsData can be undefined here — throws "Cannot read properties of undefined"
doc.wordCount = ftsData.wordCount;
```

**Fix:** move the assignment inside the guard and handle the `undefined` case:

```typescript
const ftsData = computeFtsData(doc);
if (ftsData) {
    doc.fts = ftsData.fts;
    doc.ftsTokenCount = ftsData.ftsTokenCount;
    doc.wordCount = ftsData.wordCount;
} else {
    doc.wordCount = 0;
}
```

---

### 2. `SingleContent.vue` — Non-null assertion on optional field silently renders "NaN min"

`wordCount` is `?` (optional) in the DTO, meaning it will be `undefined` for any content document that existed before the v14 migration runs (or for content that the migration skipped). The `!` assertion suppresses TypeScript but doesn't prevent the runtime problem:

```typescript
const wordCount = content.value.wordCount!;  // undefined for pre-migration docs
// Math.ceil(undefined / 200) → NaN
// NaN.toString() → "NaN"  (truthy — the v-if passes!)
return Math.ceil(wordCount / readingSpeed).toString();
```

Users would see **"NaN min"** for any article that hasn't been re-processed.

**Fix:**

```typescript
const readingTime = computed(() => {
    if (!content.value?.wordCount) return "";
    const currentLanguage = languages.value.find(l => l._id === content.value?.language);
    const readingSpeed = currentLanguage?.averageReadingSpeed || 200;
    return Math.ceil(content.value.wordCount / readingSpeed).toString();
});
```

---

## ⚠️ Warning

### 3. `v14.ts` — Falsy check incorrectly handles `averageReadingSpeed = 0`

`!doc.averageReadingSpeed` is `true` when the value is `0`, which would incorrectly overwrite a deliberately-set zero with the default 200. Use a nullish check instead:

```typescript
// Before (buggy):
if (!doc.averageReadingSpeed) { ... }

// After (correct):
if (doc.averageReadingSpeed == null) { ... }
```

---

### 4. `v14.ts` — `readingTime` is orphaned in the database

The migration writes `doc.readingTime` to content documents, but:
- `readingTime` is not added to `ContentDto` or `shared/src/types/dto.ts`
- `processContentDto.ts` only updates `wordCount`, not `readingTime`, on subsequent saves

This means: after a content editor updates the word count of an article, `readingTime` in the DB is stale (the frontend compensates by recalculating, but anyone reading the raw document — e.g. a mobile offline client querying CouchDB directly — sees stale data).

**Pick one of these two consistent paths:**

**Option A — Store and maintain `readingTime` in the DB:** Add `readingTime` to both DTOs, look up the language speed in `processContentDto`, and always update it on save. The frontend then reads the pre-computed value directly.

**Option B — Don't persist `readingTime` at all:** Remove `doc.readingTime` from `v14.ts`. Keep only `wordCount` in the DB and continue computing reading time on the frontend. The migration becomes simpler and there's no stale-data risk.

The current code is a confusing mix of both.

---

### 5. `LInput.vue` — `parseFloat` for an `@IsInt()` field

`averageReadingSpeed` is validated server-side with `@IsInt()`, but the frontend parses with `Number.parseFloat`. A user entering `"150.7"` gets `150.7` in the payload, which passes browser validation silently but fails with a server-side 400 error and no user-visible feedback.

```typescript
// Before:
model.value = type === "number" && value !== "" ? Number.parseFloat(value) : value;

// After:
model.value = type === "number" && value !== "" ? Math.round(Number(value)) : value;
```

Also consider adding `step="1"` to the `<LInput>` in `EditLanguage.vue` to enforce integers at the browser level.

---

### 6. `v14.ts` — Misleading `languagesUpdated` counter

The counter is incremented for **every language processed**, not only for ones that were actually modified. The final log message saying "X languages have been updated" is inaccurate.

```typescript
// Rename to languagesProcessed, and count actual updates separately:
let languagesProcessed = 0;
let languagesDefaulted = 0;  // was languagesWithoutSpeed

await db.processAllDocs([DocType.Language], async (doc: any) => {
    if (!doc) return;
    if (doc.averageReadingSpeed == null) {
        doc.averageReadingSpeed = 200;
        await db.insertDoc(doc);
        languagesDefaulted++;
    }
    readingSpeeds.set(doc._id, doc.averageReadingSpeed);
    languagesProcessed++;
});
```

---

## 💡 Nitpick / Suggestion

### 7. `ftsIndexing.ts` — Magic string `"title"` for word count exclusion

The word count exclusion is a fragile string comparison that will silently break if the field is ever renamed in `FTS_FIELDS`. This is inconsistent with how `isHtml` is already handled as a typed config flag. Add a `countWords` property to `FtsFieldConfig` instead:

```typescript
type FtsFieldConfig = {
    name: string;
    boost: number;
    isHtml: boolean;
    countWords: boolean;
};

const FTS_FIELDS: FtsFieldConfig[] = [
    { name: "title",   boost: 3.0, isHtml: false, countWords: false },
    { name: "summary", boost: 1.5, isHtml: false, countWords: true  },
    { name: "text",    boost: 1.0, isHtml: true,  countWords: true  },
    { name: "author",  boost: 1.0, isHtml: false, countWords: true  },
];

// In computeFtsData:
if (field.countWords) {
    wordCount += text.split(/\s+/).filter(Boolean).length;
}
```

---

### 8. `SingleContent.vue` — "min" is not localized

The app uses `vue-i18n` throughout. The hardcoded suffix should go through `t()`:

```html
{{ readingTime }} {{ t('singlecontent.reading_time_unit') }}
```

---

### 9. `processContentDto.spec.ts` — Test validates output using the same function it's testing

The test imports `computeFtsData` to derive `expectedWordCount`, then asserts that the DB doc's `wordCount` equals that value. This is circular — the test can only catch regressions if `computeFtsData` itself is broken, but `computeFtsData` already has its own unit tests. Use a concrete expected value instead:

```typescript
// Before (circular):
const expectedWordCount = computeFtsData(changeRequest.doc).wordCount;
expect(dbDoc.docs[0].wordCount).toBe(expectedWordCount);

// After (actually tests the integration):
expect(dbDoc.docs[0].wordCount).toBe(300); // 300 words in sampleText
```

---

### 10. `EditLanguage.vue` — Placeholder spacing typo

```
"Enter the average reading speed for people speaking this language ( 200 wpm is the default)"
                                                                   ^--- extra space
```

Should be `(200 wpm is the default)`.

---

### 11. `v14.ts` — Anonymous default export

Named exports produce better stack traces. A minor change that aids debugging:

```typescript
// Before:
export default async function (db: DbService) { ... }

// After:
export default async function upgradeToV14(db: DbService) { ... }
```

---

**Summary of must-fix before merge:** Items 1, 2, 3, 4, and 5 are the blockers. Items 6–11 can be addressed in this PR or tracked as follow-ups.


---

ivanslabbert (approved):

MrDirkelz (dismissed):
Tested with @ChrisTouo and the upgrade script did not work for him locally. So rather adding changes requested until this has been resolved/confirmed to be working.

PR #1582, opened 2026-05-01
Fix: Improve token refresh cache handling

### LINKED ISSUE #1581: APP: Token not being refreshed after expiration

MrDirkelz:
When the server rejects a token, the Auth0 SDK's default cache can return a stale but still client-valid token. This prevents a proper refresh, leading to infinite re-login loops.

This change introduces an `ignoreCache` option to
`refreshTokenSilently`. When true, it forces `getAccessTokenSilently` to bypass the SDK's cache and hit the `/oauth/token` endpoint. This ensures a fresh token is obtained even if the cached one is still within its leeway period but has been revoked by the server.

Tests have been added to cover this cache behavior and its regression implications.

### DIFF
--- a/app/src/auth.ts
+++ b/app/src/auth.ts
@@ -140,18 +140,18 @@ export async function setupAuth(app: App<Element>, router: Router): Promise<void
 
 /**
  * Ask Auth0 for a fresh access token via the refresh token, then push it to
- * the Authorization header and the socket. Returns true on success. Call from
- * the socket's connect_error handler before falling back to a visible
- * re-login — this is what `useRefreshTokens: true` is for.
+ * the Authorization header and the socket. Call from the socket's
+ * `connect_error` handler before falling back to a visible re-login — this is
+ * what `useRefreshTokens: true` is for.
  *
- * Pass `ignoreCache: true` after the server has rejected a token. Otherwise
- * Auth0's default `cacheMode: "on"` can hand back the same cached
- * still-client-valid token (60s leeway) that the server just rejected, so
- * we'd loop forever on any clock skew or server-side revocation.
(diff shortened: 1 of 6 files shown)


app/src/auth.ts
```diff
@@ -143,11 +143,20 @@ export async function setupAuth(app: App<Element>, router: Router): Promise<void
  * the Authorization header and the socket. Returns true on success. Call from
  * the socket's connect_error handler before falling back to a visible
  * re-login — this is what `useRefreshTokens: true` is for.
+ *
+ * Pass `ignoreCache: true` after the server has rejected a token. Otherwise
+ * Auth0's default `cacheMode: "on"` can hand back the same cached
+ * still-client-valid token (60s leeway) that the server just rejected, so
+ * we'd loop forever on any clock skew or server-side revocation.
```
ivanslabbert: This should be JSDoc on the ignoreCache parameter itself

cms/src/auth.ts
```diff
@@ -159,11 +159,20 @@ export async function setupAuth(app: App<Element>): Promise<void> {
  * the Authorization header and the socket. Returns true on success. Call from
  * the socket's connect_error handler before falling back to a visible
  * re-login — this is what `useRefreshTokens: true` is for.
+ *
+ * Pass `ignoreCache: true` after the server has rejected a token. Otherwise
+ * Auth0's default `cacheMode: "on"` can hand back the same cached
+ * still-client-valid token (60s leeway) that the server just rejected, so
+ * we'd loop forever on any clock skew or server-side revocation.
```
ivanslabbert: Move to parameter specific JSDoc

ivanslabbert (approved):

PR #1595, opened 2026-05-04
API: Refactor group deletion to use DeleteCmd event

MrDirkelz:
The `groupUpdate` event now also emits a `DeleteCmd` payload when a group is soft-deleted. This allows the permission system to properly remove deleted groups from its access maps. The upstream filtering of CouchDB tombstones ensures that only `DeleteCmd` events are processed, as the tombstones themselves lack a `type` field.

### DIFF
--- a/api/src/permissions/permissions.service.ts
+++ b/api/src/permissions/permissions.service.ts
@@ -1,6 +1,7 @@
 import { Uuid, DocType, AclPermission } from "../enums";
 import { GroupAclEntryDto } from "../dto/GroupAclEntryDto";
 import { GroupDto } from "../dto/GroupDto";
+import { DeleteCmdDto } from "../dto/DeleteCmdDto";
 import { DbService } from "../db/db.service";
 import { EventEmitter } from "node:events";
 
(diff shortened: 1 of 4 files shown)


ivanslabbert (commented):
If permission system "cleanup" was only triggered on change request receipt, other API instances will not receive the group removal command. Group removal from the permission system should rely only on the database change feed (as per the suggested change in this PR), so the functionality should now be removed from the logic where change requests are processed.

api/src/permissions/permissions.service.ts
```diff
@@ -142,23 +142,46 @@ export class PermissionSystem extends EventEmitter {
         dbService = db;
 
         let initialized = false;
-        let updateQueue: any[] = [];
+        const updateQueue: any[] = [];
```
ivanslabbert: fix typing to support both group and delete commands

api/src/permissions/permissions.service.ts
```diff
@@ -142,23 +142,46 @@ export class PermissionSystem extends EventEmitter {
...
-            }
+            if (initialized) drainQueue();
         });
 
         // Add existing groups to permission system
         const dbGroups = await dbService.getGroups();
         this.upsertGroups(dbGroups.docs);
 
-        // Add changes that might have occurred while the permission system was being initialized
-        this.upsertGroups(updateQueue);
```
ivanslabbert, line 160: A better and cleaner way would perhaps be to update upsertGroups to have delete command support, and handle deletion at that level.

api/src/permissions/permissions.service.ts
```diff
@@ -337,12 +339,19 @@ export class PermissionSystem extends EventEmitter {
     }
 
     /**
-     * Create or update groups from passed array of group database documents
+     * Create or update groups from passed array of group database documents.
+     * DeleteCmd payloads interleaved in the array trigger removal of the
+     * referenced group in arrival order.
      * @param groupDocs
      */
```
ivanslabbert, line 342: Revert these changes - we still need to clear the update queue after applying updates

ivanslabbert (approved):

PR #1609, opened 2026-05-11
APP: Add search shortcute to homepage

MrDirkelz:
Introduces a new `HomePageSearch` component that displays a search bar on the homepage. The component includes an animated pulse effect on first load to draw user attention. It also supports keyboard shortcuts for opening the search overlay. The search heading and hint are now localized.

image

image

https://github.com/user-attachments/assets/7eec0eb5-47e7-4b2c-ba8c-8e1aad1243ea


MrDirkelz (commented):
## Code Review — PR #1609: Add Search Shortcut to Homepage

Good component structure and accessibility awareness overall. One critical logic inversion needs to be fixed before this ships.

---

### 🚨 Critical — Inverted screen-size guard

**File:** `app/src/pages/HomePage.vue:15`

```html
<HomePageSearch v-if="isMdScreen" />
```

`isMdScreen` is defined as `windowWidth < 768` — meaning it is `true` **only on mobile** (< 768 px). The component therefore renders exclusively on small phone screens and is hidden on every tablet and desktop. That's the exact opposite of the intended behaviour: keyboard shortcuts (⌘K / Ctrl+K) are a desktop feature.

**Fix:**

```html
<HomePageSearch v-if="!isMdScreen" />
```

This shows the component on md+ screens (≥ 768 px), which aligns with the inner `<kbd v-if="!isMobileScreen">` guard that already handles hiding the shortcut hint on screens < 1024 px.

> The name `isMdScreen` is a trap — it does not mean "is a medium screen"; it means "is *smaller than* the md breakpoint". This naming has now caused at least one bug. Consider documenting it prominently in `globalConfig.ts`, or renaming to `isBelowMd` / `isSmallScreen`.

---

### ⚠️ Warning — Unguarded `localStorage` read at module init

**File:** `app/src/components/HomePage/HomePageSearch.vue:17`

```ts
const shouldPulse = ref(!localStorage.getItem(PULSE_FLAG_KEY));
```

The *write* in `onMounted` is wrapped in a `try/catch`, but the *read* at initialisation is not. In certain test environments (JSDOM without a `localStorage` stub) or unusual browser configurations this throws synchronously, crashing the component before any error boundary can catch it.

**Fix:**

```ts
function readPulseFlag(): boolean {
    try {
        return !localStorage.getItem(PULSE_FLAG_KEY);
    } catch {
        return false; // fail silently; the pulse is a nice-to-have
    }
}
const shouldPulse = ref(readPulseFlag());
```

---

### 💡 Nitpick — CSS comment verbosity

The `<style scoped>` comments describe *what* the CSS does ("Two-stop colour ramp", "a tight yellow-500 core fades into…") rather than *why* specific choices were made. A reader can see `box-shadow` values; they cannot easily infer intent. Prefer:

```css
/* ::before overlay carries the glow so box-shadow doesn't clip rounded corners */
/* `forwards` retains the final transparent frame — prevents the shadow snapping back */
```

---

### 💡 Nitpick — `h2` has no explicit type scale

```html
<h2 class="mb-3 text-center text-zinc-800 dark:text-slate-100">
```

Tailwind preflight removes default heading sizes, so this renders at `text-base`. Add an explicit size like `text-lg font-semibold` to make the intent stable.

---

**Summary:** The inverted `v-if` guard is a ship-stopper — the feature is invisible to its target audience in the current code. Everything else is low-risk. Happy to approve once that single line is corrected.

---

ivanslabbert (approved):

PR #1611, opened 2026-05-12
APP: Refactor TopBar to use ResizeObserver

MrDirkelz:
Replaced window resize listener with ResizeObserver for more robust screen size detection. This ensures the logo's small screen state is updated correctly even when only the container size changes without a viewport resize. Also removed unnecessary watch and nextTick.


ivanslabbert (approved):
Approved, but please check failing tests

PR #1612, opened 2026-05-12
1226 api remove redundant environmental variables and update setup-dev.sh with updated auth setup

### LINKED ISSUE #1226: API: Remove redundant environmental variables

ChrisTouo:


scripts/setup-dev.sh
```diff
@@ -186,7 +186,7 @@ apply_auth0_env() {
     info "  Example: https://your-api.com/api or https://luminary.example.com"
     info "  This should match the Identifier you set in Auth0 APIs section."
```
MrDirkelz: The whole auth0 setup can be removed from this script and the new setup can rather be referenced as a next step after this setup has taken place

MrDirkelz (commented):
Just a comment + check conflicts

MrDirkelz (approved):
Really nice and clean job!

PR #1619, opened 2026-05-18
APP: Refactor syncLocalChanges to drain queue sequentially

MrDirkelz:
The `syncLocalChanges` function has been refactored to sequentially process local changes from the database when the application is online. This change removes the `processChangeReqLock` and introduces a `drain` function that reads the head of the queue directly from the database to ensure it's always up-to-date. The watcher now triggers the `drain` function when connectivity or the local changes array changes.

### DIFF
--- a/shared/src/rest/syncLocalChanges.ts
+++ b/shared/src/rest/syncLocalChanges.ts
@@ -8,40 +8,26 @@ import { changeReqErrors, changeReqWarnings } from "../config";
 
 let running = false;
 
-/**
- * Drain pending local changes one-at-a-time while online. Exits on empty
- * queue, disconnect, or a non-ack response. The watcher re-enters on the
- * next localChanges mutation or reconnect.
- *
- * We read the head of the queue from the db (not the ref) so that we see
- * the result of `applyLocalChangeAck` immediately — the Dexie live query
- * refreshes async, so the ref would be one step behind inside the loop.
- */
 async function drain() {
     if (running) return;
     running = true;
-    let attempts = 0;
     try {
-        while (isConnected.value) {
-            const change = (await db.localChanges.toCollection().first()) as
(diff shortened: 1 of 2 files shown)


shared/src/rest/syncLocalChanges.ts
```diff
@@ -1,72 +1,59 @@
...
-            return;
-        }
-        // When coming online, allow processing and attempt immediately
-        processChangeReqLock.value = false;
-        attemptSync();
-    });
+    watch(
+        [isConnected, localChanges],
+        () => {
+            drain().catch((err) => console.error("syncLocalChanges drain error:", err));
```
ivanslabbert: I think the while loop in drain() can be removed, and we rely only on localChanges firing a change when an entry is removed from the queue.

The isConnected = false and localChanges.length = 0 early return can also be implemented in this watcher.

ivanslabbert (commented):
Move retry logic into watcher. When going offline, stop the retry cycle, and bypass the failed notification.

shared/src/rest/syncLocalChanges.ts
```diff
@@ -1,72 +1,54 @@
...
+                if (!change) return;
+
+                for (let attempt = 0; attempt < 3; attempt++) {
+                    if (!isConnected.value) return;
+                    if (await send(change)) return;
+                    if (attempt < 2) await new Promise((r) => setTimeout(r, 100));
+                }
+
+                changeReqErrors.value.push(
+                    "Unable to save changes. Please refresh the page and try again.",
```
ivanslabbert:
```suggestion
                    "Unable to submit saved changes. Please refresh the page to try again.",
```

ivanslabbert (approved):
Need to do proper user testing before going to staging.

PR #1628, opened 2026-05-21
Fix: Skip stale deleteCmd when local doc is newer

ivanslabbert:
Guards against an unpublish-then-republish race where a stale deleteCmd arrives in a later sync batch after the republished content, preventing the newer doc from being incorrectly removed.

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>


MrDirkelz (approved):
This looks good to me

PR #1632, opened 2026-05-22
enhancement: add local validation before saving a language

ChrisTouo:


api/src/dto/LanguageDto.ts
```diff
@@ -27,7 +27,7 @@ export class LanguageDto extends _contentBaseDto {
 
     @IsOptional()
     @IsInt()
-    @Min(1)
+    @Min(0)
```
ivanslabbert: Change back to 1 for safety

ivanslabbert (commented):
Setting of default word per minute value in CMS:
This should be triggered when the Word per Minute value is clear or invalid (<1), and the text input loses focus instead of on save.

cms/src/components/languages/EditLanguage.vue
```diff
@@ -425,10 +475,11 @@ const contentActions = computed(() => {
                 <LInput
                     type="number"
                     label="Average reading speed (words per minute)"
+                    :onBlur="handleSpeedBlur"
                     name="averageReadingSpeed"
                     v-model="editable.averageReadingSpeed"
                     class="mb-4 w-full"
-                    placeholder="Enter the average reading speed for people speaking this language ( 200 wpm is the default)"
+                    placeholder="Enter the average reading speed for people speaking this language ( 200 wpm is the default value)"
```
ivanslabbert:
```suggestion
                    placeholder="200"
```

ivanslabbert (approved):
approved with comment

PR #1635, opened 2026-05-25
APP: add a trigger field to indicate if dom has been populated

MrDirkelz:


ivanslabbert (approved):
Remove references to SSG in naming as SSG / ISR is not part of the Luminary main project but can be implemented by implementers by using this event.

Ensure this event is documented in the Readme

PR #1645, opened 2026-05-27
Hybrid query routing + publishDate selective sync

### LINKED ISSUE #1333: Shared, APP: Hybrid Mango queries
split of from #1244 

- [ ] Implement live mango query
- [ ] Implement hybrid mango query
- [ ] Implement hybrid mango queries in APP

ivanslabbert:
## Summary

Adds a **hybrid query module** that routes a Mango query to local IndexedDB when the syncList already covers it and falls back to the `POST /query` API otherwise, plus **`publishDate` as a third selective-sync dimension** (alongside languages and memberOf groups). Also lands the API-side `/query` hardening that the new flexible query path makes worthwhile.

### Part A — publishDate selective sync (`shared/src/rest/sync2`)
- `SyncListEntry`/options carry `publishDateMin/Max`; `OPEN_MIN/MAX` sentinels (MIN/MAX_SAFE_INTEGER) are resolved at `sync()` entry and `initSync()` load, so downstream code sees concrete numbers and legacy entries upgrade in place.
- `syncBatch` injects `selector.publishDate` only when narrowed — byte-identical wire format for callers that don't set a cutoff.
- `_sync` spawns a new column for an uncovered publishDate window; merge is range-aware (horizontal memberOf/language union gated on overlapping ranges); `trim` honours publishDate; DeleteCmd stays publishDate-unscoped so deletes always propagate.
- API sync validator accepts optional `publishDate {$gte,$lte}`.

### Part B — hybridQuery module (`shared/src/util/hybridQuery`)
- `isQueryCovered(query)` coverage check against the syncList.
- `hybridQuery()` one-shot + `useHybridQuery()` reactive composable (local Dexie liveQuery, or remote POST + socket `data` updates). Wired via `initHybridQuery` in `luminary.ts`.
- New API `hybridQuery` validator template; README with flow + column diagrams.

### Performance (low-end device audit)
- Socket merge O(N*M) -> O(N+M) via a lazy `Map<_id,index>` (swap-pop deletes).
- Remote fallback caps at `DEFAULT_REMOTE_QUERY_LIMIT` (500) when no `$limit` is supplied.
- Re-route watcher debounced over a coverage signature (chunkType + EOF counts), so EOF flips re-route and sync bursts coalesce.
- Coverage facts memoized (WeakMap); redundant `$or` re-expansion removed; candidate publishDate ranges precomputed once per check; `arraysEqual` single-pass multiset.
- Remaining Medium-severity notes tracked in `shared/src/util/hybridQuery/performance-notes.md`.

### Security hardening (API `/query`)
- Central policy in the `validateMongoQuery` dispatcher (applies to every identifier): reject `limit > query.maxLimit` (env `QUERY_MAX_LIMIT`, default 500) and reject `$regex`/`$where` anywhere plus `$elemMatch` on any field other than `memberOf`.
- Client-side: coverage warnings log selector keys only (never values); `_resetHybridQueryForTests` gated to test mode.
- Verified non-issue: User-doc retrieval is already blocked at the permission layer (403 before any query runs).

## Commits
- `feat(shared): hybrid query routing + publishDate selective sync`
- `security(query): central limit cap + operator policy on POST /query`

## Testing
- **shared**: hybridQuery + sync2 — 302 pass; full shared suite 959/960 (one pre-existing `syncLocalChanges` flake, unrelated); build clean.
- **api**: `validateMongoQuery` 50/50 (incl. new global-policy tests); full api suite 586/586 (CouchDB + MinIO running).


shared/src/util/hybridQuery/coverage.ts
```diff
@@ -0,0 +1,320 @@
...
+                const got = readLanguages(criteria);
+                if (got) facts.languages = got;
+            } else if (field === "publishDate") {
+                const { min, max } = readPublishDateRange(criteria);
+                facts.publishDateMin = min;
+                facts.publishDateMax = max;
+            }
+            // All other fields (status, slug, expiryDate, …) don't affect coverage.
+        }
+    }
```
MrDirkelz: I feel a switch might read easier at quick glancing over for this perhaps? The else if is quite a big chain. What do you think @ivanslabbert ?

```suggestion
       for (const field in cond) {
    const criteria = (cond as Record<string, unknown>)[field];

    switch (field) {
        case "type": {
            const eq = readEqString(criteria);
            if (eq === undefined) {
                return { uncoverable: "top-level non-equality on 'type'" };
            }
            facts.type = eq as DocType;
            break;
        }
        case "parentType": {
            const eq = readEqString(criteria);
            if (eq !== undefined) facts.parentType = eq as DocType;
            break;
        }
        case "docType": {
            const eq = readEqString(criteria);
            if (eq !== undefined) facts.docType = eq as DocType;
            break;
        }
        case "memberOf": {
            const got = readMemberOf(criteria);
            if (got) facts.memberOfRequired = got;
            break;
        }
        case "language": {
            const got = readLanguages(criteria);
            if (got) facts.languages = got;
            break;
        }
        case "publishDate": {
            const { min, max } = readPublishDateRange(criteria);
            facts.publishDateMin = min;
            facts.publishDateMax = max;
            break;
        }
        // All other fields (status, slug, expiryDate, …) don't affect coverage.
    }
}
```

ivanslabbert: Thanks! This file is however removed.

app/vitest.config.ts
```diff
@@ -31,6 +31,23 @@ export default mergeConfig(
         test: {
             globals: true,
             environment: "jsdom",
+            // HybridQuery exercises the real (un-mocked) mangoToDexie, making the
+            // multi-query page chains heavier than the old mocked path. Under full
+            // parallel load these can exceed vitest's 5s default; give them headroom
+            // so wait-for-expect (not the test timeout) governs.
+            testTimeout: 20000,
+            hookTimeout: 20000,
+            // Cap worker concurrency (~half this host's 8 cores). HybridQuery's real
```
MrDirkelz: "~half this host's 8 cores" could probably be removed

app/vitest.config.ts
```diff
@@ -31,6 +31,23 @@ export default mergeConfig(
...
+            // (un-mocked) mangoToDexie read path is heavier than the old mock, and at
+            // full parallelism the contention occasionally pushed a heavy content-list
+            // test past its timeout (an intermittent flake). Trades a little suite
+            // wall-time for deterministic runs. Set on both pools since vitest's default
+            // is "threads"; a top-level `maxWorkers: "50%"` tripped a tinypool
+            // RangeError on 1.6.1.
```
MrDirkelz: This comment seems specific to the AI agent tracking its plan and not the actual code

app/vitest.setup.ts
```diff
@@ -1,10 +1,23 @@
 import "fake-indexeddb/auto";
-import { RouterLinkStub, config } from "@vue/test-utils";
+import { RouterLinkStub, config, enableAutoUnmount } from "@vue/test-utils";
 import { afterEach, beforeAll, beforeEach, vi } from "vitest";
-import { initConfig, initDatabase } from "luminary-shared";
+
+// Auto-unmount mounted components after each test so their HybridQuery live Dexie
+// subscriptions are disposed. Without this, leaked subscriptions accumulate across
+// a file's tests and re-run real mangoToDexie on every beforeEach mutation, which
+// compounds badly under parallel load (the old mocked mangoToDexie hid this cost).
```
MrDirkelz: "the old mocked mangoToDexie hid this cost" don't think this part of the comment is required perhaps?

cms/src/main.ts
```diff
@@ -44,52 +44,19 @@ async function Startup() {
...
         apiUrl,
+        // CMS editors need the full content set — omit any publishDate cutoff so
+        // sync2 keeps its OPEN_MIN default (full content sync) and HybridQuery's
+        // older-tail supplement never fires.
+        contentPublishDateCutoff: undefined,
+        // What gets synced (and may be persisted to IndexedDB) is owned by sync2
+        // (see src/sync.ts); sync2 joins the socket rooms for those types. This list
+        // is now ONLY the transitional live-only types served by ApiLiveQuery — docs
+        // we display live but never sync. Their rooms are joined at the connect
+        // handshake so ApiLiveQuery receives updates. Remove once the CMS migrates
```
MrDirkelz, line 55: Should I make an issue regarding this? Seems that this has to be removed after the cms migrates them to hybridquery?

shared/src/luminary.ts
```diff
@@ -24,5 +27,22 @@ export async function init(config: SharedConfig) {
 
     // Create HTTP service instance for use in sync operations
     const http = new HttpReq(config.apiUrl || "");
-    initSync(http);
+    // MUST be awaited: initSync's body (db.getSyncList load → validation →
+    // legacy publishDate resolve → findDegenerateChunkTypes reset) has to
+    // complete before consumers register their {immediate:true} sync
+    // watchers — otherwise the first sync() races the degenerate-state
+    // reset and persists fresh entries that re-introduce the broken shape.
```
MrDirkelz: This also seems more like an AI comment to note its thjinking progress for its plan given to it

MrDirkelz (commented):
Just some comments. But looks well thought out and implemented. Also some tests failing

MrDirkelz (commented):
Looks good. Performance feels better than last time so ghost placeholder may maybe not be required here anymore. Just some tests that are failing

MrDirkelz (approved):

PR #1658, opened 2026-05-29
APP: Rethink desktop profile menu to use mobile-style slide-in panel

MrDirkelz:
The desktop profile menu opened a small dropdown that diverged from the mobile experience. It now opens the same slide-in side panel as mobile — a rich header (avatar, name, email) plus the full navigation list — so both platforms share one presentation.

- ProfileMenu: drop the DropdownMenu dropdown; both triggers now open the shared MobileSidebar panel. Desktop trigger is the avatar + chevron in the top-right (the conventional account-menu control, no redundant name/email); mobile keeps its labelled "Menu" tab in the bottom bar.
- MobileSidebar: render via LTeleport instead of a native Teleport, so it teleports to body in production but renders inline in tests — consistent with BaseModal and makes the panel testable.
- Tests: select menu items by label instead of fragile indices, open the panel to assert the signed-in user's name (it now lives in the header), and assert the profile-menu trigger is present when logged out.


app/src/components/common/DropdownMenu.vue
ivanslabbert, line 1: Rename component as it is not a dropdown menu anymore

ivanslabbert (approved):
Approved with comment

PR #1686, opened 2026-06-08
APP: Research and implement Vite SSG

MrDirkelz:


app/src/ssg/NEXT-STEPS.md
MrDirkelz, line 1: This may be out of date now, so can probably be removed

app/src/components/content/ContentTile.vue
```diff
@@ -36,7 +36,9 @@ const publishDateText = computed(() => {
     ) {
         return "";
     }
-    return db.toDateTime(props.content.publishDate).toLocaleString(DateTime.DATETIME_MED);
+    return DateTime.fromMillis(props.content.publishDate)
+        .setLocale(typeof navigator !== "undefined" ? navigator.language || "en-US" : "en-US")
```
ivanslabbert: Check if we can use the CMS configured default language here?

app/src/components/navigation/SearchModal.vue
```diff
@@ -521,6 +522,18 @@ const goToResult = (result: EnrichedResult) => {
 let handleGlobalKeydown: ((event: KeyboardEvent) => void) | null = null;
 
 onMounted(() => {
+    const applyExploreQuery = (value: unknown) => {
+        // Public structured-data contract: /explore?q=<URL-encoded query>.
```
ivanslabbert: Use a different route than "explore" which is used for the explore page. Explain why we need this route with a URL query-string, and how it is used / called from other locations in the codebase / external callers.

This might also need a dedicated section in a Readme file / it's own Readme file.

app/src/components/navigation/TopBar.vue
```diff
@@ -25,6 +26,13 @@ withDefaults(defineProps<Props>(), {
 });
 
 const router = useRouter();
+
+// On web/SSG the logo + nav links (DesktopMenu) ARE prerendered (crawlable), but
+// the per-user ProfileMenu (avatar/language/auth → Dexie) is rendered only after
+// mount so it doesn't crash during the Node prerender and the signed-out shell
+// matches the first client render. Native renders it immediately (unchanged).
+const showProfile = useHydrated();
+
```
ivanslabbert: Check - the topbar menu was removed in favour of the new sidebar menu

app/src/components/BasePage.vue
```diff
@@ -10,9 +10,28 @@ import type { ContentDto } from "luminary-shared";
...
+// On the web/SSG build the desktop sidebar IS prerendered — its nav / logo /
+// theme+language controls are public, the profile block self-gates on auth (false
+// during the prerender, so it matches the first client render), and its
+// interactive/Dexie-backed overlays mount client-side (see DesktopSidebar). What
+// still cannot exist during the Node prerender is the per-user notification chrome
+// (synced Dexie data) + the toast Teleport — those stay gated to AFTER mount so
+// the prerendered HTML and the first client render match (clean hydration). On
+// native showChrome is always true → behaviour unchanged.
+const isWeb = import.meta.env.VITE_BUILD_TARGET === "web";
+const showChrome = useHydrated();
```
ivanslabbert: unused variable

app/src/components/BasePage.vue
```diff
@@ -13,6 +13,23 @@ import { getRouteHistory } from "@/router";
...
 
+// On the web/SSG build the desktop sidebar IS prerendered — its nav / logo /
+// theme+language controls are public, the profile block self-gates on auth (false
+// during the prerender, so it matches the first client render), and its
+// interactive/Dexie-backed overlays mount client-side (see DesktopSidebar). What
+// still cannot exist during the Node prerender is the per-user notification chrome
+// (synced Dexie data) + the toast Teleport — those stay gated behind
+// `notificationsReady` so the prerendered HTML and the first client render match
+// (clean hydration). On native, `notificationsReady` starts true → unchanged.
+const isWeb = import.meta.env.VITE_BUILD_TARGET === "web";
```
ivanslabbert: As this is used purely for SSG builds, should we rather call it "isSsg"?

app/src/composables/useBucketInfo.ts
```diff
@@ -1,39 +1,57 @@
...
+    useHybridQuery,
+    type Uuid,
+    queryRemote,
+    structuralCacheKey,
+    writeResponseCache,
+} from "luminary-shared";
+
+// Storage is fully-synced public reference data. The query shape is constant, so a
+// fixed `cacheId` keeps its response-cache entry distinct from any same-shaped query.
+// Unlike useContentQuery's cacheId, this is NOT auth-scoped — bucket/CDN
```
ivanslabbert: This sentence is not completely correct. The storageDto objects are still auth-scoped, although typically assigned with public access

app/src/composables/useContentQuery.spec.ts
ivanslabbert: unused imports

ivanslabbert, line 1: Add tests for SSG render settings

app/src/composables/useContentQuery.ts
```diff
@@ -93,23 +123,67 @@ export function useContentQuery(
...
+        // entry (see the SSR branch below), so a returning logged-in client must
+        // read/write a different `:auth` entry rather than painting the public seed
+        // or overwriting it with personalized data (see useContentQuery.spec.ts).
+        cacheId: `${rest.cacheId ?? ""}:${hasPersistedSession() ? "auth" : "anon"}`,
+    };
+
+    // --- Web/SSG PRERENDER (Node) only. The browser client + native both fall through
+    // to the identical hybrid query below; on the client `cache: true` seeds the first
+    // render synchronously from the response cache this branch primed at build time. ---
+    if (import.meta.env.SSR) {
```
ivanslabbert: Add tests for this code path (as requested in previous comment)

app/src/composables/useHydrated.ts
```diff
@@ -0,0 +1,17 @@
+import { computed, onMounted, ref } from "vue";
+
+/**
+ * True immediately in the native SPA and after the first client mount in the
+ * web/SSG build. Use it for client-only UI that must not participate in SSR
+ * hydration.
+ */
+export function useHydrated() {
+    const isWeb = import.meta.env.VITE_BUILD_TARGET === "web";
```
ivanslabbert: See suggestion on naming above

app/src/pages/SingleContent/SingleContent.vue
```diff
@@ -168,9 +162,11 @@ function routeRedirect(redirect: RedirectDto): boolean {
     return true;
 }
 
-// Loading until the content query produces an answer for the current slug.
-const isLoading = ref(true);
-const is404 = ref(false);
```
ivanslabbert, line 173: We still need to handle a "404" response when content is not found in a hydrated app. A hydrated app will never fetch from the CDN, but still needs to display the "content not found" error message to the user.

app/src/pages/SingleContent/SingleContent.vue
```diff
@@ -199,142 +195,194 @@ const scheduleNotFound = () => {
...
     () =>
         content.value?.parentId
             ? [{ parentId: content.value.parentId }]
             : [{ parentId: { $in: [] } }],
     {
         publishedFilter: false,
-        // Seek siblings by parentId rather than scanning the publishDate index. The
-        // provably-empty `$in: []` branch short-circuits before any API call, so the
-        // index/sort only matter on the resolved-parentId (equality) branch.
+        cache: true,
```
ivanslabbert, line 307: Not needed to set cache:true for the related translations. We do not want to persist them offline. Also, cache:true should not be set when stripping fts fields

app/src/pages/SingleContent/SingleContent.vue
```diff
@@ -368,30 +416,46 @@ const isBookmarked = computed(() => {
...
-        metaTag = document.createElement("meta");
-        metaTag.setAttribute("name", "description");
-        document.head.appendChild(metaTag);
-    }
-    // Update the content attribute
-    metaTag.setAttribute("content", content.value?.seoString || content.value?.summary || "");
-});
+// Native sets the title/description imperatively (it has no @unhead plugin). The web
+// build's `useHead` above owns the head there (and serializes it into the prerendered
+// HTML), so skip this path on web.
```
ivanslabbert: See previous comment on SSR build naming. Avoid ambiguous naming "web"

app/src/pages/SingleContent/SingleContent.vue
```diff
@@ -368,30 +416,46 @@ const isBookmarked = computed(() => {
...
+// and so it never reaches for `db` (absent during the Node prerender).
+const formatPublishDate = (ms: number) =>
+    DateTime.fromMillis(ms)
+        .setLocale(
+            (typeof navigator !== "undefined" && navigator.language) ||
+                cmsDefaultLanguage.value?.languageCode ||
+                "en-US",
+        )
+        .toLocaleString(DateTime.DATETIME_MED);
+
```
ivanslabbert: Format the date in the locale of the selected language, falling back to en-US

app/src/pages/SingleContent/SingleContent.vue
```diff
@@ -740,15 +657,15 @@ watch([isLoading, content, is404], async () => {
         >
             <div
                 class="flex flex-grow justify-center"
-                :class="{ 'items-center': isLoading }"
+                :class="{ 'items-center': isLoading && !content }"
             >
                 <LoadingBar
-                    v-if="isLoading"
+                    v-if="isLoading && !content"
                     label="Loading..."
```
ivanslabbert: add i18n support

app/src/pages/HomePage.vue
```diff
@@ -21,16 +22,31 @@ async function checkReady() {
 }
 
 onActivated(checkReady);
+useLocalizedStaticHead("/");
```
ivanslabbert, line 25: We need to include a site description with our home page SEO data. This is not currently included in Luminary, so we need to find a good way to add it (separate follow-up PR?)

app/src/pages/VideoPage.vue
```diff
@@ -17,15 +18,26 @@ async function checkReady() {
 }
 
 onActivated(checkReady);
+useLocalizedStaticHead("/watch");
```
ivanslabbert, line 21: See comment on HomePage.vue

app/src/pages/ExplorePage.vue
```diff
@@ -17,15 +18,26 @@ async function checkReady() {
 }
 
 onActivated(checkReady);
+useLocalizedStaticHead("/explore");
```
ivanslabbert, line 21: See comment on HomePage.vue

app/src/router/localizedRoutes.ts
ivanslabbert, line 1: Add "search" as discussed in previous comments

app/src/seo/contentHead.ts
```diff
@@ -0,0 +1,307 @@
...
+
+export type PublicTaxonomy = { name: string; url: string };
+
+type SocialImage = {
+    url: string;
+    width: number;
+    height: number;
+};
+
+const staticPageCopy = {
```
ivanslabbert, line 26: Add Search

app/src/seo/contentHead.ts
```diff
@@ -0,0 +1,307 @@
...
+            item: item.item,
+        })),
+    };
+}
+
+/**
+ * Social crawlers do not support responsive image source sets. Pick the largest
+ * original file so the single OG/Twitter URL is suitable for every preview.
+ */
+export function primaryArticleImage(
```
ivanslabbert, line 172: Do not return the largest image, but find a suitable size for thumbnail display

app/src/seo/json-ld.md
```diff
@@ -0,0 +1,77 @@
...
+| `articleSection` | Supplies topical context beyond the headline, improving retrieval and answer grounding. | The content's `parentTags`, resolved at prerender time to published category content documents and their user-facing names. Emit an array when the content belongs to multiple categories. |
+
+Do not emit empty strings, placeholder URLs, unpublished tags, or image values
+when their underlying public data is unavailable. JSON-LD should describe the
+prerendered public page, not internal identifiers or editor-only metadata.
+
+## Site-wide structured data still to add
+
+These objects are not article-specific and should be emitted once per relevant
+page through the same centralized SEO helper.
```
ivanslabbert: Check if it is added already

app/src/seo/json-ld.md
```diff
@@ -0,0 +1,77 @@
...
+### `WebSite` and `SearchAction` on `/`
+
+Emit a `WebSite` object on the home page with the canonical site URL and name.
+Add a `potentialAction` `SearchAction` that points to the public search URL and
+uses the anonymous query parameter accepted by the app:
+
+```json
+{
+  "@type": "SearchAction",
+  "target": "https://example.org/explore?q={search_term_string}",
```
ivanslabbert: change to /search as mentioned in previous comments

app/src/seo/publicSite.ts
```diff
@@ -0,0 +1,57 @@
...
+    // `q` is intentionally limited to the anonymous, public FTS overlay on
+    // /explore. Values are encoded with URLSearchParams / encodeURIComponent.
+    return {
+        "@context": "https://schema.org",
+        "@type": "WebSite",
+        name: organizationName,
+        url,
+        potentialAction: {
+            "@type": "SearchAction",
+            target: `${canonicalUrl("/explore")}?q={search_term_string}`,
```
ivanslabbert: change to /search as discussed above

shared/src/ssg/facetKeys.ts
ivanslabbert, line 1: Segregate ssg-doc-facets.json into smaller files (perhaps a hierarchy that is easily walkable) to prevent a heavy in-memory only object as the site is is growing.

ivanslabbert (commented):
To avoid json-ld "search trickery", migrate the search modal to a full-blown search page. This page is used on mobile when selecting the search button from the bottom menu. On desktop we still use the search modal, but use the same component as embedded on the dedicated search page. when a user would add /search to the URL, it will open the dedicated search page on desktop or mobile, but on Desktop we do not have a visible route to the page, as the modal suffices.

Add a field in EditContent to select if the author is a person or organization. This will drive the jsonLD Author Type field

window.__INITIAL_STATE__ = '{"renderLang":"lang-eng",...
The renderLang should show the language name and not the language ID (which in many cases will be a non-interpretable UUID)

Logo: Check if possible to decide on the logo variant (small vs big) in pure CSS, which will prevent flicker on hydration.

404 error: check if still working correctly in an hydrated app

Render a 404 page to be used as default 404 error page (hydratable, to be uploaded to S3 storage, and used as the CDN default error page). Check which other error codes we also need to support with custom error pages.

Ensure SSG functions are well tested

SSR / SSG / ISR Use consistent naming. e.g. if we are using SSR to render for both SSG and ISR, we can just name it SSG.

Back button in SingleContent is not working when JS is disabled

SingleContent: Setting of localstorage items should be excluded in the SSR rendered version. On Hydration the hydrated code should handle this. It is currently included as a separate script in the SSR rendered page.

og:locale currently appends a hard-coded country code to the language code. Rather set to the language code, and if the language code includes a country code (e.g. en_UK) it will automatically be locale-correct.

app/src/ssg/README.md
```diff
@@ -0,0 +1,249 @@
...
+
+The web build is driven from a **separate deployment repo** (this repo is a submodule of
+it); that repo owns deployment and incremental-regeneration orchestration.
+
+---
+
+## The goal
+
+Luminary is an offline-first PWA; the native shells (Capacitor) cover logged-in, offline
+use. But the app was **invisible to search engines and link previews** — everything
```
ivanslabbert: This is not fully correct. Logged-in and offline use is also possible with the web / PWA client

shared/src/ssg/redirectHtml.ts
ivanslabbert, line 1: Include the correct response code in the redirect (as available on the redirectDto)

shared/src/ssg/routeIndex.ts
ivanslabbert, line 1: Comment to explain why this file is here

ivanslabbert: See above comment on splitting the file in smaller chunks

app/src/App.vue
```diff
@@ -115,9 +116,25 @@ const routeKey = computed(() => {
     return router.currentRoute.value.fullPath;
 });
 
+// On the web/SSG tier the prerendered HTML is the signed-out, content-first
+// baseline: the interactive/auth-aware chrome (modals, audio player) is
+// rendered only AFTER mount so the first client render matches the SSR output
+// (clean hydration). On native/SPA there is no prerender, so chrome renders
+// immediately as before — behaviour is unchanged there.
+const showChrome = useHydrated();
```
ivanslabbert: Naming a bit ambigious

app/src/main.web.ts
ivanslabbert, line 1: Check naming. Using Web and SSG interchangeably is a bit confusing, an normal "web" users will be using an hydrated version of the website

app/.env.example
```diff
@@ -23,7 +23,16 @@ VITE_COPYRIGHT_ID="page-copyright"
...
+# path is resolved against VITE_WEB_ORIGIN and must be publicly accessible.
+VITE_PUBLIC_ORGANIZATION_NAME="Luminary"
+VITE_PUBLIC_LOGO_PATH="/logo.svg"
+
 VITE_SENTRY_DSN=
 
 
-VITE_BYPASS_MINIFY=false
\ No newline at end of file
+VITE_BYPASS_MINIFY=false
```
ivanslabbert, line 40: Add note that this is an optional key

app/.gitignore
ivanslabbert, line 1: Add ignores for SSG generated files, including facetkey files needed for SSG, html & JS files generated in root folder, etc. The two added entries are probably covering this already, but just double check to prevent repo pollution

app/package.json
```diff
@@ -50,6 +53,7 @@
         "@types/m3u8-parser": "^7.2.2",
         "@types/node": "^18.19.3",
         "@types/videojs-mobile-ui": "^0.8.3",
+        "@unhead/vue": "^1.11.20",
```
ivanslabbert, line 56: Check if still used? Think there is a comment in SingleContent that unHead is not used?

app/vite.config.ts
```diff
@@ -10,9 +10,30 @@ import { buildTargetVirtuals } from "./vite-plugins/buildTargetVirtuals";
 
 const env = loadEnv("", process.cwd());
 
+const blockGeneratedOutput = (): Plugin => ({
```
ivanslabbert, line 16: Add comment

ivanslabbert (commented):
Could we, instead of storing all data in hqcache and window initial state (duplicate to the pre-rendered HTML) rather use a cheap hash as fingerprint to determine if we need a repaint or not? Not sure if I understand the mechanics well enough, but we are including a lot of text in duplicate in the SSG rendered HTML file now

app/src/ssg/NEXT-STEPS.md
MrDirkelz: Removed — this doc no longer exists on this branch.

app/src/components/BasePage.vue
MrDirkelz: Already renamed (`isWeb`/`IS_WEB` → `isSSG`/`IS_SSG`, commit 42ca0f77) before this review landed.

app/src/composables/useBucketInfo.ts
MrDirkelz: Fixed — the comment now says storage docs go through the same ACL system and aren't exempt from auth scoping, they just happen to always be assigned public access in practice, which is what actually makes the single shared cache entry safe.

app/src/composables/useContentQuery.spec.ts
MrDirkelz: No unused imports in the current file — both top-level imports (`vitest`, `useContentQuery`) are used. Must have been cleaned up in a later revision of this branch.

MrDirkelz: Done — `useContentQuery.ssr.spec.ts` now covers the SSR/prerender branch end-to-end (queryRemote fetch, stripFields on the live output, response-cache write scoped `:anon`, ssrCacheStripFields vs cacheStripFields precedence, dependency-key reporting, and the ssrChain ordering).

app/src/composables/useContentQuery.ts
MrDirkelz: Done — see `useContentQuery.ssr.spec.ts`, which exercises exactly this `import.meta.env.SSR` branch.

app/src/composables/useHydrated.ts
MrDirkelz: Already renamed to `isSSG` (matches BasePage.vue's naming) before this review landed.

app/src/pages/SingleContent/SingleContent.vue
MrDirkelz: Already fixed (commit 08b24002) — not-found resolution now waits for the query to genuinely settle (`isContentFetching` false, with a redirect-probe fallback) instead of a fixed wall-clock guess, and applies identically on any client render (hydrated web or native), not just the initial CDN-served static page. Covered by `SingleContent.404.spec.ts`.

app/src/pages/SingleContent/SingleContent.vue
MrDirkelz: Fixed — added `persistOffline: false` to the translations query so these sibling-translation docs (language dropdown / hreflang only) no longer accumulate in the offline IndexedDB store. Kept `cache: true` though: that's the localStorage response-cache seed, unrelated to offline persistence, and it's what keeps the SSR-rendered translations list matching the client's first paint (no hydration mismatch). stripFields/cacheStripFields were already correct — they've never touched IndexedDB persistence, only the live-output heap and the response-cache write.

app/src/pages/SingleContent/SingleContent.vue
MrDirkelz: This one's intentional, not a mixup — see `ssg/README.md`'s "Naming convention" section: `web` is the deliberate product/build-target name (`VITE_BUILD_TARGET === "web"`), while `Ssg`/`SSR` name the build-time mechanism. Commit 63d11b22 added clarifying comments (here and in `main.web.ts`) explaining that a normal web visitor only ever sees the hydrated result, not the SSG output itself, rather than renaming away from `web`.

app/src/pages/SingleContent/SingleContent.vue
MrDirkelz: Fixed — `formatPublishDate` now uses `selectedLanguageCode` (the translation actually being displayed), falling back to en-US, instead of `navigator.language` (the visitor's browser locale, which has no relationship to the reading language). Added a regression test in `SingleContent.spec.ts` (French content renders "janv." not "Jan 1, 2024").

app/src/pages/SingleContent/SingleContent.vue
MrDirkelz: Fixed — the hardcoded `label="Loading..."` now goes through `t('singlecontent.loading')`, with the key added to the English seed.

app/src/pages/HomePage.vue
MrDirkelz: Turns out this was already solved via the existing i18n mechanism rather than needing a new Luminary data field: `useLocalizedStaticHead` sets `<meta name="description">` from `t(copy.description)`, and `home.description` ("Discover the latest articles, videos, and audio.") is already in the English seed, editable by translators/admins like any other UI string — same path as explore/watch below. No follow-up PR needed.

app/src/pages/VideoPage.vue
MrDirkelz: Same answer as HomePage.vue — already resolved via the `watch.description` i18n key + `useLocalizedStaticHead`.

app/src/pages/ExplorePage.vue
MrDirkelz: Same answer as HomePage.vue — already resolved via the `explore.description` i18n key + `useLocalizedStaticHead`.

app/src/router/localizedRoutes.ts
MrDirkelz: Already done — `/search` (and localized `/<code>/search`) is registered in both `routes.ts` and `localizedRoutes.ts`, backed by a dedicated `SearchPage.vue`, and the bottom-nav/desktop search now routes there instead of the old `/explore?q=` trickery.

app/src/seo/contentHead.ts
MrDirkelz: Already done — `staticPageCopy` includes `"/search"` with its own title/description keys, served through `useLocalizedStaticHead`.

app/src/seo/contentHead.ts
MrDirkelz: Fixed — `primaryArticleImage` now picks the processed size closest to the OG/Twitter recommended width (1200), not the largest original. With the real presets `[180, 360, 640, 1280, 2560]` that resolves to 1280 instead of the full 2560 original. Added `contentHead.spec.ts` covering the selection logic.

app/src/seo/json-ld.md
MrDirkelz: Checked — every item on this doc's checklist (Article `publisher`/`image`/`mainEntityOfPage`/`wordCount`/`articleSection`, `BreadcrumbList`, `WebSite`+`SearchAction`, the `/search` route) is already implemented and wired up in `contentHead.ts`/`publicSite.ts`/`SingleContent.vue`. Removed the doc as a stale TODO rather than leaving a checklist of already-shipped work.

app/src/seo/json-ld.md
MrDirkelz: Already done — removed this doc entirely (see the sibling thread on line 30): the `/search` route it was asking for already exists and is wired into `websiteJsonLd()`'s `SearchAction` target.

app/src/seo/publicSite.ts
MrDirkelz: Already done — `websiteJsonLd()`'s `SearchAction.target` points at `${canonicalUrl("/search")}?q={search_term_string}`, matching the dedicated `/search` page.

shared/src/ssg/facetKeys.ts
MrDirkelz: Done — `docFacetShards.ts` shards `ssg-doc-facets.json` into `ssg-doc-facets/<shard>.json` via a fixed-count fnv1a32-mod-shardCount hash bucket (constant regardless of site size), so a scoped rebuild only touches the shards its changed docs land in.

app/src/ssg/README.md
MrDirkelz: Fixed — reworded to say the existing app (both the browser-installed PWA and the native Capacitor shells) already covers logged-in/offline use via its service worker + local-first sync, not just the native shells.

shared/src/ssg/redirectHtml.ts
MrDirkelz: Already fixed (commit 70ef6b8d) — `redirectStatus()` derives the real 301/302 from the redirect's `redirectType`, baked into both the `x-redirect-status` meta tag on the static artifact and the `ssg-redirect-index.json` sidecar.

shared/src/ssg/routeIndex.ts
MrDirkelz: Addressed both parts: expanded the top comment to explain why this sidecar exists (a DeleteCmd only carries an id, never the slug it was rendered to) and its relationship to sharding, and added `routeIndexShards.ts` — the same fnv1a32-mod-shardCount scheme as `docFacetShards.ts` — so `vite.config.web.ts` now writes `ssg-route-index/<shard>.json` instead of one growing file. `resolveContentDelete` is unchanged and shard-agnostic (content ids and parent ids are each sharded by their own id, so a consumer resolving one DeleteCmd loads a single shard file).

app/src/main.web.ts
MrDirkelz: Same answer as the SingleContent.vue naming thread — `web` is the deliberate product/build-target name (see `ssg/README.md`'s naming convention), kept and clarified via comments (commit 63d11b22) rather than renamed, since `main.web.ts` genuinely does cover both the SSG prerender phase and the eventual hydrated-client phase under one build-target name.

app/.env.example
MrDirkelz: Added — noted it's optional (debugging aid only; any value other than "true", or unset, minifies normally).

app/.gitignore
MrDirkelz: Double-checked: every write in `vite.config.web.ts` goes through `OUT_DIR`/`dist-web` (or its subdirectories) — nothing is generated at the repo root. `dist-web` + `.vite-ssg-temp` in `.gitignore` fully cover it, no changes needed.

app/package.json
MrDirkelz: Still used — `@unhead/vue`'s `useHead` is called from `contentHead.ts`, and `main.web.ts` installs the plugin (`useHead: true`) for the web/SSG build. Not dead weight.

app/vite.config.ts
MrDirkelz: Already fixed (commit 63d11b22, same pass that clarified the web/SSG naming) — `blockGeneratedOutput` now has a 2-line comment explaining why it 404s dev-server requests for `/dist`/`/dist-web`.

api/src/db/seedingDocs/lang-fra.json
ivanslabbert, line 1: Update French to match changes to English

api/src/dto/_contentParentDto.ts
```diff
@@ -54,4 +54,9 @@ export class _contentParentDto extends _contentBaseDto {
     @IsString()
     @Expose()
     mediaBucketId?: string; // S3 bucket ID for media storage
+
+    @IsOptional()
+    @IsNumber()
+    @Expose()
+    authorType?: number; // Drives jsonLD author @type: 1 = Organization, 0/undefined = Person
```
ivanslabbert: I think it would be cleaner through the rest of the codbase to rather use a predefined string type instead of a number, that will need to be translated back to string everywhere it is used
```suggestion
    authorType?: "person" | "org"; // Drives jsonLD author 
```

api/src/dto/ContentDto.ts
```diff
@@ -155,6 +155,11 @@ export class ContentDto extends _contentBaseDto {
     @Expose()
     parentUseVerticalTileLayout?: boolean;
 
+    @IsOptional() // Optional as it is set upon change request processing
+    @IsNumber()
+    @Expose()
+    parentAuthorType?: number; // Mirrors parent authorType: 1 = Organization, 0/undefined = Person
```
ivanslabbert: See above comment

app/src/router/localizedRoutes.ts
```diff
@@ -0,0 +1,36 @@
+import type { RouteRecordRaw } from "vue-router";
+
+const HomePage = () => import("@/pages/HomePage.vue");
+const ExplorePage = () => import("@/pages/ExplorePage.vue");
+const SearchPage = () => import("@/pages/SearchPage.vue");
+const VideoPage = () => import("@/pages/VideoPage.vue");
+
+export function localizedStaticRoutes(langCodes: string[], defaultCode: string): RouteRecordRaw[] {
```
ivanslabbert: Add JSDOC comment to this function

app/src/seo/contentHead.ts
```diff
@@ -0,0 +1,345 @@
...
+}
+
+export function languageCodeForContent(
+    languageId: string | undefined,
+    languages: Pick<LanguageDto, "_id" | "languageCode">[],
+): string {
+    return (
+        (languageId && languages.find((language) => language._id === languageId)?.languageCode) ||
+        (languageId || "").replace(/^lang-/, "") ||
+        "en"
```
ivanslabbert: This code is not applicable. We should never use the language ID for the code (even as fallback). Language codes MUST be correctly defined in the CMS. To be implemented in #1859

app/src/ssg/queryDrain.ts
```diff
@@ -0,0 +1,101 @@
+/**
+ * Route enumeration for the SSG build: drains a doc type's full set from `/query`
+ * via keyset (not offset) pagination. Offset pagination over a live, mutating
+ * collection is non-deterministic — a doc changing mid-drain shifts later pages
+ * and silently drops rows (this once capped a build at ~101 of ~1934 routes).
```
ivanslabbert: remove domain specific data

shared/src/ssg/routeIndex.ts
ivanslabbert, line 1: Instead of having to keep track of deleted content / redirect document's slugs, the deleteCmd document must rather contain the slug of the document deleted. This will help keep a single source of truth, avoinding the risk of having documents "orphaned" in the build output if a delete was missed due to tracking getting out of sync with the database "truth".

conversation-authortype-jsonld.md
ivanslabbert, line 1: This should not be committed to the project. Add it as short note to README where / as applicable. Also applies to the following conversation .md files

ssg-standup.html
ivanslabbert, line 1: No HTML as documentation please. Should be in .md format

app/src/pages/SingleContent/SingleContent.vue
```diff
@@ -168,173 +182,234 @@ function routeRedirect(redirect: RedirectDto): boolean {
...
             : [{ parentId: { $in: [] } }],
     {
         publishedFilter: false,
-        // Seek siblings by parentId rather than scanning the publishDate index. The
-        // provably-empty `$in: []` branch short-circuits before any API call, so the
-        // index/sort only matter on the resolved-parentId (equality) branch.
+        cache: true,
+        // Per-slug discriminator so the per-document cache is SAFE — a shape-only key
+        // would seed this page from a previously-viewed post's translations.
+        cacheId: `translations:${props.slug}`,
```
ivanslabbert: Only set the cache ID when isSSG is true, as SSG depends on this query cache entry. Using a cache ID indefinitely does not benefit normal use, and will accumulate a lot of cache entries, which only will speed up the query on article reload.

Add a note on why we need a unique cache ID for SSG

app/src/pages/SingleContent/SingleContent.vue
```diff
@@ -84,27 +84,61 @@ const props = defineProps<Props>();
...
-    publishDate: 0,
-    parentTags: [],
-};
+// Content by slug — the SAME local-first hybrid query on every build (web and
+// native). On the web build the prerender primes this query's response cache
+// (`cacheId = slug` makes the per-document seed safe), so the hydrating client shows
+// the article on first paint with no flash; during the Node prerender the seam's
+// onServerPrefetch fetches it so it is present in the static HTML. `text` (the body,
+// rendered below) and `memberOf` (read by canEdit) are kept off the default strip set.
+const { output: contentArr, isFetching: isContentFetching } = useContentQueryWithState(
```
ivanslabbert, line 94: We want to persist the response offline for the content document opened in SingleContent. In order to surface in offline FTS searches, stripFields should not be set.

app/src/pages/SingleContent/SingleContent.vue
```diff
@@ -84,27 +84,61 @@ const props = defineProps<Props>();
...
+        cacheId: props.slug,
+        // Seek the single slug doc via the slug-led index. The publishDate sort is required
+        // for CouchDB to engage the index (slug eq alone falls back to a full scan).
+        useIndex: "content-slug-publishDate-index",
+        sort: [{ publishDate: "desc" }],
+        stripFields: ["fts", "ftsTokenCount", "_rev"],
+        // `text` is the heaviest field on the page and already sits in the prerendered
+        // `[data-ssr-article-text]` node — omit it from the SSR-authored cache write so it
+        // isn't shipped twice; the hydration patch below recovers it from that node.
+        ssrCacheStripFields: ["text"],
```
ivanslabbert:
```suggestion
        ssrCacheStripFields: ["text", "fts", "ftsTokenCount","_rev"],
```

app/src/pages/SingleContent/SingleContent.vue
```diff
@@ -84,27 +84,61 @@ const props = defineProps<Props>();
...
+// the article on first paint with no flash; during the Node prerender the seam's
+// onServerPrefetch fetches it so it is present in the static HTML. `text` (the body,
+// rendered below) and `memberOf` (read by canEdit) are kept off the default strip set.
+const { output: contentArr, isFetching: isContentFetching } = useContentQueryWithState(
+    () => [{ slug: props.slug }],
+    {
+        includeScheduled: false,
+        languageFilter: false,
+        cache: true,
+        cacheId: props.slug,
```
ivanslabbert: Only set the cacheId when isSSR is true, to prevent a dedicated query cache entry for each slug

api/src/db/db.service.spec.ts
ivanslabbert, line 1: Add test to ensure the slug is set when deleting a content doc

app/src/router/localizedRoutes.ts
ivanslabbert, line 1: Setup language prefix for the default language as well, which will avoid issues should the default language be changed

app/src/seo/contentHead.ts
```diff
@@ -0,0 +1,344 @@
...
+    const sections = [...new Set((options.taxonomy ?? []).map((tag) => tag.name).filter(Boolean))];
+    const wordCount = typeof c.wordCount === "number" ? c.wordCount : fallbackWordCount(c.text);
+    const publisher = publisherJsonLd();
+    return {
+        "@context": "https://schema.org",
+        "@type": "Article",
+        headline: c.title,
+        description,
+        author: c.author
+            ? { "@type": c.parentAuthorType ? "Organization" : "Person", name: c.author }
```
ivanslabbert:
```suggestion
            ? { "@type": c.parentAuthorType === "org" ? "Organization" : "Person", name: c.author }
```

shared/src/types/dto.ts
```diff
@@ -106,6 +110,8 @@ export type ContentDto = ContentBaseDto & {
     parentAlwaysOffline?: boolean;
     parentPinned?: number;
     parentUseVerticalTileLayout?: boolean;
+    /** Drives the jsonLD author @type: 1 = Organization, 0/undefined = Person. Mirrors parent authorType. */
+    parentAuthorType?: number;
```
ivanslabbert: Change to string?

shared/src/types/dto.ts
```diff
@@ -128,6 +134,8 @@ export type ContentParentDto = ContentBaseDto & {
     media?: MediaDto;
     mediaBucketId?: Uuid;
     useVerticalTileLayout?: boolean;
+    /** Drives the jsonLD author @type: 1 = Organization, 0/undefined = Person (default). */
+    authorType?: number;
```
ivanslabbert: Change to string?

api/src/db/seedingDocs/lang-fra.json
```diff
@@ -149,6 +154,7 @@
         "openapp_warning.button_cancel": "Annuler",
         "auth.sign_in": "Se connecter",
         "auth.no_methods_available": "Aucun méthode d'authentification disponible. Veuillez réessayer dans un moment.",
-        "singlecontent.listen": "Écouter"
+        "singlecontent.listen": "Écouter",
```
johan-bell:
```suggestion
        "singlecontent.listen": "Écoutez",
```

api/src/db/seedingDocs/lang-fra.json
```diff
@@ -54,10 +54,15 @@
         "settings.data_saver.slow_connection_note": "Votre connexion semble lente. L'utilisation réduite des données sera donc appliquée automatiquement.",
         "home.title": "Accueil",
         "home.continue": "Continuer",
+        "home.description": "Découvrez les derniers articles, vidéos et contenus audio.",
         "home.continueListening": "Continuer à écouter",
         "explore.title": "Explore",
+        "explore.description": "Explorez les sujets et découvrez du contenu.",
         "explore.other": "Autre",
+        "search.title": "Rechercher",
+        "search.description": "Recherchez des articles, vidéos et bien plus dans toute la bibliothèque.",
```
johan-bell:
```suggestion
        "search.description": "Recherchez des articles, vidéos et bien plus dans la bibliothèque.",
```

johan-bell (commented):
i18n comments

api/src/changeRequests/documentProcessing/processPostTagDto.spec.ts
```diff
@@ -253,10 +258,14 @@ describe("processPostTagDto", () => {
         const deleteCommands = await db.getDocsByType(DocType.DeleteCmd);
         expect(deleteCommands.docs.find((d) => d.docId == "post-blog3")).toBeDefined();
         expect(deleteCommands.docs.find((d) => d.docId == "post-blog3").docType).toBe(DocType.Post);
+        expect(deleteCommands.docs.find((d) => d.docId == "post-blog3").slug).toBeUndefined();
         expect(deleteCommands.docs.find((d) => d.docId == "content-en")).toBeDefined();
         expect(deleteCommands.docs.find((d) => d.docId == "content-en").docType).toBe(
             DocType.Post, // This is needed as the permission system does not include Content documents, but bases permissions on the parent type (Post / Tag).
         );
+        expect(deleteCommands.docs.find((d) => d.docId == "content-en").slug).toBe(
```
ivanslabbert: Test should be in db.service.spec.ts

ivanslabbert (commented):
To safely handle slug changes (deleting SSG generated files), we need an array of previous slugs on content and redirect DTO's. This field should be optional, and when there are no previous slugs, the key should not be set.

ivanslabbert (commented):
In order to safely process delete commands, we should generate delete command documents (perhaps in a separate folder) which the file uploader can process and delete files from the final web storage (e.g. when uploading to S3 storage). The uploader should not listen to the DB change feed (or not solely rely on it), but have a safe queue to process these delete commands, surviving reboots, crashes, etc.

app/src/pages/SingleContent/SingleContent.vue
```diff
@@ -81,30 +81,75 @@ type Props = {
...
+// onServerPrefetch fetches it so it is present in the static HTML. Native gets no
+// such benefit, so `cacheId` (and thus a dedicated cache entry per slug) is scoped
+// to isSSG only.
+const { output: contentArr, isFetching: isContentFetching } = useContentQueryWithState(
+    () => [{ slug: props.slug }],
+    {
+        includeScheduled: false,
+        languageFilter: false,
+        cache: true,
+        cacheId: isSSG ? props.slug : undefined,
```
ivanslabbert, line 100:
```suggestion
        cacheId: isSSG ? props.slug : undefined,
        persistOffline: isSSG ? false : true, // Disable offline persistence on a SSG run
```

app/src/pages/SingleContent/SingleContent.vue
```diff
@@ -168,173 +193,237 @@ function routeRedirect(redirect: RedirectDto): boolean {
...
 );
-const allLanguages = useHybridQuery<LanguageDto>(() => ({ selector: { type: DocType.Language } }), {
-    live: true,
-    // Only the i18n singleton in globalConfig needs `translations`; this query reads
-    // just id/name/languageCode/averageReadingSpeed, so drop the heavy strings map.
-    stripFields: ["translations", "_rev"],
+
+const availableTranslations = computed<ContentDto[]>(() => {
+    if (!content.value) return [];
+    const published = translationsArr.value.filter((c) => c.status === PublishStatus.Published);
```
ivanslabbert: This query will return docs that are published but the publish date is in the future. Perhaps we should use mangoIsPublished and compile it to a filter function to keep one "source of truth"

app/src/pages/SingleContent/SingleContent.vue
```diff
@@ -168,173 +193,237 @@ function routeRedirect(redirect: RedirectDto): boolean {
...
+                memberOf: [],
+                languageCode: t.language.replace(/^lang-/, ""),
+                name: t.language,
+                translations: {},
+            } as LanguageDto),
+    );
+});
+
+// Reciprocal hreflang alternates (language code + slug) for the SEO head.
+const hreflangAlternates = computed(() =>
```
ivanslabbert: Gate behind isSSG

app/src/pages/SingleContent/SingleContent.vue
```diff
@@ -168,173 +193,237 @@ function routeRedirect(redirect: RedirectDto): boolean {
...
-            availableTranslations.value = published;
-            languages.value = allLanguages.value.filter((lang) =>
-                published.some((t) => t.language === lang._id),
-            );
-        }
-    },
-    { immediate: true },
+const localLanguages = ref<LanguageDto[]>([]);
+
+if (!import.meta.env.SSR) {
```
ivanslabbert:
```suggestion
if (!isSSG) {
```

app/src/pages/SingleContent/SingleContent.vue
```diff
@@ -168,173 +193,237 @@ function routeRedirect(redirect: RedirectDto): boolean {
...
+        includeScheduled: false,
+        cache: true,
+        // Per-slug discriminator so the per-document cache is safe.
+        cacheId: `tags:${props.slug}`,
+    },
 );
 
 const categoryTags = computed(() => tags.value.filter((t) => t.parentTagType == TagType.Category));
-const selectedCategoryId = ref<Uuid | undefined>();
+const publicTaxonomy = computed<PublicTaxonomy[]>(() => {
```
ivanslabbert: gate behind isSSG

app/src/pages/SingleContent/SingleContent.vue
```diff
@@ -168,173 +193,237 @@ function routeRedirect(redirect: RedirectDto): boolean {
...
+        }
+        seen.add(tag.parentId);
+        return [{ name: tag.title, url: `/${tag.slug}` }];
+    });
+});
 
-// isLoading / is404 are declared alongside the content sources above.
+// SEO head — driven by resolved public content and taxonomy so the prerendered
+// page, canonical metadata, Open Graph image and JSON-LD always agree.
+useContentHead(content, hreflangAlternates, publicTaxonomy);
```
ivanslabbert: gate behind isSSG

app/src/pages/SingleContent/SingleContent.vue
```diff
@@ -368,30 +457,45 @@ const isBookmarked = computed(() => {
...
-        metaTag.setAttribute("name", "description");
-        document.head.appendChild(metaTag);
-    }
-    // Update the content attribute
-    metaTag.setAttribute("content", content.value?.seoString || content.value?.summary || "");
-});
+// Native sets the title/description imperatively (it has no @unhead plugin). The web
+// build's `useHead` above owns the head there (and serializes it into the prerendered
+// HTML), so skip this path on web.
+if (import.meta.env.VITE_BUILD_TARGET !== "web")
```
ivanslabbert:
```suggestion
if (!isSSG)
```

app/src/pages/SingleContent/SingleContent.vue
```diff
@@ -368,30 +457,45 @@ const isBookmarked = computed(() => {
...
+        // SEO meta tag settings
+        let metaTag = document.querySelector("meta[name='description']");
+        if (!metaTag) {
+            // If the meta tag doesn't exist, create it
+            metaTag = document.createElement("meta");
+            metaTag.setAttribute("name", "description");
+            document.head.appendChild(metaTag);
+        }
+        // Update the content attribute
+        metaTag.setAttribute("content", content.value?.seoString || content.value?.summary || "");
```
ivanslabbert: This is not needed to run in a hydrated app. It should only be rendered into the SSG built page

app/src/pages/SingleContent/SingleContent.vue
```diff
@@ -81,30 +81,75 @@ type Props = {
...
-    parentId: "",
-    language: appLanguagePreferredIdAsRef.value || "",
-    status: PublishStatus.Published,
-    title: "Loading...",
-    slug: "",
-    publishDate: 0,
-    parentTags: [],
-};
+// Content by slug — the SAME local-first hybrid query on every build (web and
+// native). On the web build (isSSG) the prerender primes this query's response cache
```
ivanslabbert:
```suggestion
// native/hydrated). On the web build (isSSG) the prerender primes this query's response cache
```

app/src/pages/SingleContent/SingleContent.vue
```diff
@@ -81,30 +81,75 @@ type Props = {
...
+        // for CouchDB to engage the index (slug eq alone falls back to a full scan).
+        useIndex: "content-slug-publishDate-index",
+        sort: [{ publishDate: "desc" }],
+        // No stripFields here (unlike overview feeds): this is the document actually
+        // opened in SingleContent, and it should persist offline with `fts`/
+        // `ftsTokenCount` intact so it surfaces in offline FTS search.
+        // `text` is the heaviest field on the page and already sits in the prerendered
+        // `[data-ssr-article-text]` node — omit it (and the FTS/rev fields, which the
+        // live/persisted copy above now keeps) from the SSR-authored cache write so it
+        // isn't shipped twice; the hydration patch below recovers `text` from that node.
```
ivanslabbert: Remove comment referring to deleted code

app/src/pages/SingleContent/SingleContent.vue
```diff
@@ -81,30 +81,75 @@ type Props = {
...
+    },
+);
+
+// One-time hydration patch (client only): the response-cache seed above may have
+// omitted `text` (`ssrCacheStripFields`) — recover it from the DOM before first
+// render so the `v-html` binding below matches the prerendered HTML exactly. Runs
+// synchronously in setup (not a watcher) so it lands before the template's first
+// evaluation. A no-op when the seed already carries `text` (warm client cache) or
+// there's nothing to recover.
+if (!import.meta.env.SSR) {
```
ivanslabbert:
```suggestion
if (!isSSG) {
```

app/src/pages/SingleContent/SingleContent.vue
```diff
@@ -168,173 +193,237 @@ function routeRedirect(redirect: RedirectDto): boolean {
...
 }
 
-// Loading until the content query produces an answer for the current slug.
-const isLoading = ref(true);
-const is404 = ref(false);
+// Loading until the content query produces an answer for the current slug — cleared
+// by the bind watch on the client, or by the not-found resolver below. During the
+// prerender the content is fetched in onServerPrefetch and present at render, so the
+// loading branch is never serialized.
+const isLoading = ref(!import.meta.env.SSR);
```
ivanslabbert:
```suggestion
const isLoading = ref(!isSSG);
```

ivanslabbert (commented):
import.meta.env.SSR should perhaps be renamed to import.meta.env.SSG, as we are only referring to SSG in the Luminary project (the project does not support an SSR concept)

app/src/pages/SingleContent/SingleContent.vue
```diff
@@ -168,173 +193,237 @@ function routeRedirect(redirect: RedirectDto): boolean {
...
-            }
-            if (!contentArr.value.length) {
-                content.value = undefined;
-                isLoading.value = false;
-            }
-        },
-        isConnected.value ? 5000 : 1500,
-    );
+// never emits. Rather than guessing off a wall-clock timeout (which races any query
+// slower than the guess — slow device, cold sync, congested network: see the SSR
```
ivanslabbert:
```suggestion
// slower than the guess — slow device, cold sync, congested network: see the SSG
```

app/src/pages/SingleContent/SingleContent.vue
```diff
@@ -168,173 +193,237 @@ function routeRedirect(redirect: RedirectDto): boolean {
...
-        )[0];
-        if (props.slug === slug && redirect) routeRedirect(redirect);
-    },
-    { immediate: true },
-);
+// Client-only (web + native): redirect / not-found / loading wiring + retention. The
+// Node prerender skips this — it has no Dexie, and `content` is already populated via
+// the seam's onServerPrefetch + rendered. Guard on import.meta.env.SSR (with vite-ssg
+// mock:true, `window` exists in Node, so a window check would run this in the build).
+if (!import.meta.env.SSR) {
```
ivanslabbert:
```suggestion
if (!isSSG) {
```

app/src/ssg/routeIndex.ts
```diff
@@ -0,0 +1,76 @@
+/**
+ * Content-id / parent-id → prerendered route sidecar (`dist-web/ssg-route-index/`).
+ *
+ * Why this exists: prerendering writes static HTML files keyed by **route** (slug),
+21 * but ISR deletion is driven by a CouchDB DeleteCmd, which historically carried only
```
ivanslabbert: Typo?
```suggestion
 * but ISR deletion is driven by a CouchDB DeleteCmd, which historically carried only
```

app/src/ssg/deleteQueue.ts
```diff
@@ -0,0 +1,169 @@
...
+ * deploy repo reads it, performs the actual delete + purge, then deletes that one file
+ * itself once done — a crash at any point just means the file is still there on the
+ * next pass, no shared/merged state to lose sync with.
+ *
+ * One file per entry rather than sharded, unlike `ssg-route-index`/`ssg-doc-facets`:
+ * this sidecar is consumed entirely locally by the deploy repo (never uploaded to
+ * R2/Cloudflare, per ADR 0018 — sharding's rationale of bounding what a consumer must
+ * load/rewrite doesn't need a bucketing scheme when each entry already IS its own
+ * file), and "processed" becomes a plain `rm <id>.json` with no read-merge-write race
+ * against other entries sharing a file.
```
ivanslabbert: Remove domain specific info

app/src/ssg/deleteQueue.ts
ivanslabbert, line 1: add slug rename support (generate delete commands for all slugs in the previousSlugs field of a content doc (to be added)

shared/src/util/HybridQuery/HybridQuery.ts
```diff
@@ -106,19 +106,24 @@ export async function queryRemote<T = unknown>(query: MangoQuery): Promise<T[]>
 
     const payload: Record<string, unknown> = {
         selector: query.selector,
-        identifier: "hybridQuery",
+        // Callers outside the app's live queries override this so their load is separable in
+        // the API's expensive-query logs.
+        identifier: query.identifier ?? "hybridQuery",
```
ivanslabbert, line 112: Having a switch like this can open up for an attacker to set the switch, and bypass the expensive query backoff, putting the API at risk for DoS attacks.

A better way would be to have a whitelist in the API, allowing specified IP addresses to bypass the "heavy query penalty"

ivanslabbert (commented):
The mechanism used to detect changes to e.g. related content is too intertwined with HybridQuery at the moment, which adds a risk of breaking SSG/ISR if we make changes to HybridQuery. We need to decouple it, so that SSG/ISR can run by itself with less risk of being broken by app changes.

HANDOFF-ssg-analysis.md
ivanslabbert, line 1: This file should not be included when merging to main. Log to follow-up tickets where applicable

ssg-standup.md
ivanslabbert, line 1: This file should not be included when merging. Move to separate tickets where applicable

tobediscussed.md
ivanslabbert, line 1: This file should not be included when merging. Move to separate tickets where applicable

todos.md
ivanslabbert, line 1: This file should not be included when merging. Move to separate tickets where applicable

_potential_performance_increases
ivanslabbert, line 1: This file should not be included when merging. Move to separate tickets where applicable

.worktree-ports.env
ivanslabbert, line 1: Not sure this should be included in the repo? Delete & gitignore if it should not be included

docs/adr/0018-web-ssg-and-isr-tier.md
ivanslabbert, line 1: Luminary does not include a capacitor build. Remove references to 3-rd party builds from the documentation

app/src/ssg/README.md
ivanslabbert, line 1: Luminary does not include a Capacitor build. Remove references to third-party builds from the docs

ivanslabbert (approved):
Approved with comments

PR #1690, opened 2026-06-10
fixes the flaky test

ChrisTouo:


MrDirkelz (commented):
Ran 2 times on the dev deployment workflow, no tests failed at all. But not easy to determine whether it is just luck or if it is actually working

ivanslabbert (approved):

PR #1691, opened 2026-06-10
Changed isMobileScreen to isLargeScreen and fixed sizes

GitTaijitu:


cms/src/components/BasePage.vue
```diff
@@ -33,7 +33,7 @@ const router = useRouter();
 const isEditContentPage = computed(() => router.currentRoute.value.name === "edit");
 const isEditLanguagePage = computed(() => router.currentRoute.value.name === "language");
 const breakpoints = useBreakpoints(breakpointsTailwind);
-const isMobileScreen = breakpoints.smaller("lg");
+const isLargeScreen = breakpoints.smaller("lg");
```
ivanslabbert, line 36: We already have isLargeScreen in globalConfig, and can import it from there instead of re-calculating it here

cms/src/globalConfig.ts
```diff
@@ -32,8 +32,8 @@ const windowWidth = ref(window.innerWidth);
 window.addEventListener("resize", () => {
     windowWidth.value = window.innerWidth;
 });
-export const isMobileScreen = computed(() => windowWidth.value < 1024);
-export const isSmallScreen = computed(() => windowWidth.value < 1500);
+export const isLargeScreen = computed(() => windowWidth.value < 1024);
+export const isSmallScreen = computed(() => windowWidth.value < 640);
```
ivanslabbert, line 36: Need to rethink this. Perhaps better to use the Tailwind breakpoints instead of a pixel value (see how it is done in Basepage)

Perhaps just export the breakpoints variable, and use e.g. breakpoints.smaller("lg") in the actual code where we need to determine the screen size programatically

PR #1706, opened 2026-06-16
feat(shared,app): route FTS search local vs server-side /fts

ivanslabbert:
Add a routing layer over offline FTS that decides, per search, whether to run locally or against the server /fts endpoint (#1684):

- offline or full sync -> local; online + a publishDate cutoff (or CMS) -> API
- single source per search (no merge): the server corpus is a superset of the synced subset, so its result is already complete
- API failure falls back to local; expose source + isPartial; re-run against the API on reconnect to upgrade a partial result
- server results are display-only (trimmed of fts/ftsTokenCount) and must never be persisted to IndexedDB

Optimize the local engine (~1.45s -> ~700ms on a full-sync corpus):

- language pre-filter before loading docs (index-only ID scan)
- word-match bonus only over the top-K by BM25
- high-df trigram pruning (extracted as a unit-tested helper)
- parallelized index scans

App shows an offline/partial-results hint (new search.partialResults i18n key, eng + fra). Docs: ADR 0011 + shared FTS README and CLAUDE.md. Tests cover the routing matrix, the API adapter, word-match ranking, and the pruning helper.


MrDirkelz (commented):
Felt fast even on 3G. took about 3s for 3G but that is good I feel.

MrDirkelz (approved):

PR #1719, opened 2026-06-19
API: add opt-in per-token auth identity cache

MrDirkelz:
Every REST request through AuthGuard re-runs the full identity pipeline (RS256 verify + 3-4 Mango user lookups + a lastLogin write + accessMap projection). The socket handshake amortizes this per connection; the REST path does not, so a busy authenticated client hammers resolveIdentity and the DB.

Add IdentityCacheService, a transport-agnostic cache in front of AuthIdentityService.resolveOrDefault, used by both the REST guard and the socket handshake. It caches the resolved membership + identity keyed by a sha256 of the token and re-derives the accessMap live from PermissionSystem on each hit, so an ACL (groupUpdate) change needs no invalidation.

Design notes:
- TTL = min(token exp, IDENTITY_CACHE_TTL_MS): a hit never serves a token past expiry; a different/expired token misses and is fully verified.
- Bounded by a maxEntries cap + lazy/sweep eviction, no timers (BoundedTtlCache, reusing the strikeLimiter pattern).
- Invalidate on User memberOf removal (PermissionChange DeleteCmd), AuthProvider/AutoGroupMappings changes, and DB disconnect. Not on groupUpdate (accessMap is re-projected) or plain User updates, so lastLogin writes never evict.
- lastLogin now updates once per TTL per instance instead of per request.
- Ships disabled (IDENTITY_CACHE_ENABLED=false); enable per environment.
- Keeps all auth resolution behind one seam, easing a future SSE transport.

NOTE ON THE FOLLOWING DIAGRAM:
The accessMap is not cached in the current code, this part of the plan is just if we decide we want to deduplicate accessMaps that can be the same like super-admin access
```mermaid
flowchart TB
    A(["/query arrives with token"]) --> C{"IDENTITY<br>lookup token?"}
    C -- if found --> D["got mergedGroups<br>skip verify + DB + lastLogin"]
    C -- else --> E["resolveIdentity:<br>verify, DB lookups, lastLogin<br>store in identity cache under token"]
    D --> F["key2 = hash sorted mergedGroups"]
    E --> F
    F --> G{"Access map caching<br>lookup key2?"}
    G -- if found --> H["got shared accessMap<br>skip projection"]
    G -- else --> I["getAccessMap mergedGroups<br>store under key2<br>link key1 → key2"]
    H --> J["QueryService injects filters<br>run query"]
    I --> J
    J --> K(["return results"])
    E1["event: User.memberOf change"] --> A1@{ label: "Evict that identity's Layer I entry<br>NOTE: ignore lastLogin-only changes" }
    E2["event: AutoGroupMappings /<br>AuthProvider change"] --> A2["Full flush of Identity Cache"]
    E3["event: groupUpdate"] --> A3["Recompute each distinct projection<br>diff vs cached, update in place<br>linked tokens auto-update"]
    E4["event: DB disconnect"] --> A4["Clear BOTH caches"]
    A1 --> L1[("Identity Cache<br>token → membership")]
    A2 --> L1
    A4 --> L1 & L2[("Access map cache<br>membership-set → accessMap")]
    A3 --> L2

    A1@{ shape: rect}
```


MrDirkelz (commented):
### On re-projecting the accessMap instead of caching it

Agree with this call, and I measured it so the trade-off is on the record.

Caching membership but re-deriving the accessMap from `PermissionSystem` on every hit is the right split. It means a `groupUpdate` needs no invalidation path at all — an ACL change takes effect on the very next request, with no window where a cached map grants access that has just been revoked. Caching the projection would have bought a little CPU and cost exactly the invalidation complexity this design avoids.

**What the re-projection costs**, benchmarked against the real 61-group graph from a local dev database (Node 26, warm, 2000 iterations per case):

| membership | `getAccessMap()` | serialized accessMap |
| --- | ---: | ---: |
| 1 group (app user) | 1.6 µs | 0.3 KB |
| 3 groups (typical) | 3.0 µs | 1.1 KB |
| 10 groups (editor) | 7.9 µs | 2.1 KB |
| 61 groups (super-admin) | 162 µs | 50 KB |

Synthetic scaling, flat groups each granting themselves every permission: 100 → 279 µs, 250 → 542 µs, 500 → 1.09 ms, 1000 → 2.15 ms. Linear, roughly 2.1 µs per group.

For context on the other side of the trade: I measured the uncached auth path at **~15.9 ms mean and 7 CouchDB round trips per request** (4 finds, 2 gets, 1 write), so for a typical 3-group user the re-projection is ~0.02% of what the cache saves, and even for a 61-group super-admin it is ~1%. Comfortably the right trade today.

**Two things worth knowing before it stops being free.**

It is synchronous CPU on the event loop rather than I/O, so it becomes a throughput ceiling rather than a latency cost once the cache removes the DB round trips. At 162 µs a super-admin tops out around 6,100 req/s per core on projection alone; at 1,000 groups that is roughly 460 req/s. Nowhere near current traffic, but it is the number that moves as the group graph grows, and group count is the thing most likely to grow quietly.

The same projection is also what the socket handshake serializes to the client on connect — 50 KB for a 61-group super-admin, and ~877 KB at 1,000 groups. That is per connect and per reconnect, so it is worth watching independently of the REST path.

**Suggestion:** leave it as is, and treat the Layer 2 idea from the diagram (memoize by a hash of the sorted membership set) as a documented follow-up with a trigger rather than something to build now. A reasonable trigger would be the point where a common membership set exceeds a few hundred groups, since the memo key is the membership set and admins tend to share one — that is precisely when dedup starts paying and not before.

One small aside found while benchmarking, unrelated to the caching decision: `PermissionSystem.upsertGroups()` drains the array it is handed (`groupDocs.splice(0, 1)` in a `while` loop), so the caller's array is empty afterwards. Every current caller passes a throwaway array so nothing is broken, but it surprised me and is worth a line of JSDoc.

johan-bell (changes requested):
Invalidation design checks out (live re-derivation, change-feed removals, write-side `permissionChange` event). Blockers:

- `configuration.ts:191` — `IDENTITY_CACHE_ENABLED !== "false"` ships the cache ON; PR title/body and `.env.example` all say opt-in/off. Every deployment silently gains a 5-min token-revocation staleness window.
- `identityCache.README.md:18` — "Off by default" heading contradicts line 20.
- `boundedTtlCache.ts:49` — `maxEntries` isn't enforced: sweep only drops expired entries, so live tokens above the cap grow unbounded with O(n) scans that free nothing.
- `identityCache.service.ts:140` — cache hits omit `jwtPayload`, so `request.user.jwtPayload` differs between hit and miss.
- `permissions.service.ts:197` — the memoised AccessMap is shared mutable state guarded only by a doc comment; consider freezing it.

PR #1724, opened 2026-06-22
feat(app): adaptive image quality on slow connections + data-saver toggle

MrDirkelz:
Serve lighter images when bandwidth is constrained, decided by three signals (OR): a measured-slow connection, a new user Data Saver toggle, and the OS/browser Data Saver flag.

- useNetworkSpeed: probe-based speed estimate (times a download of a small same-origin asset) instead of navigator.connection.downlink, which is Chromium-only. Same code path on every browser; refreshes on startup, regained connectivity, tab focus, and toggling Data Saver off; throttled and skipped when offline or already saving data.
- LImageProvider: in reduced-data mode, trim the srcset ladder to <=600px (keeping the smallest) and advertise reduced `sizes`. Recomputes live.
- LToggle: native <button role="switch"> toggle (no HeadlessUI), themed for the app; used for the Settings Data Saver control.
- SettingsPage: connection speed now reads the live composable value; adds the Data Saver card, i18n'd via new settings.data_saver.* keys.
- globalConfig: add persisted userDataSaverEnabled; drop the unused Chromium-only getConnectionSpeed.
- api: seed settings.data_saver.{title,description} in lang-eng / lang-fra.

Adds app/public/network-probe.bin (100 KB incompressible) as the probe target. New unit tests for the composable, LToggle, and the image trim.|

image


app/src/composables/useNetworkSpeedEstimator.ts
ivanslabbert, line 1: rename to useNetworkSpeedEstimator (or something shorter that better describes the composable)

app/src/composables/useNetworkSpeed.ts
```diff
@@ -0,0 +1,98 @@
...
+ * and silently absent on Safari/Firefox, we time the download of a small same-origin asset. The same
+ * code path runs on every browser, so behaviour is consistent everywhere. The estimate is refreshed
+ * "when relevant" (startup, regained connectivity, tab focus) rather than continuously.
+ *
+ * NOTE: a small probe is influenced by latency as much as raw bandwidth, so fast links under-report.
+ * That's fine here — the only decision we make is the `< SLOW_CONNECTION_MBPS` threshold, and the
+ * Settings page surfaces the live number so the constants below can be tuned against real readings.
+ */
+
+const isTestEnv = import.meta.env.MODE === "test";
```
ivanslabbert: We should preferably not have testing switches in the production code. This causes our tests to run in a different environment than in production. See if you can mock rather?

app/src/pages/SettingsPage.vue
ivanslabbert, line 1: When browser data savings mode is detected, force the toggle to the on position (visual only, do NOT set the app level data saving mode to true), and add a note that browser data saving mode is detected.

When slow network speed is detected (and neither browser or app level data saving mode is enabled), add a note that data saving mode will be applied due to slow network conditions.

ivanslabbert (approved):
Add note in settings when detecting that the browser is in low-data mode, stating why the setting is disabled (as the browser setting takes preference). Note to be in end-user understandable language with i18n support

PR #1735, opened 2026-06-24
Resolving: Title on vertical image not bottom-aligned

brucefrankstad26-beep:
APP: Title on vertical image not bottom-aligned #1721


MrDirkelz (approved):
We just need to test 1 more time before merging, as think a rebase has to be done . But nice

PR #1736, opened 2026-06-25
fix(app): fix HorizontalContentTileCollection nav buttons on desktop

MrDirkelz:
- Raise wrapper z-index from z-10 to z-30 so it sits above ContentTile's
  z-20 overlay divs (video/audio icons) — these were intercepting clicks
  and bubbling up to the RouterLink, causing navigation instead of scroll
- Remove duplicate @click from the arrow icons (was calling spinLeft/Right
  twice per click, scrolling 200px instead of 100px)
- Fix right arrow missing base mt-7 and conflicting hardcoded md:mt-10
- Add transition-opacity to smooth the 80%→100% hover opacity change

### DIFF
--- a/app/src/components/content/HorizontalContentTileCollection.vue
+++ b/app/src/components/content/HorizontalContentTileCollection.vue
@@ -113,29 +113,27 @@ useInfiniteScroll(
 
         <div class="relative">
             <div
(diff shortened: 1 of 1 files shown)


ivanslabbert (approved):
Make sure that when the spin button is faded out (not visible anymore) that clicks are passed to the underlying content tile, otherwise it will feel buggy to the user.

PR #1747, opened 2026-07-01
fix(app,cms): reconnect socket when a backgrounded tab becomes visible again

MrDirkelz:
A tab idle/backgrounded long enough (laptop sleep, OS-throttled timers) can leave the socket disconnected with auto-reconnect turned off (socketio.ts disables it after an auth_failed) and no future retry scheduled, since socket.io's own reconnection backoff is JS-timer-driven and gets throttled while hidden. Foregrounding the tab now triggers an immediate reconnect attempt instead of waiting on a possibly-stalled timer, reusing the existing connect_error -> refresh -> re-login recovery path unchanged.

### DIFF
--- a/app/src/main.ts
+++ b/app/src/main.ts
@@ -85,8 +85,8 @@ async function Startup() {
     socket.on(
         "connect_error",
         async (err: Error & { data?: { type?: string; reason?: string } }) => {
-            reconnecting = false;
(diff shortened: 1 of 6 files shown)


ivanslabbert (commented):
As discussed, this can be moved to the shared library

ivanslabbert (approved):

PR #1751, opened 2026-07-02
fix(shared): cancel pending debounce when runSearch is called explicitly

johan-bell:


MrDirkelz (commented):
Just have to test but lgtm

MrDirkelz (commented):
Some conlicts

ivanslabbert (approved):

PR #1756, opened 2026-07-03
1693 cms always available offline toggle to force sync evergreen content

johan-bell:


api/src/db/schemaUpgrade/v20.ts
ivanslabbert, line 1: as this is (or should be) a new optional feature, I do not think we need an upgrade script.

ivanslabbert (commented):
alwaysOfflineOnly could perhaps be renamed to just be alwaysOffline?

Investigate if it is feasible to handle this transparently in the shared library without needing to make changes to the app's sync.ts?

MrDirkelz (commented):
Just need to test, but looks good to me

api/src/changeRequests/documentProcessing/processPostTagDto.ts
```diff
@@ -170,6 +170,7 @@ export default async function processPostTagDto(
 
         contentDoc.parentPublishDateVisible = doc.publishDateVisible;
         contentDoc.parentShowComingSoon = doc.showComingSoon ?? false;
+        contentDoc.parentAlwaysOffline = doc.alwaysOffline ?? false;
```
ivanslabbert: This will set the field to false when the parent doc's alwaysOffline is set, instead if deleting / not adding the optional field.
This increases the "weight" unnecessarily, as most documents will not need the field to be set.

ivanslabbert (approved):
approved with comment (must fix)

PR #1771, opened 2026-07-06
fix(api): derive language cap from named constants instead of a magic number (#1765)

johan-bell:
## What & why

The non-CMS `/query` language cap was the magic number **4**, restated in three places and kept "in step" with the client's 3-preferred-language cap only by a comment:

- `DEFAULT_MAX_LANGUAGES = 4` in `validateQuery.ts`
- `parseInt(process.env.QUERY_MAX_LANGUAGES, 10) || 4` in `configuration.ts`
- `?? 4` in `query.controller.ts`

Constants that must move together (GitHub #1765) had nothing enforcing it — the `3 (preferred) + 1 (default) = 4` relationship lived in prose, and the literal `4` was duplicated across three files. If the client's preferred-language cap ever changed, all three call sites had to be remembered and updated by hand.

An earlier version of this PR also added a `LANGUAGE_CAP_HEADROOM` slot (cap = 5) per #1765's "zero headroom" framing. Per review discussion (Ivan, then confirmed by tracing the code), the cap only counts values under `language` in the query selector — that's unrelated to how many Language docs sync (unbounded regardless), and the client-side preferred-language cap is a fixed UX invariant (`normalizePreferredLanguages` hard-slices to `MAX_PREFERRED_LANGUAGES`), not something that grows as the language catalog grows. There was no case where headroom was actually load-bearing, so it was dropped; the cap stays at exactly 4.

## Change

- `DEFAULT_MAX_LANGUAGES` is now **derived** from named constants that move together:
  `MAX_PREFERRED_LANGUAGES (3)` + `AUTO_APPENDED_DEFAULT_LANGUAGES (1)` = **4** (unchanged value).
  `MAX_PREFERRED_LANGUAGES` documents that it mirrors `app/src/globalConfig.ts`'s constant.
- The two duplicate fallbacks in `configuration.ts` and `query.controller.ts` now reference `DEFAULT_MAX_LANGUAGES` instead of a literal — single source of truth.
- Adds a test asserting the derived cap equals the client's actual maximum, and updates two stale comments.

No behavioural change for legitimate clients (the cap is still 4); this removes the "must move together" fragility from #1765 without adding unjustified slack.

## Verification

`npx jest src/validation/query/validateQuery.spec.ts` → **44 passed**. `npx tsc --noEmit` clean.

Closes #1765

### DIFF
--- a/api/src/validation/query/validateQuery.ts
+++ b/api/src/validation/query/validateQuery.ts
@@ -33,22 +33,15 @@ export const MAX_PREFERRED_LANGUAGES = 3;
 export const AUTO_APPENDED_DEFAULT_LANGUAGES = 1;
 
 /**
- * Extra languages allowed above the exact client maximum. Without it the cap would sit exactly on
- * the largest legitimate query (zero headroom, GitHub #1765): any future language dimension — or a
- * query that legitimately references one more id — would trip a spurious 400. One slot of headroom
- * removes that fragile boundary; CMS queries are exempt and the query-cost delta of one extra
- * language is negligible.
- */
-export const LANGUAGE_CAP_HEADROOM = 1;
-
-/**
  * Fallback language cap when the caller doesn't supply one. Derived so the pieces move together
- * instead of being restated as a magic number: preferred (3) + auto-appended default (1) +
- * headroom (1) = 5. A legitimate non-CMS content query references at most preferred + default
- * distinct languages; the headroom keeps the boundary off that exact maximum.
+ * instead of being restated as a magic number: preferred (3) + auto-appended default (1) = 4. A
+ * legitimate non-CMS content query never references more than this — the client-side selection
+ * cap is a fixed UX invariant (`normalizePreferredLanguages` hard-slices to
+ * `MAX_PREFERRED_LANGUAGES`), independent of how many languages exist in the system. Growing the
+ * language catalog doesn't grow a single query's language count; only bumping the client cap
+ * itself would, and that already requires updating `MAX_PREFERRED_LANGUAGES` here in lockstep.
  */
-export const DEFAULT_MAX_LANGUAGES =
-    MAX_PREFERRED_LANGUAGES + AUTO_APPENDED_DEFAULT_LANGUAGES + LANGUAGE_CAP_HEADROOM;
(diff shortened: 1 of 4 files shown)


api/src/validation/query/validateQuery.ts
```diff
@@ -22,12 +22,26 @@ export type ValidateQueryOptions = {
...
+/**
+ * Fallback language cap when the caller doesn't supply one. Derived so the pieces move together
+ * instead of being restated as a magic number: preferred (3) + auto-appended default (1) = 4. A
+ * legitimate non-CMS content query never references more than this — the client-side selection
+ * cap is a fixed UX invariant (`normalizePreferredLanguages` hard-slices to
+ * `MAX_PREFERRED_LANGUAGES`), independent of how many languages exist in the system. Growing the
+ * language catalog doesn't grow a single query's language count; only bumping the client cap
+ * itself would, and that already requires updating `MAX_PREFERRED_LANGUAGES` here in lockstep.
+ */
+export const DEFAULT_MAX_LANGUAGES = MAX_PREFERRED_LANGUAGES + AUTO_APPENDED_DEFAULT_LANGUAGES;
```
MrDirkelz, line 44: This feels unnessasary. Since default max languages was already set before in the original. I feel like this does not really solve a real issue since this same variable is just being reused through this PR

johan-bell: Ivan raised the same question above — headroom was in the first commit, but I traced the code path and couldn't find anywhere it'd actually be load-bearing (the cap only counts values under `language` in the query selector; that's unrelated to how many Language docs sync, which is unbounded regardless). Dropped it in the follow-up commit, so `DEFAULT_MAX_LANGUAGES` is back to exactly 4.

What's left is the other half of #1765: the `4` was a magic number duplicated in three files (`validateQuery.ts`, `configuration.ts`, `query.controller.ts`), kept "in sync" only by a comment. It's now derived once from `MAX_PREFERRED_LANGUAGES + AUTO_APPENDED_DEFAULT_LANGUAGES`, and the other two spots reference that instead of restating the literal — no behavior change today, but one constant to update instead of three call sites to remember if the client cap ever moves. Title/description were still describing the first commit (headroom); updating those now.

PR #1774, opened 2026-07-06
1760 cms save content before creating slug-change redirect

MrDirkelz:
Renaming a published post's slug produced a redirect the API rejected ("Published content already exists for slug"): the redirect change request reached the API before the content slug-change, so the old slug was still occupied by published content.

Build the redirects from pre-save state, save the content first (vacating the old slug), then queue the redirects claiming it. The sequential FIFO localChanges drain guarantees the content commits before the redirect is sent.


cms/src/composables/awaitLocalChangeResolution.ts
```diff
@@ -0,0 +1,28 @@
...
+        }
+        scope.run(() => {
+            const count = useDexieLiveQuery(() => db.localChanges.where({ docId }).count(), {
+                initialValue: undefined as number | undefined,
+            });
+            watch(count, (n) => n === 0 && finish("resolved"), { immediate: true });
+            watch(isConnected, (online) => !online && finish("pending"));
+        });
+    });
+}
```
MrDirkelz: Tthis is done in the wrong place

.claude/settings.json
```diff
@@ -1,12 +1,8 @@
 {
   "permissions": {
-    "allow": [
-      "Bash(npx vitest:*)",
-      "Bash(npm view:*)",
-      "Bash(npm run:*)"
-    ]
+    "allow": ["Bash(npx vitest:*)", "Bash(npm view:*)", "Bash(npm run:*)"]
   },
   "enabledPlugins": {
```
ivanslabbert: remove

ivanslabbert (approved):

PR #1775, opened 2026-07-06
Add vite-pwa support to the cms and fix the titlebar bug in the app

MrDirkelz:


ivanslabbert (approved):
Detect (if possible) if the browser supports installing PWA's. Show a note or instructions on installing (e.g. that the app can only be installed using Safari on iOS) if the browser does not support installing it by default.

PR #1778, opened 2026-07-06
docs: ADR 0015 — remain on CouchDB; fix query/index performance in place

ivanslabbert:
Records the outcome of the DocumentDB migration evaluation in #1700.

## Decision
**Remain on CouchDB.** Do not re-platform to the open-source DocumentDB / FerretDB stack. Address the server-side indexing/query-performance concerns within CouchDB (index-design audit + extending ADR 0011 response caching).

## Why
- The migration is not a driver swap but a **scaling-architecture redesign**: our model rests on two CouchDB primitives (the `_changes` feed and multi-master replication) that the open-source DocumentDB stack does not provide.
- The ADR also records the **previously-implicit scaling strategy** — low-cost, cloud-agnostic, homogeneous cheap peers via multi-master replication — as the binding constraint DocumentDB can't meet. Future proposals introducing a central write primary or a managed data service in the critical path are now measured against it.
- FTS is already solved **offline** via our own trigram engine (ADR 0009), so CouchDB Nouveau is explicitly **not** adopted.

## Scope
Single new file: `docs/adr/0015-remain-on-couchdb.md`. No code changes.

Closes #1700

### DIFF
--- a/docs/adr/0015-remain-on-couchdb.md
+++ b/docs/adr/0015-remain-on-couchdb.md
@@ -1,0 +1,118 @@
+# 15. Remain on CouchDB; fix query/index performance in place
+
+Date: 2026-07-06
+
+## Status
+
+Accepted
+
+## Context
+
+CouchDB's server-side indexing and `/query` performance for a handful of hard-to-index
+query classes prompted an evaluation (issue #1700) of migrating the backend off CouchDB to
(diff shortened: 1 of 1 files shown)


johan-bell (approved):

ivanslabbert (commented):
Possible conflicts with (merged) #1779. Check and fix before merging

PR #1781, opened 2026-07-06
Redesign LanguageModal rows as segmented bars with a right-side offline toggle

MrDirkelz:
Moves the offline-sync checkbox to the right side of each language row as a labelled segment (matches ticket 1759), keeps reorder/remove controls left-aligned and always rendered (disabled at the ends) so columns stay aligned, and adds section dividers plus dashed "add language" pills for the not-yet-selected list.

https://github.com/user-attachments/assets/10ecef51-698d-4bcd-96a5-c3a5a294143f

### DIFF
--- a/app/src/components/icons/DragHandleIcon.vue
+++ b/app/src/components/icons/DragHandleIcon.vue
@@ -1,0 +1,42 @@
+<!-- A 2×3 dot grid drag handle ("grip"). Heroicons ships no grip/drag-handle icon, so this is
+     hand-rolled to match the shape used by Material Symbols (`drag_indicator`) and Lucide
+     (`grip-vertical`). Sized by the caller via `class` (e.g. `class="h-4 w-4"`); colour follows
+     `currentColor`. -->
+<template>
+    <svg
+        viewBox="0 0 16 16"
+        fill="currentColor"
+        aria-hidden="true"
+    >
+        <circle
+            cx="5.5"
+            cy="3.5"
+            r="1.4"
+        />
+        <circle
+            cx="10.5"
+            cy="3.5"
+            r="1.4"
+        />
+        <circle
+            cx="5.5"
+            cy="8"
+            r="1.4"
+        />
+        <circle
+            cx="10.5"
+            cy="8"
+            r="1.4"
+        />
+        <circle
+            cx="5.5"
+            cy="12.5"
+            r="1.4"
+        />
+        <circle
+            cx="10.5"
+            cy="12.5"
+            r="1.4"
+        />
+    </svg>
+</template>
--- a/app/src/components/navigation/LanguageModal.vue
+++ b/app/src/components/navigation/LanguageModal.vue
@@ -15,18 +15,12 @@ import {
     normalizeSyncedLanguages,
 } from "@/globalConfig";
 import LModal from "../form/LModal.vue";
-import {
-    ArrowDownIcon,
-    ArrowDownTrayIcon,
-    ArrowUpIcon,
-    CheckIcon,
-    XMarkIcon,
-} from "@heroicons/vue/24/solid";
+import { ArrowDownTrayIcon, CheckIcon, XMarkIcon } from "@heroicons/vue/24/solid";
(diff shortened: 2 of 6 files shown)


johan-bell (approved):

MrDirkelz (commented):
Image

Think we can make it draggable rather, then the UI is cleaner perhaps

app/src/components/navigation/LanguageModal.vue
```diff
@@ -206,95 +276,159 @@ const cancel = () => emit("close");
...
+                                cy="12.5"
+                                r="1.4"
+                            />
+                            <circle
+                                cx="10.5"
+                                cy="12.5"
+                                r="1.4"
+                            />
+                        </svg>
+                    </button>
```
MrDirkelz: This should be a seperate component perhaps @johan-bell ?

app/src/components/icons/DragHandleIcon.vue
```diff
@@ -0,0 +1,42 @@
+<!-- A 2×3 dot grid drag handle ("grip"). Heroicons ships no grip/drag-handle icon, so this is
+     hand-rolled to match the shape used by Material Symbols (`drag_indicator`) and Lucide
+     (`grip-vertical`). Sized by the caller via `class` (e.g. `class="h-4 w-4"`); colour follows
+     `currentColor`. -->
+<template>
+    <svg
+        viewBox="0 0 16 16"
+        fill="currentColor"
+        aria-hidden="true"
+    >
```
ivanslabbert, line 10: Is there not an heroicons icon we can use here?

app/src/components/navigation/LanguageModal.vue
```diff
@@ -123,6 +124,76 @@ const decreasePriority = (id: string) => {
...
+// ── Drag-to-reorder ──────────────────────────────────────────────────────────
+// Pointer events, not HTML5 drag-and-drop: the modal is touch-first and DnD never fires on touch.
+// The grabbed row tracks the finger with a translate, and every full row of travel splices
+// `draftOrder` so the neighbours shuffle live underneath it.
+const draggingId = ref<string | null>(null);
+const dragTranslate = ref(0);
+let dragStartY = 0;
+let dragStartIndex = 0;
+let rowStep = 0; // px between adjacent row tops (rows are uniform height)
+
```
ivanslabbert: Extract dragable logic to a separate component if possible. It clutters the file a bit. Are there not existing components (e.g. vue-use) that we could re-use?

ivanslabbert (approved):
Looks very good!
Approved with some comments.

app/src/components/icons/DragHandleIcon.vue
MrDirkelz: Checked — heroicons has no grip/drag-handle icon; closest are `Bars3Icon`/`EllipsisVerticalIcon`, both read as a menu/list icon rather than "drag me", so using one would be confusing here. Kept the hand-rolled SVG (2×3 dot grid, matches Material Symbols `drag_indicator` / Lucide `grip-vertical`) — noted why in the component's own comment.

app/src/components/navigation/LanguageModal.vue
MrDirkelz: Pulled it out into a composable — `app/src/composables/useDragReorder.ts` — LanguageModal.vue now just destructures `{ draggingId, dragTranslate, startDrag, onDragMove, endDrag }` from it and wires them up in the template. Went with a composable rather than a component since there's no markup to render, just refs + pointer handlers.

Also simplified it while extracting: since `startDrag` already calls `setPointerCapture`, the browser retargets all later pointermove/pointerup/pointercancel for that pointer back to the handle regardless of where the finger ends up — so `onDragMove`/`endDrag` can just be plain `v-on` directives on the handle itself, no more manual `window.addEventListener`/`removeEventListener` pairing.

Checked vue-use: `useDraggable` only tracks one element's free x/y, no concept of a list or index — wouldn't remove the reorder logic itself. `useSortable` (vue-use/integrations + SortableJS) is the closer fit but adds a new dependency for something this now does in ~45 lines.

PR #1783, opened 2026-07-06
fix(api): reconcile seeded language translations on startup

MrDirkelz:
Add an idempotent startup step that syncs seeded English/French translation keys into language docs, prunes stale keys across all languages, and preserves existing production values.


api/src/db/languageTranslationSeedReconciliation.ts
```diff
@@ -0,0 +1,97 @@
...
+    for (const translations of Object.values(seedTranslations)) {
+        Object.keys(translations).forEach((key) => allowedKeys.add(key));
+    }
+
+    const seenSeedLanguages = new Set<string>();
+    let updatedCount = 0;
+    let unchangedCount = 0;
+
+    await db.processAllDocs([DocType.Language], async (doc: LanguageDto) => {
+        if (!doc || doc.type !== DocType.Language) return;
```
ivanslabbert: This line will never resolve. processAllDocs already filters on DocType.Language, so it the line can be removed

api/src/db/languageTranslationSeedReconciliation.ts
```diff
@@ -0,0 +1,97 @@
...
+
+        doc.translations = nextTranslations;
+        doc.updatedTimeUtc = Date.now();
+        await db.insertDoc(doc);
+        updatedCount++;
+    });
+
+    for (const languageId of SEEDED_LANGUAGE_IDS) {
+        if (!seenSeedLanguages.has(languageId)) {
+            console.warn(
```
ivanslabbert: Use the API logging system instead of logging to the console. The warning message is a bit misleading? We are not creating new language documents, but only updating translations fields in the documents?

api/src/db/languageTranslationSeedReconciliation.ts
ivanslabbert, line 1: Can we get a shorter name for this?

api/src/main.ts
```diff
@@ -54,6 +55,7 @@ export async function bootstrap() {
         // fresh-DB baseline and v18 backfills that index. Safe in seed mode: on a fresh DB only
         // v18 runs, which is DB-only (no S3, no PermissionSystem needed).
         await upgradeDbSchema(dbService);
+        await reconcileLanguageTranslationSeeds(dbService);
```
ivanslabbert: We do not need to call this twice? Probably the last call is the correct one?

ivanslabbert (approved):
Approved with comments

PR #1787, opened 2026-07-07
chore(env): remove deprecated env configuration

MrDirkelz:
Remove legacy JWT_MAPPINGS from API env examples because auth mappings now live in AuthProvider and AutoGroupMappings docs.

Drop unused PERMISSION_MAP configuration and the dead ChangeRequestService config fields that only existed to read it.

Remove no-op VITE_FAV_ICON runtime favicon updates from app and CMS because index.html already uses VITE_LOGO_FAVICON.

Remove CMS VITE_HIDE_PRIVACY_POLICY from the example env because CMS never reads it.

Delete stale app plugin env guidance for VITE_PLUGIN_PATH and VITE_PLUGINS because plugin targets are now wired in buildTargetVirtuals.

Remove dangling navigationItems return left by the existing branch cleanup.

Verified:
- api: npm run typecheck
- cms: npm run type-check

Note: app typecheck still fails due stale local luminary-shared installed types missing pruneUnsyncedLanguageContent; shared/src and shared/dist already export it.

### DIFF
--- a/api/CLAUDE.md
+++ b/api/CLAUDE.md
@@ -75,7 +75,7 @@ Multi-provider OIDC. `AuthProvider` documents in the DB define each provider (do
 
 **Failure-reason codes** (`AuthFailureReason`) are critical for client behavior: `provider_not_found` tells the client to evict its cached provider doc and re-pick; `token_invalid` triggers a similar eviction so Dexie can resync. The Socket.io gateway in `socketio.ts` injects these into `connect_error` payloads; the REST `AuthGuard` does the equivalent via `UnauthorizedException`.
 
-The `JwtModule` is registered globally in `app.module.ts` (no signing secret — tokens are verified against each provider's JWKS, not a shared secret). Group-membership and identity claims are derived from the claim/group mappings stored on each `AuthProvider` doc and evaluated by `AuthIdentityService` (plus `AutoGroupMappings` docs) — the legacy `JWT_SECRET`/`JWT_MAPPINGS` env vars are gone.
+The `JwtModule` is registered globally in `app.module.ts` with no signing secret — tokens are verified against each provider's JWKS. Group-membership and identity claims are derived from the claim/group mappings stored on each `AuthProvider` doc and evaluated by `AuthIdentityService` (plus `AutoGroupMappings` docs).
 
 ### Socket.io (`src/socketio.ts`)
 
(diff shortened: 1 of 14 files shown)


johan-bell (approved):

api/CLAUDE.md
```diff
@@ -75,7 +75,7 @@ Multi-provider OIDC. `AuthProvider` documents in the DB define each provider (do
 
 **Failure-reason codes** (`AuthFailureReason`) are critical for client behavior: `provider_not_found` tells the client to evict its cached provider doc and re-pick; `token_invalid` triggers a similar eviction so Dexie can resync. The Socket.io gateway in `socketio.ts` injects these into `connect_error` payloads; the REST `AuthGuard` does the equivalent via `UnauthorizedException`.
 
-The `JwtModule` is registered globally in `app.module.ts`. Group-membership and identity claims are derived from arrow-function strings in `JWT_MAPPINGS` env var — these are `eval`-equivalent so the env value is trusted config.
+The `JwtModule` is registered globally in `app.module.ts` (no signing secret — tokens are verified against each provider's JWKS, not a shared secret). Group-membership and identity claims are derived from the claim/group mappings stored on each `AuthProvider` doc and evaluated by `AuthIdentityService` (plus `AutoGroupMappings` docs) — the legacy `JWT_SECRET`/`JWT_MAPPINGS` env vars are gone.
```
ivanslabbert: Comment should not refer to old removed logic

app/src/main.ts
```diff
@@ -34,13 +34,6 @@ export const app = createApp(App);
 // resolve a store (e.g. useNotificationStore) have an active Pinia instance.
 app.use(createPinia());
 
-if (import.meta.env.VITE_FAV_ICON) {
-    const favicon = document.getElementById("favicon") as HTMLLinkElement;
-    if (favicon) {
-        favicon.href = import.meta.env.VITE_LOGO_FAVICON;
-    }
-}
-
```
ivanslabbert, line 43: How is the favicon configured now? Is it part of the VITE build options?

ivanslabbert (approved):
Approved with a couple of comments

app/src/main.ts
MrDirkelz: Yes — it's build-time only now, no JS needed. `app/index.html` already has `<link id="favIcon" rel="icon" href="%VITE_LOGO_FAVICON%">`; Vite natively substitutes any `%VITE_*%` token in `index.html` from the loaded env at build time. The `VitePWA` manifest in `vite.config.ts` separately reuses `VITE_LOGO_FAVICON` as the PWA install icon. The removed code was actually dead already — it did `getElementById("favicon")` while the element's id is `favIcon` (case mismatch), so it never ran.

PR #1788, opened 2026-07-07
feat(cms,api): add toggles to link publish/expiry dates across translations

### LINKED ISSUE #278: CMS: Add toggle to link dates on content documents belonging to the same parent

MrDirkelz:
Editors currently set publishDate/expiryDate independently per language, making it easy for a multi-language rollout to go live at different times by accident. Adds two independent parent-level toggles (linkPublishDates, linkExpiryDates) that, when enabled, propagate that date from whichever translation is saved onto every sibling Content doc sharing the same parentId, enforced API-side via the existing sibling fan-out cascade so it holds even for translations not open in the same CMS session.


api/src/changeRequests/documentProcessing/processContentDto.ts
ivanslabbert, line 1: Move permission checking to validateChangeRequestAccess, where the other permission checks are implemented. We should not allow the API to do half of the work and then fail on a permission check.

cms/src/components/content/EditContent.vue
ivanslabbert, line 1: Add a check if the user has access to all languages by comparing visible translations to the availabletranslations field on a content doc.

api/src/changeRequests/documentProcessing/processContentDto.ts
```diff
@@ -68,6 +63,7 @@ export default async function processContentDto(
         else delete doc.parentAlwaysOffline;
 
         doc.parentUseVerticalTileLayout = parentDoc.useVerticalTileLayout;
+        doc.parentLinkDates = parentDoc.linkDates;
```
ivanslabbert: Only set when parentDoc.linkDates is true
```suggestion
        parentDoc.linkDates && doc.parentLinkDates = true;
```

api/src/changeRequests/validateChangeRequestAccess.ts
```diff
@@ -9,8 +9,33 @@ import { GroupAclEntryDto } from "../dto/GroupAclEntryDto";
...
+        return PermissionSystem.verifyAccess(
+            l.memberOf,
+            DocType.Language,
+            AclPermission.Translate,
+            groupMembership,
+            "any",
+        );
+    });
+}
+
```
ivanslabbert: Extract to helper function

ivanslabbert (approved):
Approved with comments.

PR #1789, opened 2026-07-07
fix(app): deduplicate related-content items across topic rows

johan-bell:
## Summary
- The app's "Related" section on a content page shows one horizontal row per topic tag (e.g. "People in the Bible", "Heroes of faith"). Any item tagged with multiple topics was repeated verbatim in every matching row — same thumbnails, same order, no added value.
- `contentByTag` gains an opt-in `dedupeAcrossTags` option: it claims each content item for a single row, processing topics in ascending order of match count so a broad topic doesn't swallow items a niche topic needs.
- `RelatedContent.vue` is the only consumer that opts in. The five other callers (`HomePagePinned`, `PinnedVideo`, `UnpinnedVideo`, `PinnedTopics`, `UnpinnedTopics`) call the helper with no options, so their behavior (intentional duplication across topic browsing) is unchanged.

## Why
Fixes #1737 — related-content rows currently duplicate the same content across categories instead of surfacing distinct related items, which undermines the "related content" section's purpose of encouraging further reading.

## Test plan
- [x] `npx vitest run src/components/contentByTag.spec.ts` — 4/4 passing, including two new cases covering dedup-on and default (dedup-off) behavior
- [x] `npx vitest run src/components/content/RelatedContent.spec.ts` — 5/5 passing (unchanged)
- [x] `npx eslint` on the three changed files — clean
- [ ] Manual check in the browser that a post with overlapping topics no longer repeats items across rows (not yet done in this session)

### DIFF
--- a/app/src/components/content/RelatedContent.vue
+++ b/app/src/components/content/RelatedContent.vue
@@ -37,7 +37,7 @@ const filtered = computed(() =>
     contentDocs.value.filter((item) => item._id !== props.selectedContent._id),
 );
 
-const contentByTopic = contentByTag(filtered, toRef(props.tags));
+const contentByTopic = contentByTag(filtered, toRef(props.tags), { dedupeAcrossTags: true });
 </script>
 
 <template>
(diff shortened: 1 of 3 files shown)


MrDirkelz (commented):
This deduplication feels much better when scrolling through the app

ivanslabbert (commented):
I think we need a complete redesign of the way we show related content. E.g. vertical scrolling by post, marked with applicable topics (or should we even show the topics?)

Please come with a conceptual design for discussion

PR #1790, opened 2026-07-07
feat(cms): notify editors of concurrent server changes (#932)

### LINKED ISSUE #932: CMS: EditContent - Implement function to notify about changes made by other users
Implement Network Changes checking functionality made in #743 (EditContent.vue) here.

Acceptance Criteria:
- [ ] User must be able to have a normal workflow while someone else is editing the same content document.
- [ ] User must be able to see if someone else pushed changes to the same document.
- [ ] Find a good way to either save the current progress of the user, or force the user to refresh before continuing.
- [ ] Bugs must be fixed, and original functionality may not be negatively affected.

```ts
// Check for any changes from the server when no local changes has been made
const networkChanges = computed(() => {
    if (!initialContent.value || !initialParent.value) return false;
    // Check if the current existing (from DB/server) parent is different from what was initially loaded
    const parentChangedFromServer = !_.isEqual(
        { ...existingParent.value, updatedBy: "", _rev: "" },
        { ...initialParent.value, updatedBy: "", _rev: "" },
    );

    // Check if the current existing (from DB/server) content is different from what was initially loaded
    const contentChangedFromServer = !_.isEqual(
        existingContent.value.map((c) => {
            // Create a copy of each content and delete values that can reduce data quality for accuracy
            const content = { ...c } as any;
            delete content._rev;
            delete content.updatedTimeUtc;
            return content;
        }),
        initialContent.value.map((c) => {
            // Create a copy of each content and delete values that can reduce data quality for accuracy
            const content = { ...c } as any;
            delete content._rev;
            delete content.updatedTimeUtc;
            return content;
        }),
    );

    return (parentChangedFromServer || contentChangedFromServer) && !isLocalChange.value;
});
```
...
```vue
                <LBadge v-if="isLocalChange" variant="warning">Offline changes</LBadge>
                <LBadge
                    v-if="networkChanges"
                    title="Please refresh the page before continuing"
                    variant="warning"
                    >Changes recieved from the server</LBadge
                >
```
...

MrDirkelz:
Surface a warning banner in EditContent when someone else changes the same document on the server while the user holds unsaved edits, backed by toEditable's existing isModified signal (no shared/ change). Clicking "Review changes" opens a read-only Remote-vs-Yours diff modal; "Use remote version" reloads to the fresh doc, "Keep my changes" dismisses.

### DIFF
--- a/cms/src/components/content/IncomingChangesModal.vue
+++ b/cms/src/components/content/IncomingChangesModal.vue
@@ -1,0 +1,138 @@
+<script setup lang="ts">
+import LDialog from "@/components/common/LDialog.vue";
+import { cmsLanguages } from "@/globalConfig";
+import type { ContentDto, ContentParentDto, Uuid } from "luminary-shared";
+import { computed } from "vue";
+import * as _ from "lodash";
+
+type Props = {
+    existingParent?: ContentParentDto;
+    editableParent?: ContentParentDto;
+    existingContent?: ContentDto[];
+    editableContent: ContentDto[];
+    /** Filters to only the docs that changed on the server (parent + content), keyed by `_id`. */
+    isIncomingChange: (id: Uuid) => boolean;
+};
+const props = defineProps<Props>();
+
+const open = defineModel<boolean>("open");
+
+const emit = defineEmits<{
+    /** User chose the server version — the caller reloads to discard local edits. */
+    accept: [];
+    /** User chose to keep their edits — dismiss and continue. */
+    dismiss: [];
+}>();
+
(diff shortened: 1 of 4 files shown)


johan-bell (approved):

cms/src/components/content/IncomingChangesModal.vue
```diff
@@ -0,0 +1,138 @@
+<script setup lang="ts">
```
johan-bell, line 1: Can't we just reuse LModal or LDialog here instead of creating again another branch ?

ivanslabbert (approved):

PR #1798, opened 2026-07-08
fix(auth): stabilize provider resolution for CMS access maps

MrDirkelz:
Prefer the persisted active provider when a matching Auth0 session cache exists, clear stale provider cache keys during login, and guard CMS startup from anonymous socket handshakes that can purge local content.

Add app parity tests and API warning coverage for provider-scoped group loss.

### DIFF
--- a/api/src/auth/authIdentity.service.spec.ts
+++ b/api/src/auth/authIdentity.service.spec.ts
@@ -1024,6 +1024,65 @@ describe("AuthGuard (Integrated)", () => {
         expect(groups).not.toContain("group-readonly");
     });
 
+    it("should warn when provider scoping removes all static groups from a matched user", async () => {
+        mockJwtService.verifyAsync = jest
+            .fn()
+            .mockResolvedValue({ sub: "auth0|123", email: "test@bccsa.org", email_verified: true });
+
+        const docOtherProvider = {
+            _id: "user-other",
+            _rev: "1-bbb",
+            email: "test@bccsa.org",
+            name: "Test User",
+            providerId: "other-provider",
+            memberOf: ["group-super-admins"],
+        };
+
+        mockDbService.executeFindQuery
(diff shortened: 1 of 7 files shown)


api/src/auth/authIdentity.service.spec.ts
```diff
@@ -1024,6 +1024,65 @@ describe("AuthGuard (Integrated)", () => {
...
 
+    it("should warn when provider scoping removes all static groups from a matched user", async () => {
+        mockJwtService.verifyAsync = jest
+            .fn()
+            .mockResolvedValue({ sub: "auth0|123", email: "test@bccsa.org", email_verified: true });
+
+        const docOtherProvider = {
+            _id: "user-other",
+            _rev: "1-bbb",
+            email: "test@bccsa.org",
```
ivanslabbert: remove domain specific data from test

ivanslabbert (approved):

api/src/auth/authIdentity.service.spec.ts
MrDirkelz:
```suggestion
            email: "test@user.org",
```

api/src/auth/authIdentity.service.spec.ts
```diff
@@ -1024,6 +1024,65 @@ describe("AuthGuard (Integrated)", () => {
         expect(groups).not.toContain("group-readonly");
     });
 
+    it("should warn when provider scoping removes all static groups from a matched user", async () => {
+        mockJwtService.verifyAsync = jest
+            .fn()
+            .mockResolvedValue({ sub: "auth0|123", email: "test@bccsa.org", email_verified: true });
```
MrDirkelz:
```suggestion
            .mockResolvedValue({ sub: "auth0|123", email: "test@user.org", email_verified: true });
```

ivanslabbert (approved):

PR #1714, opened 2026-06-18
1698 cms implement hybridquery

johan-bell:

PR #1812, opened 2026-07-13
CMS: Fix auth provider labels not appearing (#1714)

brucefrankstad26-beep:
CMS: Auth Provider dropdown when editing a user shows the API URL instead of a descriptive provider name

### DIFF
--- a/cms/src/components/users/EditUser.spec.ts
+++ b/cms/src/components/users/EditUser.spec.ts
@@ -32,7 +32,7 @@ const mockAuthProvider: AuthProviderDto = {
     domain: "example.auth0.com",
     audience: "https://example.auth0.com",
     clientId: "client-abc",
-    label: "login.bcc.button",
+    label: "login.test.button",
     displayName: "Example",
     memberOf: ["group-super-admins"],
     updatedTimeUtc: 1704114000000,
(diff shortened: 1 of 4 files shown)


MrDirkelz (commented):
Nice, just need to fix the test, maybe @ChrisTouo you can help him?

cms/src/components/users/EditUser.spec.ts
```diff
@@ -266,7 +266,7 @@ describe("CreateOrEditUser.vue", () => {
             expect(providerSelect.exists()).toBe(true);
             const options = providerSelect.props("options") as { label: string }[];
             expect(options.some((o) => o.label.includes("Choose a provider"))).toBe(true);
-            expect(options.some((o) => o.label === "Example")).toBe(true);
+            expect(options.some((o) => o.label === "login.bcc.button")).toBe(true);
```
MrDirkelz: We should not use organisation specific references in code. It is mostly to keep the open source project generic

MrDirkelz (commented):
Looks good, just one comment

MrDirkelz (commented):
Looks good. As discussed with Bruce the String type cast was added to resolve an error

The organization specific references just has to be removed still ( see unresolved comment )

MrDirkelz (approved):

Issue #813, opened 2025-04-03
APP: Video player 1 px downward shift on Safari / iOS


ivanslabbert:
![Image](https://github.com/user-attachments/assets/2b0ac792-cb37-4c1e-89a2-3a9963169b15)

MrDirkelz:
Hi @johan-bell - you can look at the css selectors or whatever method I used to check if it is the safari browser for another bug fix and then make the required changes based off of that perhaps?

johan-bell:
Hey @MrDirkelz yes thanks. But where did you implement the fix ?

PR #1813, opened 2026-07-13
fix(app): correct 1px video shift on desktop Safari too (#813)

johan-bell:
Re-opens #813 — the 1px downward video shift is back on **desktop (macOS) Safari**.

## Problem
`.video-player` is absolutely positioned, and Safari renders the video 1px lower than that container — a 1px gap at the top and an overhang at the bottom. This affects Safari on **both macOS and iOS**, but not Chrome/Firefox.

The existing correction was gated behind `@supports (-webkit-touch-callout: none)`, which only matches **iOS** — `-webkit-touch-callout` isn't supported by desktop macOS Safari, so the −1px correction never applied there. That's the "safari web" case in the screenshot on the issue.

## Fix
Drive the correction off the platform, not a CSS feature query. The component sets an `isSafariRendering` flag from **video.js's own browser detection** (`videojs.browser.IS_SAFARI || IS_IOS`) — the same source already used for `nativeAudioTracks` — and toggles a `.safari-shift-fix` class. The CSS applies `transform: translateY(-1px)` to that class.

This covers desktop macOS Safari **and** iOS, and excludes Chrome/Firefox (which don't have the shift, so they must not be nudged).

## Tests
- `VideoPlayer.spec.ts`: 12/12. `type-check` + eslint clean.

## ⚠️ Not verified on Safari
This is a Safari-specific rendering fix and I can't drive macOS/iOS Safari in this environment, so it's **verified by reasoning + the existing unit suite, not visually on Safari**. Worth a 10-second look before merge: open a video in **desktop Safari** and confirm the 1px gap/overhang is gone, and in **Chrome** confirm nothing moved (no new 1px misalignment). `videojs.browser.IS_SAFARI` is the same detection the player already trusts, so the risk is low — but low isn't verified.

### DIFF
--- a/app/src/components/content/VideoPlayer.css
+++ b/app/src/components/content/VideoPlayer.css
@@ -194,11 +194,10 @@
     display: none !important;
 }
 
-/* iOS Safari renders the video-js tech element 1px lower than its absolutely-positioned
-   .video-player container, leaving a 1px gap at the top and an overhang at the bottom
-   (issue #813). Same iOS-Safari feature-query technique used in LHighlightable.vue. */
-@supports (-webkit-touch-callout: none) {
-    .video-player {
-        transform: translateY(-1px);
-    }
+/* Safari renders the video 1px lower than its absolutely-positioned .video-player
+   container, leaving a 1px gap at the top and an overhang at the bottom (issue #813).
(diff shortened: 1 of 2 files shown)


johan-bell:
Closing: #813 is closed and this branch conflicts with main. Will reopen with a fresh fix if the Safari shift comes back.

PR #1818, opened 2026-07-13
Fix hybrid content queries with multi-parent server-side index fan-out

MrDirkelz:
Split `hybridQuery` content requests using `parentId: {$in: [...]}` into individual `parentId` equality queries pinned to
`content-parentId-publishDate-index`, then merge, sort, and limit their results server-side. This avoids Mango’s full content-partition scan caused by combining `$in` with a `publishDate` sort.

Add query-service coverage for the fan-out behavior and document why clients may use the above-cap multi-parent query path now that the API can execute it efficiently.

### DIFF
--- a/api/src/endpoints/query.service.ts
+++ b/api/src/endpoints/query.service.ts
@@ -1,4 +1,5 @@
 import { Injectable, Inject, HttpException, HttpStatus } from "@nestjs/common";
+import { ConfigService } from "@nestjs/config";
 import { DbQueryResult, DbService } from "../db/db.service";
 import { AclPermission, DocType, PublishStatus, Uuid } from "../enums";
 import { PermissionSystem } from "../permissions/permissions.service";
@@ -20,6 +21,7 @@ export class QueryService {
         @Inject(WINSTON_MODULE_PROVIDER)
         private readonly logger: Logger,
         private db: DbService,
+        private readonly configService: ConfigService,
     ) {
         // Get list of languages from the database for content doc filtering by user accessible language.
         // This list is kept updated in memory to reduce database load
@@ -309,19 +311,29 @@ export class QueryService {
         const parentIds = [...new Set(rawIds as string[])];
         if (parentIds.length === 0) return { docs: [], blockStart: 0, blockEnd: 0 };
 
-        // Unbounded fan-out; add a concurrency cap if a parent set ever gets large.
-        const results = await Promise.all(
-            parentIds.map((id) =>
-                this.db.executeFindQuery({
-                    ...query,
-                    selector: {
-                        $and: (query.selector.$and || []).map((condition) =>
-                            condition.parentId === parentIdCriteria ? { parentId: id } : condition,
-                        ),
-                    },
-                    use_index: "content-parentId-publishDate-index",
-                }),
-            ),
+        // Reject an oversized parentId fan-out. The client's own fan-out cap is
+        // client-side only, so this is the authoritative backstop.
+        const maxFanoutParents = this.configService.get<number>("query.maxFanoutParents") ?? 200;
+        if (parentIds.length > maxFanoutParents) {
+            throw new HttpException(
+                `'parentId.$in' exceeds the maximum fan-out size (${maxFanoutParents})`,
+                HttpStatus.BAD_REQUEST,
+            );
+        }
+
+        // Cap concurrent CouchDB requests for the fan-out, so one request can't open
+        // one connection per parent at once.
+        const fanoutConcurrency = this.configService.get<number>("query.fanoutConcurrency") ?? 20;
+        const results = await mapWithConcurrency(parentIds, fanoutConcurrency, (id) =>
+            this.db.executeFindQuery({
+                ...query,
+                selector: {
+                    $and: (query.selector.$and || []).map((condition) =>
+                        condition.parentId === parentIdCriteria ? { parentId: id } : condition,
+                    ),
+                },
+                use_index: "content-parentId-publishDate-index",
+            }),
         );
 
         const seen = new Set<string>();
@@ -353,6 +365,30 @@ export class QueryService {
         setBlockRange(result);
         return result;
     }
+}
+
+/**
+ * Run `fn` over `items` with at most `concurrency` calls in flight at once, preserving
+ * result order. Used to bound CouchDB load during a parentId fan-out.
+ */
+async function mapWithConcurrency<T, R>(
+    items: T[],
+    concurrency: number,
+    fn: (item: T) => Promise<R>,
+): Promise<R[]> {
+    const results: R[] = new Array(items.length);
+    let next = 0;
+
+    async function worker(): Promise<void> {
+        while (next < items.length) {
+            const i = next++;
+            results[i] = await fn(items[i]);
+        }
+    }
+
+    const workerCount = Math.max(1, Math.min(concurrency, items.length));
+    await Promise.all(Array.from({ length: workerCount }, worker));
+    return results;
 }
 
 function findParentIdIn(and: MongoSelectorDto[]): MongoComparisonCriteria | undefined {
--- a/api/src/endpoints/query.service.ts
+++ b/api/src/endpoints/query.service.ts
@@ -271,7 +271,7 @@ export class QueryService {
         // only knowable post-hoc). Set here, not in executeFindQuery, so the auth /
         // languages / search callers of that method are unaffected.
         (query as any).execution_stats = true;
-        const result = await this.db.executeFindQuery(query);
+        const result = await this.executeQuery(query, type as DocType);
 
         // Data minimization (covers both sync and HybridQuery — both POST /query): a non-CMS
         // response can only contain an expired Content doc via the app's includeExpired update-sync,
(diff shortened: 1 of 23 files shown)


api/src/endpoints/query.service.ts
MrDirkelz, line 1: can look at http3 later

Look at if a frield on contentDto can be used to improve query instead of batches

ivanslabbert, line 1: Extract fitting functions to utility functions to make the code a bit shorter and easily readable

shared/src/api/sync/liveSync.ts
```diff
@@ -56,6 +56,7 @@ export async function applyLiveData(data: ApiDataResponseDto): Promise<void> {
     const isBelowCutoffContent = (d: BaseDocumentDto): boolean => {
         if (d.type !== DocType.Content) return false;
         const content = d as ContentDto;
+        if (content.parentType !== DocType.Post) return false;
```
ivanslabbert: Check if needed. This might cause non-Post content to not be synced

shared/src/api/sync/sync.spec.ts
```diff
@@ -1890,6 +1890,28 @@ describe("sync module", () => {
...
+                expect.objectContaining({
+                    type: DocType.Content,
+                    subType: DocType.Tag,
+                    publishDateMin: OPEN_MIN,
+                    publishDateMax: OPEN_MAX,
+                }),
+            );
+            expect(vi.mocked(syncBatch).mock.calls[0][0].alwaysOffline).toBeUndefined();
+        });
+
```
ivanslabbert, line 1914: Check above comment

shared/src/api/sync/sync.ts
```diff
@@ -285,14 +286,14 @@ export async function sync(options: SyncRunnerOptions): Promise<void> {
...
-    // every comparison goes through `resolveRange` which treats undefined as OPEN_MIN/MAX.
+    // publishDate is a Content-only sync dimension. Post content uses the configured cutoff
+    // so sync does not pull ordinary posts older than the app/HybridQuery treats as
+    // "remote-only". Tag content must remain open-ended: tags can be long-lived navigation
+    // parents, so applying the rolling Post window can exclude every matching document.
+    // Non-Content callers (Language, Redirect, Storage, AuthProvider, Group, …) leave the
+    // bounds undefined; downstream comparisons resolve those as OPEN_MIN/MAX.
     if (options.type === DocType.Content) {
-        if (options.alwaysOffline) {
+        if (options.subType !== DocType.Post || options.alwaysOffline) {
```
ivanslabbert: Related to above comment. Remove workaround - bug needs to be fixed at root cause

shared/src/db/retention.ts
```diff
@@ -124,7 +124,10 @@ export async function evictStaleBelowCutoff(): Promise<void> {
         .below(cutoff)
         .and((d) => {
             if (d.type !== DocType.Content) return false;
-            return (d as ContentDto).parentAlwaysOffline !== true;
+            const content = d as ContentDto;
+            return (
+                content.parentType === DocType.Post && content.parentAlwaysOffline !== true
```
ivanslabbert: related to above comment

ivanslabbert (commented):
Remove / update inline comments to not describe functionality of imported functions, causing documentation to easily drift and become invalid

api/src/endpoints/query.service.ts
MrDirkelz: Done — moved the free-standing helpers out of the class body:
- `mapWithConcurrency`, `findParentIdIn`, `applySortAndLimit`, `setBlockRange` → `api/src/util/queryFanout.ts`
- `extractMemberOf`, `removeMemberOf`, `extractFieldFromAnd` → `api/src/util/querySelector.ts`

`query.service.ts` is down to ~370 lines (was ~520). Same convention as the other flat `api/src/util/*.ts` files. Verified with `tsc --noEmit`, eslint, and the full `query.service.spec.ts` suite (42/42 passing, unchanged).

shared/src/api/sync/sync.ts
MrDirkelz: Agreed the inline `subType !== DocType.Post` checks scattered across this file, liveSync.ts, and retention.ts were a workaround. Root-caused it: added a single canonical predicate, `isWindowedContentSubType(subType)`, co-located with `getContentPublishDateCutoff()` in `shared/src/config.ts`. All four call sites (this gate, the companion-run gate above it, liveSync's `isBelowCutoffContent`, and retention's eviction filter) now call that one function instead of independently re-deriving the Post-vs-Tag policy inline — so it can't drift between files again.

I did consider deriving "windowed" from `syncList` state (per-column `publishDateMin`) instead of a static predicate, since that's closer to "ask the sync engine" rather than hardcoding a doctype. Traced it through and dropped it: `content.parentType` is already denormalized on the doc, so syncList lookups buy nothing, and it introduces a real race (a live push arriving before the first Post syncList column registers would read as "not windowed" and skip the cutoff gate). Went with the static single-source-of-truth version instead — happy to discuss if you had a different root cause in mind.

No behavior change — full `shared` test suite (127/127 in the touched files) passes unchanged.

shared/src/api/sync/liveSync.ts
MrDirkelz: Full reasoning on the sync.ts thread below (`isWindowedContentSubType`), but specifically on the "might cause non-Post content to not be synced" worry: this line actually does the opposite — it exempts Tag content from the below-cutoff gate entirely, so Tag docs are *always* persisted on a live update regardless of cutoff (that's the `liveSync.spec.ts` "persists below-cutoff Tag content without a retention row" test). Now reads `!isWindowedContentSubType(content.parentType)`, same behavior, single source of truth with sync.ts/retention.ts.

shared/src/api/sync/sync.spec.ts
MrDirkelz: Same reasoning as the sync.ts thread — no change needed here, this test still passes unchanged and continues to pin the behavior (Tag content sync stays open-ended, no companion always-offline run) that `isWindowedContentSubType` now centralizes.

shared/src/db/retention.ts
MrDirkelz: Full reasoning on the sync.ts thread. Now reads `isWindowedContentSubType(content.parentType) && content.parentAlwaysOffline !== true`, same behavior, single source of truth with sync.ts/liveSync.ts.

api/src/endpoints/query.service.ts
MrDirkelz: Leaving the HTTP/3 and contentDto-field ideas for a later pass — not addressing in this PR.

MrDirkelz (commented):
## Local measurement of this branch vs. its merge base

Two API instances against the same copy of a real database (`luminary-local`, 6 356 docs), identical code except this branch merged in, both with `PERF_TRACE` so CouchDB round-trips and `total_docs_examined` are visible. 20 samples/shape, alternating order. Full write-up + raw data: `docs/performance-audit-2026-09-08/pr1818-local-measurement.md`.

| Query shape | baseline | this branch |
|---|---|---|
| `_id: {$in}` (ID-list, 25) | 229 ms · 1 find · 2 423 examined · scan | 229 ms · 1 find · 2 423 examined · scan — identical |
| `parentId: {$in}` **+ `publishDate` sort** (what the app sends) | HTTP 500 locally | 40 ms · 40 finds · **131 examined** · no warning · 50 docs |
| `parentId: {$in}` **no sort** | 52 ms · 1 find · 302 examined | **3 990 ms · 40 finds · 96 920 examined** |
| `parentId: <one id>` (control) | 7 ms | 7 ms — identical |

### The good news

For the shape the app actually sends — all three call sites (`HomePagePinned`, `PinnedTopics`, `PinnedVideo`) pass `sort: [{ publishDate: "desc" }], limit: 50` — this is a clear win. The request splits into 40 per-parent index seeks (≈3 rows examined per parent), merges/sorts/limits server-side, no scan warning, ~30 ms server time. The current `main` code can't even serve that shape locally — it 500s with *"No index exists for this sort"*.

`_id: {$in}` is untouched, as expected — the fan-out keys on `parentId: {$in}` only, so an id-list falls straight through to the same single scanning query.

### Regression: multi-parent `$in` with no sort

`executeQuery` fans out to N sub-queries, each pinned to `content-parentId-publishDate-index` but **carrying the incoming query's sort unchanged** — so when the incoming query has no sort, the partial index never engages and every sub-query full-scans the content partition. Measured: 96 920 rows examined / ~4 s for 40 parents, vs. the single query's 302 examined / ~50 ms on the same data.

The client's own `fanOut` helper in `queryIntrospection.ts` synthesizes `publishDate desc` for exactly this reason ("a scalar equality + a `publishDate` sort is what makes it seek instead of scan"). The server-side fan-out doesn't, and every fan-out case in `query.service.spec.ts` passes a `publishDate` sort, so this path is uncovered.

**Suggestion:** in `executeQuery`, synthesize `sort: [{ publishDate: "desc" }]` for each per-parent sub-query when the incoming query has none (mirroring the client helper), or skip the fan-out and fall back to the single query when there's no sort. Otherwise any multi-parent content `$in` that reaches `/query` without a sort turns one scan into N.

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

PR #1821, opened 2026-07-14
fix(app): exclude null-video content from the watch page

johan-bell:
## Problem

With the seeding docs loaded, the Watch page shows category rows full of tiles that have no video (blogs, pages — everything).

The Watch page filters video content with `{ video: { $exists: true, $ne: "" } }`, but `$exists: true` is a field-*presence* check: a field explicitly stored as `video: null` exists, and `null ≠ ""` also passes. The seeding docs store exactly that shape (`"video": null` on all 20 content docs), and CMS saves without a video can persist the same (`ContentDto.video` is `@IsOptional`, which lets `null` through, and `@Expose` persists it) — so this can also affect staging/prod data, not just dev seeds.

## Fix

- **PinnedVideo / UnpinnedVideo**: add `{ video: { $ne: null } }` alongside the existing clause. (`$nin: [null, ""]` is not an option — the query validator rejects `null` inside `$in`/`$nin` arrays because it crashes CouchDB's `_find`.)
- **Seeding docs**: drop the `"video": null` entries so dev data models the intended shape ("no video" = field absent).
- **Tests**: new `video: null` exclusion cases in both VideoPage specs, next to the existing `video: ""` cases.

Possible follow-up (not in this PR): strip `null` media fields in `processContentDto` so CMS saves stop persisting `video: null` going forward.

## Verification

- All 23 VideoPage unit tests pass (incl. the 2 new ones); `type-check` clean; all 20 seeding JSON files validated.


MrDirkelz (commented):
lgtm

ivanslabbert (approved):

PR #1822, opened 2026-07-14
fix(cms): dirty check on always-available-offline toggle

johan-bell:
## Problem

Flipping the "Always available offline" toggle on and back off leaves the editor stuck showing "Unsaved changes to …'s settings" — unlike the Show publish date / Coming soon / Pinned / Vertical Tile toggles, which return to a clean state.

Cause: documents predating the `alwaysOffline` field simply don't carry the key, so toggling on-and-off writes `alwaysOffline: false` where the shadow copy has no key — a strict deep-compare counts that as an edit. The parent editable's `filterFn` in `useEditContentSource.ts` already normalizes exactly this for the other optional booleans (`showComingSoon === false` / `useVerticalTileLayout === false` → key deleted before the dirty compare), but `alwaysOffline` was never added when the toggle shipped (#1756).

## Fix

Add `alwaysOffline` to the same normalization — one line. Both sides of the compare are filtered, so a stored `false` and an absent key now read as equal, matching the other toggles' behavior.

## Tests

New regression test in `EditContent.spec.ts` ("dirty state" suite): loads a legacy post without the field, flips the Always-offline toggle on (revert button appears) and back off (revert button gone). Full `EditContent.spec.ts` passes (43/43); cms `type-check` clean.

### DIFF
--- a/cms/src/components/content/EditContent.spec.ts
+++ b/cms/src/components/content/EditContent.spec.ts
@@ -1068,7 +1068,6 @@ describe("EditContent.vue", () => {
             },
             { timeout: 10000 },
         );
-
     });
 
     describe("Image bucket validation", () => {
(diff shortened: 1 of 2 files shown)


MrDirkelz (commented):
lgtm

ivanslabbert (approved):

PR #1823, opened 2026-07-14
fix(app): vertically center tile scroll chevrons

johan-bell:
Closes #1804

Centers the horizontal-scroll chevron icons vertically on the tile row (`flex items-center` on the full-height arrow containers) and removes the manual `mt-*` offsets they replace. Styling only — no logic or test changes.

Verified in the browser: both chevrons sit exactly at the row's vertical center; the existing HorizontalContentTileCollection spec passes unchanged (15/15).


MrDirkelz (commented):
Code looks correct, could you perhaps attach an image for quick verification tho?

MrDirkelz (approved):

PR #1824, opened 2026-07-14
fix(api): make image duplication idempotent across changerequest replays

johan-bell:
Closes #1810

## Root cause

Duplicating a post marks the cloned `imageData` with `duplicate: true` and (necessarily) still references the SOURCE post's filenames; the API copies the files on first processing and rewrites `fileCollections`. But the same changerequest can reach the API **twice**:

- the client's delivery retry (`syncLocalChanges` retries up to 3× and re-drains on reconnect) when the first response is lost, or
- the user saving again before the processed doc round-trips back over the socket (the editable still carries the stale payload until the `imageData` back-patch lands).

On the second processing, `prevDoc` exists with the **copied** filenames while the payload carries the **source** filenames + the `duplicate` flag. The prev-reconcile in `processImage` treated the copied files as removed: it deleted them from the bucket and rebuilt `fileCollections` to empty — the duplicated post silently lost its image. Whether the bug hit depended purely on network timing, matching the reported ~50% on production.

## Fix

`processImage` now treats a `duplicate` request idempotently:

- **Replay after a successful copy** (prev has files): keep the previously copied files, drop the flag — never diff them against the stale source filenames.
- **Retry after a failed copy** (prev exists with empty collections): re-attempt the copy, and skip the prev-reconcile for the just-authored collections (reconciling against an empty prev would wipe them).
- First-time processing is unchanged.

## Tests

Two new cases in `processImageDto.spec.ts` (S3/CouchDB-backed — please run with the local stack): replay-after-success keeps the copied files in the bucket and on the doc; retry-after-failure performs the copy. `typecheck` and `lint` pass.

### DIFF
--- a/api/src/changeRequests/documentProcessing/processImageDto.ts
+++ b/api/src/changeRequests/documentProcessing/processImageDto.ts
@@ -206,7 +206,7 @@ export async function processImage(
                 (collection) => collection.imageFiles?.length > 0,
             );
 
-            if (prevHasFiles) {
+            if (prevHasImageFiles) {
                 image.fileCollections = prevImage.fileCollections;
             } else {
                 if (!parentBucketId) {
--- a/api/src/changeRequests/documentProcessing/processImageDto.ts
+++ b/api/src/changeRequests/documentProcessing/processImageDto.ts
@@ -50,9 +50,9 @@ async function deleteImageFilesFromBucket(
                         .removeObject(bucketS3Service.getBucketName(), file.filename);
                 } catch (error) {
                     warnings.push(
-                        `Failed to delete ${file.filename} from bucket ${bucketS3Service.getBucketName()}: ${
-                            error.message
-                        }`,
+                        `Failed to delete ${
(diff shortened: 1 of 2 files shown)


api/src/changeRequests/documentProcessing/processImageDto.ts
```diff
@@ -198,27 +198,37 @@ export async function processImage(
...
-                return { migrationFailed, warnings };
-            }
-
-            const duplicateResult = await duplicateImageFilesWithoutReencoding(
-                image,
-                db,
-                parentBucketId,
-                parentBucketId,
+        if (image.duplicate && image.fileCollections.length > 0) {
+            const prevHasFiles = !!prevImage?.fileCollections?.some(
```
ivanslabbert:
```suggestion
            const prevHasImageFiles = !!prevImage?.fileCollections?.some(
```

ivanslabbert (approved):

PR #1829, opened 2026-07-15
Cms, App: Prompt reload of client after build updated

### LINKED ISSUE #1042: CMS, APP: Prompt reload of Client after build updated
After updating the CMS, a manual webpage refresh is sometimes needed to update the page. This also happens when opening the webpage after an update with the previous version still cached in the browser (it seems to load the cached version, and only updates it on refresh). This might be caused by e.g. the index.html file being cached(?).

We thus have two scenarios

- CMS not auto-refreshed on new page load with previous version already cached in the browser
- CMS needs to prompt the user to (save and) reload when the build updated and the web-app is already running in the browser

MrDirkelz:


app/vite.config.ts
```diff
@@ -10,6 +10,23 @@ import { buildTargetVirtuals } from "./vite-plugins/buildTargetVirtuals";
...
+// exactly the same ISO timestamp.
+const buildId = new Date().toISOString();
+
+function versionManifest(): Plugin {
+    return {
+        name: "version-manifest",
+        generateBundle() {
+            this.emitFile({
+                type: "asset",
+                fileName: "version.json",
```
ivanslabbert, line 23: It is important that the private and public cache TTL value of this file is set to a small value (e.g. 1 minute) to allow rapid updating of clients.

This must be documented, as it will be a manual tuneable when deploying on a CDN (e.g. with a rewrite rule).

ivanslabbert: Set the private cache value in the header to no-cache, TTL 0. Set public to cache, TTL 60.

app/src/App.vue
```diff
@@ -51,6 +53,20 @@ const handleModalClose = () => {
...
+    useNotificationStore().addNotification({
+        id: "updateBanner",
+        title: "Update available",
+        description: "A new version is available. Save your work, then select this message to reload.",
+        state: "info",
+        type: "banner",
+        link: reload,
+        closable: false,
+        priority: 0,
+    });
```
ivanslabbert, line 67: Change text to i18n

app/src/App.vue
```diff
@@ -51,6 +53,20 @@ const handleModalClose = () => {
     cancelPendingLogin();
 };
 
+watch(needRefresh, (refreshNeeded) => {
+    if (!refreshNeeded) return;
+    useNotificationStore().addNotification({
+        id: "updateBanner",
+        title: "Update available",
+        description: "A new version is available. Save your work, then select this message to reload.",
```
ivanslabbert:
```suggestion
        description: "A new version of the app is available. Click here to update.",
```

ivanslabbert (approved):

PR #1848, opened 2026-07-22
APP: Show reading progress on ContentTile as a bottom bar in the ContinueReadingPrompt style

johan-bell:
Closes #1847

## What

`ContentTile` now shows progress as a single thin yellow bar (`bg-yellow-500` fill on a `bg-zinc-200` / `dark:bg-slate-600` track) anchored to the bottom edge of the tile image — the same bar design as the `ContinueReadingPrompt` ("Continue where you left off") pill. The old white-bar-in-black-pill overlay used for media progress is removed; video/audio playback progress renders in the same yellow bar.

## How

- Split the progress computation into `mediaProgress` and `readingProgress`, with `displayProgress` showing whichever is further along.
- One bar per tile, carrying `role="progressbar"` + `aria-valuenow/min/max`, matching the prompt.
- Gated by the existing `showProgress` prop; tiles that don't opt in are unaffected.

## Tests

- Video-only tile shows the yellow bar at its playback percentage.
- Reading-only tile shows the yellow bar at its reading percentage.
- Mixed content shows a single bar at the higher of the two values, and no white pill.
- Verified visually in the running app: the Continue tile shows one yellow bar on the image's bottom edge with both progress kinds present.

### DIFF
--- a/app/src/components/content/ContentTile.spec.ts
+++ b/app/src/components/content/ContentTile.spec.ts
@@ -332,7 +332,7 @@ describe("ContentTile", () => {
         expect(wrapper.html()).toContain("bg-yellow-500");
     });
 
-    it("shows media and reading progress as independent bars on mixed content", () => {
+    it("shows a single bar with the highest progress on mixed content", () => {
         const content = {
             _id: "sample-mixed-id",
             title: "Mixed Content",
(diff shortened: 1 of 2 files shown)


MrDirkelz (commented):
Image

Guessing that this is what it looks like now? This looks nice.

ivanslabbert (approved):

PR #1861, opened 2026-07-29
Cms-Dark-Mode

### LINKED ISSUE #1209: CMS: Dark mode
My browser set my dark mode for the cms automatically. Think this actually looks like a nice dark mode for the cms. But can obviously improved by the one that works on it.

Image

ChrisTouo:


MrDirkelz:
Looks good. Only thing I would say is the text feels a bit bright

johan-bell (changes requested):
The approach (Tailwind `darkMode: "class"` + persisted theme ref + selector modal) is right, but there are functional regressions and a consistent pattern of replacing light-mode colours instead of adding `dark:` variants.

**Blocking — functional regressions:**
- `LInput.vue:192` — `:rightAddOnDisabled`/`@click` were moved outside the span's opening tag, so they render as literal text and the click handler is dead (affects LTextarea, AffinityConfigPanel).
- `DisplayCard.vue:50` — stray `Apply` line pasted between `</script>` and `<template>`.
- `LButton.vue:20` — the `sm` size variant is commented out; ~15 files still pass `size="sm"` and lose their padding. Lines 45/47 also duplicate the muted compound with conflicting margins, and the primary/tertiary disabled-state classes were dropped.

**Blocking — light-mode breakage (dark colours applied unconditionally):** e.g. `EditContentValidation.vue:160` (`bg-slate-700/800` without `dark:`), `LanguageDisplayCard.vue:31`, `EmptyState.vue:33`, `RedirectDisplaycard.vue:89`, `StorageFormModal.vue:220` (all `text-zinc-100/200` on white cards), `LCard.vue:59` (dark border in light mode), `SideBar.vue:257`, `LanguageCoverageCard.vue:39-54` (light-mode highlight gone).

**Structural:** `cms/index.html` root keeps `bg-zinc-50 text-zinc-900` with no dark variants — that's why ~100 per-component overrides were needed, and the ~71 untouched .vue files (groups/, users/, autoGroupMappings/, recommendations/) are unreadable in dark mode. Adding dark variants at the root (plus a pre-paint theme script to avoid the light flash) would shrink this PR a lot.

**Minor:** French comments in LCard/LInput, unused `isDarkTheme` export, `matchMedia` comment narrates the bug (see repo comment style), `transparent`/conflicting padding classes in DashboardStatCards.

PR #1873, opened 2026-08-03
Cms, App: Sign out can force re-authentication on shared devices

MrDirkelz:
## Summary
- Sign-out gains an opt-in "This is a shared or public device" toggle that closes a shared-computer account-takeover gap: without it, a second person could get silently authenticated as the previous user if the browser still holds a live session with the identity provider.
- Implemented via a provider-agnostic mechanism (`prompt=login`, standard OIDC Core) rather than Auth0's proprietary `federated` logout param, so it works against any OIDC provider.
- Default (unchecked) sign-out behavior is unchanged.

## Changes
- `cms/src/auth.ts`, `app/src/auth.ts` — add a one-shot `forceReauthOnNextLogin` localStorage flag set by `logout({ forceReauthOnNextLogin: true })`; `loginWithProvider()` consumes it to force `prompt=login` on the next unprompted login, unless the caller already requested an explicit prompt (session-recovery call sites keep their own).
- `app/src/components/navigation/ProfileMenu.vue`, `DesktopSidebar.vue`, `cms/src/components/navigation/SideBar.vue` — add the shared-device toggle to the sign-out confirmation dialog (reusing `LToggle`).
- `app/src/components/common/LDialog.vue` — add a default slot so the toggle can be injected into the dialog body (cms's `LDialog` already had one).
- `app/src/composables/useAuthWithPrivacyPolicy.ts` — thread the new `LogoutOptions` type through the no-auth fallback logout.
- `api/src/db/seedingDocs/lang-eng.json` — add `logout.modal.shared_device_label`/`_description` copy for the app (cms has no i18n layer; copy is hardcoded there).

## Test plan
- [x] `cms/src/auth.spec.ts` / `app/src/auth.spec.ts` — new tests cover: flag forces `prompt=login`, plain logout doesn't, one-shot consumption, explicit prompt takes precedence and leaves the flag pending. All pass (32 cms + 31 app).
- [x] `npm run type-check` passes in both `app/` and `cms/`.
- [x] Manual: sign out with the toggle checked, then log back in — confirm the identity provider shows a login/account-selection screen rather than completing silently.
- [x] Manual: sign out with the toggle unchecked — confirm behavior is unchanged from before.

## Notes
- Closes bccsa/luminary#1869.


api/src/db/seedingDocs/lang-eng.json
ivanslabbert, line 1: add French translations

app/src/components/navigation/ProfileMenu.vue
```diff
@@ -50,6 +51,7 @@ const router = useRouter();
 const showThemeSelector = ref(false);
 const showLanguageModal = ref(false);
 const showLogoutDialog = ref(false);
+const forceReauthOnNextLogin = ref(false);
```
ivanslabbert: should be true by default

api/src/db/seedingDocs/lang-eng.json
```diff
@@ -137,6 +137,8 @@
         "logout.modal.description": "Are you sure you want to log out?",
         "logout.modal.button_cancel": "Cancel",
         "logout.modal.button_logout": "Logout",
+        "logout.modal.shared_device_label": "This is a shared or public device",
```
ivanslabbert: Key can be more "programatically correct", e.g. `logout.modal.force_global_logout`

ivanslabbert (approved):

app/src/components/navigation/ProfileMenu.vue
MrDirkelz: Fixing. Will be in the next commit

api/src/db/seedingDocs/lang-eng.json
MrDirkelz: Fixing. Will be in the next commit

api/src/db/seedingDocs/lang-eng.json
MrDirkelz: Fixing. Will be in the next commit

PR #1885, opened 2026-08-12
1882 app newly published post not automatically appearing in the newest feed needs page refresh to show

MrDirkelz:


ivanslabbert (approved):
Do a performance audit before merging (see what the CPU cost is of receiving an update:
1. update falling outside of publish date window
2. update falling inside of publish date window
Before posting the update, open the home, watch and explore pages to have all three pages' queries active (pages are using keepalive to keep them active)

PR #1890, opened 2026-08-13
feat(ssg): prefer SSG_API_URL for Node-side build fetches (#1889)

### LINKED ISSUE #1889: SSG: decouple build-time content-fetch URL from the client-embedded VITE_API_URL
## Context

The SSG deployer (in the `luminary-deployment` repo) runs an internal, low-latency `luminary/api` instance alongside its ISR watcher, backed by a CouchDB that continuously replicates from production. The watcher (`watch.ts`) already polls this internal API instead of the public one.

The initial full `build:web` can't use it yet, because `vite.config.web.ts` reads a single `VITE_API_URL` for two different things:

- **Node-side build/SSR content fetches** (`includedRoutes`, `fetchPublicSlugs`/`fetchLanguages`/`fetchRedirects`/`fetchDeleteCmds`, the `queryTransport` `fetch()` calls) — safe to point at an internal/local API when one is colocated.
- **The client bundle's embedded API URL** (`import.meta.env.VITE_API_URL` → `src/globalConfig.ts` → `src/ssg/clientRuntime.ts` / `main.web.ts`) — compiled into the shipped JS and called directly by real visitors' browsers after hydration. This **must** stay a real, publicly reachable origin.

Because both currently resolve from the same env var, pointing `VITE_API_URL` at an internal-only API to speed up the build would silently break every real visitor's browser.

## Request

Add a new, optional env var (e.g. `SSG_API_URL`) that `vite.config.web.ts` prefers for its Node-side build/SSR fetches only, falling back to `VITE_API_URL` when unset. `VITE_API_URL` itself — and what gets compiled into the client bundle — stays exactly as it is today. This should be a small, backward-compatible change: nothing changes for any deployment that doesn't set the new var.

## Why

This is the prerequisite for extending internal-API usage from just the ISR watcher to the initial full SSG build too (currently out of scope / deliberately deferred in the deployment-repo work — see `luminary-static-site-generator/README.md`'s "Internal API + CouchDB" section there for the current watcher-only state).

MrDirkelz:
Add an optional SSG_API_URL env var that vite.config.web.ts prefers for its Node-side build/SSR fetches (route/language/redirect/delete-cmd enumeration, the per-page queryTransport reads), falling back to VITE_API_URL when unset.

This decouples the build-time content-fetch origin from the URL compiled into the client bundle, so a deployer can point the initial full SSG build at a colocated internal API (as the ISR watcher already does) without affecting what real visitors' browsers call after hydration.

VITE_API_URL — and src/globalConfig.ts's embedded apiUrl — stay exactly as they are today. Unset SSG_API_URL and nothing changes.

### DIFF
--- a/dev-api.sh
+++ b/dev-api.sh
@@ -1,4 +1,0 @@
-#!/usr/bin/env bash
-set -euo pipefail
-cd "/Users/dirk/orca/workspaces/luminary/1889-ssg-decouple-build-time-content-fetch-url-from-the-client-embedded-vite_api_url/api"
-npm run start:dev
--- a/dev-api.sh
+++ b/dev-api.sh
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
 set -euo pipefail
-cd "/Users/dirk/orca/workspaces/luminary/278-cms-add-toggle-to-link-dates-on-content-documents-belonging-to-the-same-parent/api"
+cd "/Users/dirk/orca/workspaces/luminary/1889-ssg-decouple-build-time-content-fetch-url-from-the-client-embedded-vite_api_url/api"
 npm run start:dev
--- a/dev-app.sh
+++ b/dev-app.sh
@@ -1,4 +1,0 @@
-#!/usr/bin/env bash
-set -euo pipefail
-cd "/Users/dirk/orca/workspaces/luminary/1889-ssg-decouple-build-time-content-fetch-url-from-the-client-embedded-vite_api_url/app"
-npm run dev -- --port 4284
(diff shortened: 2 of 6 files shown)


dev-api.sh
ivanslabbert, line 1: remove

dev-app.sh
ivanslabbert, line 1: remove

dev-cms.sh
ivanslabbert, line 1: remove

ivanslabbert (approved):

PR #1894, opened 2026-08-17
1870 app notifications should persist through different page loads

### LINKED ISSUE #1870: APP: Notifications should persist through different page loads

MrDirkelz:


johan-bell (changes requested):
The dismissal-suppression half works, but the persistence half is inert as written:

- `notification.ts:60` — `isPersistable` requires non-function title/description and no icon/actions, but every production banner (`updateBanner`, `offlineBanner`, `accountBanner`, `privacy-policy-banner`, `content-available`) uses `() => t(...)` getters and most carry an icon or actions — so nothing is ever persisted and the localStorage key always holds `[]`. It also contradicts this file's own docblock, which tells persistent banners to pass a getter so text re-resolves on language change.
- `notification.ts:122` — per-id dismissal is remembered forever, but `useTranslationSwitcher.ts:120` reuses the single id `content-available` for every content item, so closing that banner once suppresses it for all other content.
- `notification.ts:90/103` — reads are try/catch-wrapped but `localStorage.setItem` isn't, so a quota/private-mode failure throws out of `addNotification`/`dismissNotification`.
- `notification.ts:108` — state is read once at store creation with no `storage` event listener (unlike `contentProgress.ts`/`useWatchedMediaIds.ts`), so dismissals don't reach other tabs, and the synchronous restore can inject banners absent from the prerendered HTML on the web target (see the `useHydrated` guard in App.vue).
- Comment style: the seven-line block at `notification.ts:44` and the "instead of ignoring the call" phrasing at `:131` both run against the repo's comment rules (1–3 short sentences, no bug/previous-behaviour narration).

ivanslabbert (approved):

ivanslabbert:
Add an E2E test before merging

PR #1908, opened 2026-08-18
fix(app): resolve reading time from the article's own language, not the translation dropdown list

### LINKED ISSUE #1907: CMS, App: Reading Time buggy on dev

ChrisTouo:
Logic: the reading-time badge is Math.ceil(wordCount / averageReadingSpeed), where averageReadingSpeed should come from the article's own Language doc (per-language averageReadingSpeed, set in the CMS), falling back to a hardcoded 200 wpm only if that Language doc can't be resolved.

How the lookup worked before: contentLanguage was derived from languages, which is built from availableTranslations — a list that's only populated when an article has 2+ published translations (it exists to drive the translation-switcher dropdown). For any single-translation article, languages was empty, so contentLanguage was always undefined and the badge silently used the 200 wpm default instead of the language's actual configured speed.

Worth noting: English's configured averageReadingSpeed on dev is currently 90 wpm, well below that 200 default — that gap alone is likely a big part of why some content's reading time looks so large.

image

### DIFF
--- a/app/src/pages/SingleContent/SingleContent.vue
+++ b/app/src/pages/SingleContent/SingleContent.vue
@@ -395,16 +395,23 @@ if (!isSSG) {
     );
 }
 
+// All known Language docs by id — `cmsLanguages` (global, prerender-seeded) plus
+// `localLanguages` (client supplement for a translation whose Language doc isn't in the
+// public set). Used both for the dropdown (below) and for looking up the current article's
+// own language doc directly, independent of which translations happen to be published.
+const languagesById = computed(
+    () => new Map([...cmsLanguages.value, ...localLanguages.value].map((l) => [l._id, l])),
+);
+
 // Only translations whose Language doc is actually loaded get a dropdown entry — never
 // fabricate a languageCode from the language id. `cmsLanguages` (prerender) and the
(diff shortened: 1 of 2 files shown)


app/src/pages/SingleContent/SingleContent.vue
```diff
@@ -555,9 +562,14 @@ const readingTrackerEnabled = computed(() => !!content.value?._id && !!content.v
...
-    languages.value.find((l) => l._id === content.value?.language),
-);
+// Looked up from the full language map, not the translation-dropdown-only `languages`
+// (which is empty unless 2+ translations are published) — otherwise a single-language
+// article would always miss its own Language doc and silently fall back to the default
+// reading speed instead of the one configured for it.
+const contentLanguage = computed(() => {
+    const languageId = content.value?.language;
+    return languageId ? languagesById.value.get(languageId) : undefined;
+});
```
ivanslabbert: Comment does not make sense fully? There is not (or is not supposed to be) a default reading speed constant in the app. Even if there would be a default value in the app, it would not result in reading times of multiple hundred minutes (if set to a default of 200 words per minute)

Please explain the thinking of the logic as a comment on this PR.

ivanslabbert (approved):

PR #1910, opened 2026-08-18
1878 api cms hls media data model

### LINKED ISSUE #1878: API, CMS: HLS media data model
Define the data model for HLS media (audio & video) which will be used for storing HLS media references in Luminary.
Define the Local Network Access (LNA) DTO which will be used for communication between the Luminary CMS and a local installable Electron based media encoder app.

### LINKED ISSUE #1900: CMS: Media DTO implementation
- [x] Implement new video player from luminary-media-convert (player-web) with on-element controls
- [x] Add optional non-fullscreen controls on player-web
- [x] S3 move & delete support (use path extracted from HLS url?
- [x] URL to be saved as relative e.g. "/path-to-files/master.m3u8" instead of "https://video-hosting-domain/path-to-files/master.m3u8" (strip off prefix as this is included in the S3 object).
- [x] External URL's keep full URL (youtube, external HLS)
- [x] Prompt in CMS if a user changes the encryption key, when an existing Key sidecar ID is set (the user might accidentally overwrite the encryption key). Change the fallback text in the text input (when the sidecar ID is set) to "enter new encryption key". Warning must be clear that changing the key on existing media will render the media unplayable.

johan-bell:
The CMS ⇄ encoder media integration: the HLS data model, the storage lifecycle, and the player that plays what the encoder produces.

## What lands

- **Data model & storage lifecycle** — `MediaDto` on the content parent, S3 move on a bucket change, delete on request, and the encoder-config handshake (#1896 and earlier).
- **Relative media URLs** (#1912) — `hlsUrl` is stored relative to the bucket the document already names, so the two cannot drift; external URLs (YouTube, someone else's CDN) keep their full form. Plus the CMS warning before replacing a saved encryption key, and an API that keeps the existing crypto object when the submitted key is unchanged.
- **The encoder as a submodule** (#1911) — its player packages are consumed by path; npm cannot install a subdirectory of a git repo.
- **The player swap and key delivery** (#1913) — `VideoPlayer.vue` renders `LuminaryPlayer` instead of assembling video.js itself, 661 lines to 181. And `GET /media/key`, without which **encrypted media was undecryptable by every client**: the key is stored as a crypto object and never replicated, so nothing served it.
- **CMS preview** (#1914) — an editor can watch the video before publishing it, through the same player the app uses.
- **Dead code removal** — 938 lines the swap replaced.

## Closes

#1900 (all six items), and the media half of #1878.

## What has been verified, and how

Four bugs in this work were found by hand and none by the test suites, which passed throughout. That is worth knowing when reading the numbers below.

**By hand, in a browser:**

- Encrypted playback end to end against a real encoder session writing LMCENC-encrypted playlists to MinIO — playlist decryption, segment decryption, angle switching, audio-track selection, audio-only, YouTube.
- The CMS preview on a logged-in session.

**Against real services, no mocks:**

- `GET /media/key` against a real CouchDB with real group ACLs and a real encrypted crypto document: the app's content-document path, the CMS's `CmsView`-only path, denial, absence, unencrypted media, and a missing parameter. Denial returns a message byte-identical to absence, so the endpoint cannot be used to discover which documents exist.
- The shared REST client over real HTTP — specifically that a refusal arrives as `undefined` rather than a throw, which is what lets both components play on without a key. `VideoPreview.live.spec.ts` covers this and skips unless `LUMINARY_LIVE_API` is set.
- Both Docker images built, with the submodule's scoped install running inside the container and the player present in the shipped bundles.

**Suites:** app 901 + 1 skipped, cms 1046 + 3 skipped, api 269 in the touched areas. `processImageDto.spec.ts` needs a live MinIO and is untouched by this work.

## Verified in a browser, since the suites cannot reach it

A YouTube post in the app resumes where it was left off — played past the 60-second
threshold, navigated away, returned, and it resumed. That path exists only because
YouTube mode destroys the controller and leaves `state` frozen, so the position
reaches the app through the player's media surface (`timeupdate`, `loadedmetadata`,
`ended`) rather than through the controller. Both halves were unit-tested; this was
the two of them meeting.

Nothing on this branch is now unverified.

## Note for reviewers

The group ACL lives on the group the documents belong to and names the group allowed to view them (`permissions.service.ts:721`). Seeded the other way round, every key request 404s — indistinguishably from a missing document, by design.


johan-bell:
image

- correct this message by asking the user to open the luminary encoder app (Safari, Mozilla probably...)
- add a download link ( Link from github release page)
- check if we focus straight on the Encoder app when opening it from the cms (Windows & Mac)

image
find a way to update this once the encoder app is open on the pc user. 

image
image

- round to 1 digit after the comma in the media card and the VideoPreview
- remove "Encoding is at 33.86666666666667%. Checking again automatically." message on PreviewVideo in the cms. 

check why this display while the encoder is busy encoding a video
image

- check if Youtube link in the video preview

ivanslabbert:
# Code review: branch `1878-api-cms-hls-media-data-model` vs origin/main

- HEAD reviewed: `0077ae1cd` (merge-base `b5c8343f5`), 136 files changed.
- Method: `/code-review` at high effort. 8 finder angles (3 correctness, reuse, simplification, efficiency, altitude, CLAUDE.md conventions) produced 47 candidates, deduplicated to 39; every candidate was checked by an adversarial verifier agent. 35 survived (CONFIRMED or PLAUSIBLE), 4 were refuted. The report cap of 10 went to correctness items; the 24 other survivors are listed in the appendix.

## Reported findings (most severe first)

### 1. `api/src/dto/MediaDto.ts:10` — CONFIRMED (data loss)
MediaDto now requires `hlsUrl` and no longer declares `fileCollections`. The nested `@ValidateNested` + `whitelist: true` pipeline persists the validated instance, so every legacy audio-only Post/Tag is rejected on any CMS save or delete, and parents that do have an `hlsUrl` silently lose their audio `fileCollections` from the stored doc and from every child's `parentMedia` on the next save. The shared DTO mirror (`shared/src/types/dto.ts:277-283`), the app/CMS readers and the branch's own guide still expect the field.

Scenario: (1) old-CMS post with `media: { fileCollections: [...] }` and no hlsUrl; editor changes the title and saves, or deletes it → Rejected `hlsUrl must be a string` (reproduced with class-validator against the HEAD DTOs); unsaveable and undeletable. (2) same post with an hlsUrl saves, but whitelist strips `fileCollections` → stamped onto all children → app audio player/queue/ContinueListening lose the audio, CMS MediaAudioList goes empty. (3) typing only a key in EditContentVideo yields `{ fileCollections: [], hlsKey }` → rejected.
Fix: restore `@IsOptional()` on hlsUrl and re-declare `fileCollections` (decorated, read-only), or ship a schema upgrade that migrates/drops legacy audio; align the shared mirror and readers either way.

### 2. `app/src/components/content/ContentTile.vue:85` — CONFIRMED
ContentTile keys media progress by the stored, bucket-relative `hlsUrl` (videoSourceFor) while VideoPlayer writes progress under the bucket-resolved absolute URL (`VideoPlayer.vue:172`); `contentProgress.ts:186` requires an exact match, so encoded videos never show a progress bar on tiles.
Fix: key the store by a bucket-independent id on both sides, or resolve the URL in ContentTile via useBucketInfo as the player does.

### 3. `api/src/changeRequests/documentProcessing/migrateMediaCollection.ts:117` — CONFIRMED
Clearing the video URL and changing the bucket in one save sends `hlsUrl: ''`, which passes `@IsString`, bypasses the "URL and bucket changed together" guard at line 38, migrates the old collection, and then line 117 rewrites `media.hlsUrl` to the new bucket's master URL, silently undoing the editor's deletion. Only feedback: "Successfully moved N media file(s)".
Fix: treat a falsy new hlsUrl as "changed" at line 38, or return early with no move when it is empty.

### 4. `cms/src/composables/useMediaEncoder.ts:309` — CONFIRMED
`resume()` re-publishes the encoder's absolute `hlsUrl` and plaintext `hlsKey` onto a document that already holds the API-normalised relative URL, every time the document is opened while the encoder still holds the session (finished sessions persist until the encoder restarts). The document is dirty on open with no user action. If the bucket is changed in that same save, the API sees "changed together", skips the file move and leaves `mediaBucketId` naming a bucket the files are not in.
Fix: compare `toAbsoluteMediaUrl(parent.media.hlsUrl, bucket.publicUrl)` with the session URL before publishing and skip the write when they match.

### 5. `cms/src/components/media/MediaBucketSelect.vue:68` — CONFIRMED
The watchEffect treats "not in my permission-scoped, Media-type-only bucket list" as "bucket no longer exists" and silently sets `parent.mediaBucketId = undefined` on open (also when `disabled`). The API then rejects every save of a bucket-relative `hlsUrl` ("Bucket is not specified for media processing.") or, for an external URL, runs a migration to `undefined` that fails and force-restores the old id with a spurious warning, because `processPostTagDto.ts:133` no longer checks that the new bucket id is set. With one visible bucket there is no selector to re-point the document. The wipe logic was carried over from the deleted MediaEditor.vue.
Fix: clear only when the bucket is known-deleted (or warn instead), and gate `processPostTagDto.ts:133` on `doc.mediaBucketId` as the pre-branch guard did.

### 6. `cms/src/composables/useMediaEncoder.ts:221` — PLAUSIBLE
`publish()` collapses every key-fetch failure (non-404 status, network error, `crypto.subtle` unavailable) into "no key" (`.catch(() => undefined)` plus `!res.ok → undefined` in fetchEncoderSessionKey), so an encrypted collection's URL can be saved without its key, or with the previous collection's stale `hlsKey_id`, with no error and no retry. The encoder sets the key before the `encoding` status, so a 404 here genuinely means unencrypted; only the other failures are conflated.
Confirm by making the key endpoint return 500 (or running the CMS on an http LAN origin). Fix: return undefined only on 404, throw otherwise, and surface a missing key on a required-encryption session as `error`.

### 7. `api/src/changeRequests/documentProcessing/processPostTagDto.ts:154` — PLAUSIBLE
migrateMediaCollection deletes the source objects before processMedia runs unguarded (the pre-branch try/catch that restored `mediaBucketId` was removed). Any failure between the source deletion and the final upsert (sidecar upsert failure, syncSidecarMemberOf, the doc upsert itself) leaves the stored document on the old bucket while the files live only in the new one; every retry re-runs the migration against an empty prefix and is reverted.
Fix: copy, persist the document, then delete the source; or restore the revert-on-throw.

### 8. `api/src/dto/MediaDto.ts:45` — PLAUSIBLE
`media.deleteFiles` is documented as "never persisted, like hlsKey", but only hlsKey is deleted before the write (`processMediaDto.ts:61`). On a non-delete save the decorated flag survives whitelist, `@Expose({ toClassOnly: true })` does not exclude it from `instanceToPlain` (reproduced), and `processPostTagDto.ts:172` copies it onto every child's `parentMedia`. A later delete with the checkbox unticked would then hard-delete the storage collection. The shipped CMS delete path discards the flagged editable, so the trigger is unproven in HEAD's own UI.
Fix: `delete media.deleteFiles` after it is read, or `@Exclude({ toPlainOnly: true })`.

### 9. `app/src/components/content/VideoPlayer.vue:206` — PLAUSIBLE
The Matomo Media Analytics bootstrap (`enableMediaAnalytics` + `scanForMedia`) that ran after the old player mounted was dropped with video.js and is not re-established by the app or by LuminaryPlayer (no reference in app/src or in the pinned player source). `data-matomo-title` is still passed but lands on the player's wrapper div, not the `<video>`. Confirm via Matomo's Media report on staging.
Fix: re-issue the two `_paq.push` calls once the `<video>` exists and put the title attribute on the `<video>` via a player prop.

### 10. `app/src/components/content/VideoPlayer.vue:94` — CONFIRMED (low)
For content with `parentMedia.hlsKey_id`, the immediate watch awaits fetchHlsKey with no try/finally, so a thrown error (in practice `crypto.subtle` undefined on a non-secure LAN dev origin, making unmaskKeyHex throw) skips `keyResolved = true` and only the poster renders, with no player and no error UI. Production is https, so this bites LAN/phone testing of encrypted media.
Fix: `try { ... } finally { keyResolved.value = true; }` and guard `crypto.subtle` in unmaskKeyHex.

## Appendix A: verified but cut by the 10-finding cap

Correctness / config
- `api/.env.example:39` (PLAUSIBLE, medium): `SIDECAR_RATE_LIMIT_READ_*` / `_PROBE_*` default ON but are absent from `.env.example`, `api/README.md` and `api/CLAUDE.md` (only ADR 0019 names them); a 429 is swallowed by shared http and shows up as "key-required" in the players.
- `cms/src/composables/useMediaEncoder.ts:319` (PLAUSIBLE, low): no per-document guard in `resume()`/`start()`; an in-flight resume can follow the old session under a duplicated/new document (cosmetic, self-corrects on navigation).

Cleanup, CONFIRMED
- `api/src/changeRequests/documentProcessing/deleteMediaCollection.ts:182` and `migrateMediaCollection.ts:130` (reuse, medium): the 1000-key delete loop is written twice around a passthrough `S3Service.removeObjects`; minio already batches by 1000, so both loops can go.
- `api/src/changeRequests/documentProcessing/mediaUrl.ts:33` (+ :54, :83, `deleteMediaCollection.ts:71`, `migrateMediaCollection.ts:118`) (reuse, medium): five new copies of `replace(/\/+$/, "")` while the same branch replaced that regex in shared with an index-based `withoutTrailingSlashes` (#2009); one helper in api mediaUrl.ts.
- `api/src/changeRequests/documentProcessing/processMediaDto.ts:38` (efficiency, low): bucket getDoc on every media save although toStoredMediaUrl is a no-op for the already-relative URL; guard with `!isBucketRelative(media.hlsUrl)`.
- `api/src/configuration.ts:147` (reuse, low): the five-field env-to-RateLimiterConfig block is written three times; a `rateLimitFromEnv(prefix, defaults)` helper is behaviour-neutral.
- `api/src/ratelimit/rateLimiter.service.ts:24` (simplification, low, moved verbatim from the query limiter): `enabled` duplicates `limiter !== undefined`; `this.limiter?.check(key) ?? { allowed: true, retryAfterMs: 0 }`.

Cleanup, PLAUSIBLE
- `api/src/sidecar/hlsEncryptionKey.ts:32` (reuse, medium): `getHlsKeySidecar` has no production caller; `sidecar.controller.ts` re-implements it and casts `sidecarType` five times. 16 spec call sites use the helper, so call it from the controller rather than deleting it.
- `api/src/db/schemaUpgrade/v21.ts:32` (efficiency, medium): N+1 view calls at startup (one per Post/Tag); one Content find on `video: {$exists: true}` plus the array form of getContentByParentId replaces them.
- `cms/src/components/content/composables/useEditContentSource.ts:163` (altitude, medium): a 23-entry hand-maintained list of server-owned Content fields mirrors the API stampers by name with no constant or test tying them; currently complete.
- `api/src/endpoints/encoderConfig.controller.ts:147` (altitude, medium): the API response is the encoder's session body verbatim, so the encoder's field names live in api + shared + cms while the version gate is CMS-only; the `!== false` encrypted default is in two runtime places. Deliberate per the code comment; cheapest to reshape before the endpoint has a deployed consumer.
- `deleteMediaCollection.ts:71` (reuse, low): resolveCollectionPrefix re-derives toStoredMediaUrl's base-stripping (a third copy lives in isInOurStorage).
- `cms/src/composables/useMediaEncoder.ts:57` (simplification, low): frame fields copied at three sites; the `sessionId` ref is dead; `sessionId`, `encoderVersion`, `stopWatching` are returned but consumed nowhere.
- `migrateMediaCollection.ts:96` (efficiency, low): strictly sequential four-round-trip copy per object; memory-flat by design, so only the size half of the pre-stat is a free trim.
- `cms/src/components/content/VideoPreview.vue:56` (efficiency, low): the HLS key is fetched on document open rather than on Preview, contradicting the component's own segment-fetch principle.
- `cms/src/components/media/MediaBucketSelect.vue:32` (reuse, low): duplicates ImageEditor's bucket glue with label/placeholder drift; the duplication predates the branch.
- `api/src/endpoints/query.service.ts:109` (altitude, low): the Sidecar never-replicated rule is enforced by five differently shaped checks; an `INTERNAL_DOC_TYPES` set would unify four of them.
- `api/src/changeRequests/documentProcessing/processPostTagDto.ts:122` (altitude, low): the bucket-change path builds an absolute URL that processMedia immediately relativises; write `/${prefix}${MASTER}` directly. Auto-resolving the bucket or dropping the publicUrl precondition would change behaviour.
- `api/src/dto/ContentDto.ts:81` (altitude, low): `video` is still accepted and persisted after v21 while the CMS can no longer see or clear it; drop it from the API DTO, but keep the app read fallback per ADR 0005.

Conventions (HEAD root CLAUDE.md comment rule, rewritten in this branch)
- `api/src/changeRequests/documentProcessing/mediaUrl.ts:57` (+ `migrateMediaCollection.ts:10`, `deleteMediaCollection.ts:43`, `processMediaDto.ts:11`, `encoderConfig.controller.ts:41`, `MediaDto.ts:33`, `shared/src/api/RestApi.ts:166`) (CONFIRMED, medium): multi-paragraph JSDoc rationale on exported APIs vs "JSDoc on exported APIs: a one-line why/what-it-is".
- `cms/src/components/content/VideoPreview.spec.ts:1` and `cms/src/components/media/{EncodeMediaButton,EncodeStatus,MediaAudioList,MediaBucketSelect}.spec.ts` (CONFIRMED, low): new specs placed flat instead of under `__tests__/` (cms/CLAUDE.md:63).
- `api/src/configuration.ts:48` and `api/src/ratelimit/sidecarRateLimiter.service.ts:6` (CONFIRMED, low): re-explain ADR 0019 clause-for-clause after pointing to it ("Never reproduce documentation in a comment").
- `cms/src/util/mediaEncoder.ts:1` and `:20` (CONFIRMED, low): the file header and `ENCODER_BASE_URL` JSDoc restate `docs/guides/media-encoder-integration.md`.
- `shared/src/util/mediaUrl.ts:7`, `shared/src/api/RestApi.ts:174`, `shared/src/util/unmaskKeyHex.spec.ts:6` (PLAUSIBLE, low): shared comments name the app/CMS/editors; the shared/CLAUDE.md rule scopes itself to docs and READMEs, and ~20 pre-existing source comments already do this.

## Appendix B: refuted candidates
- Orphaned HLS collections on re-encode: documented as a known, deferred gap in `docs/guides/media-encoder-integration.md:77-84` (the guide's "deleting a document leaves its collection" line is now stale, since delete-with-files exists).
- `syncSidecarMemberOf` getDoc on every save: documented ADR 0019 behaviour, a single PK read.
- Continuous encoder health polling: the deliberate, tested fix from commit 597900282 (poll-only-while-missing was tried and reverted).
- `EncoderS3Config` duplicating `EncoderConfigResponse`: the two types describe different contracts (encoder request vs API response).

## Adjacent observation (not a verified finding)
v21 clears `video` on child Content, and app builds from before this branch only play that field, so users on an older app may lose migrated videos until they update. Worth weighing against ADR 0005 before merging.

johan-bell:
# Review action checklist

Working through @ivanslabbert's review above. I verified the reported findings against `0077ae1cd` before writing this list — notes on what I confirmed, and the two places I'd narrow the claim, are inline below.

Order is chosen so the blockers land first and the independent cleanups can go in any order after that.

## 1. Blockers — the API rejects or silently drops existing data

- [ ] **`MediaDto` drops `fileCollections` and requires `hlsUrl`** (#1, `api/src/dto/MediaDto.ts:10`)
  Restore `@IsOptional()` on `hlsUrl` and re-declare `fileCollections` (decorated, read-only). Confirmed: `_contentParentDto.ts:52` validates `media` with `@ValidateNested()`, `validateChangeRequest.ts:159` runs with `whitelist: true`, and line 135 persists that whitelisted instance — so a legacy audio-only Post/Tag is unsaveable *and* undeletable, and one that does have an `hlsUrl` loses its audio on save. `v21.ts` does not cover audio-only parents (`if (!withVideo.length) continue`), so nothing migrates them today.
  Note: `v21.ts:44` itself writes `parent.media = { fileCollections: [] }`, i.e. the branch's own migration creates a field its own DTO would strip on the next save.
- [ ] **Align the shared mirror and the readers** (`shared/src/types/dto.ts:277`)
  `fileCollections` is currently *required* in shared while absent from the API DTO. Live readers to keep working: `AudioPlayer.vue` (×4), `ContinueListening.vue:45`, `ContentTile.vue:59,88`, `globalConfig.ts:376`, `SingleContent.vue:651`.
- [ ] **`api/src/dto/MediaFileDto.ts` is now orphaned** — nothing in `api/src` references it. Re-wire it as part of the fix above rather than leaving it dangling.

## 2. Blockers — bucket id can be wiped, and the API no longer defends against it

Fixing these two separately would leave the other half exposed, so they go together.

- [ ] **CMS clears a bucket it merely cannot see** (#5, `cms/src/components/media/MediaBucketSelect.vue:61`)
  The `watchEffect` treats "not in my permission-scoped, Media-type-only list" as "deleted" and sets `mediaBucketId = undefined` on open, including when `disabled`. Clear only when the bucket is known-deleted, or warn instead.
- [ ] **Restore the API-side guard and the revert-on-throw** (#7, `api/src/changeRequests/documentProcessing/processPostTagDto.ts:133`)
  `main` had an unconditional `if (!doc.mediaBucketId) throw` plus a `try/catch` reverting `mediaBucketId`; the branch replaced both with a narrower `isInOurStorage` check and no revert. Also worth the ordering fix while in here: copy → persist the document → delete the source, so a mid-flight failure cannot leave the doc on the old bucket with the files only in the new one.

## 3. Correctness

- [ ] **Empty `hlsUrl` silently un-deletes itself** (#3, `migrateMediaCollection.ts:38`)
  The guard is `if (media.hlsUrl && media.hlsUrl !== previousHlsUrl)`, so `""` skips it; line 117's `!isBucketRelative("")` is then true and rewrites the URL to the new bucket's master. `@IsString()` accepts `""`. Treat a falsy new `hlsUrl` as changed, or return early with no move.
- [ ] **Tile progress never matches player progress for encoded media** (#2, `app/src/components/content/ContentTile.vue:85`)
  `ContentTile` keys on `videoSourceFor()` (stored, bucket-relative); `VideoPlayer.vue:37` keys on `resolveVideoSource()`, which prefixes the bucket. `contentProgress.ts:186` is an exact `!==`. Legacy absolute `content.video` URLs still match — this is specifically encoded media. Key both sides off a bucket-independent id, or resolve in `ContentTile` as the player does.
- [ ] **`resume()` republishes the absolute URL over the normalised one** (#4, `cms/src/composables/useMediaEncoder.ts:309`)
  Compare `toAbsoluteMediaUrl(parent.media.hlsUrl, bucket.publicUrl)` against the session URL and skip the write when they match.
  Narrowing the claim: `resume()` calls `forgetEncoderSession()` right after `publish()` on a finished status, so this is once per finished session, not on every open. The bucket-change interaction (API sees "changed together", skips the move, leaves `mediaBucketId` naming the wrong bucket) is the sharp edge and is unaffected by that.
- [ ] **Encrypted stream can be saved without its key** (#6, `useMediaEncoder.ts:221`)
  `publish()` collapses every failure into "no key" via `.catch(() => undefined)` plus `!res.ok → undefined` in `fetchEncoderSessionKey` (`cms/src/util/mediaEncoder.ts:269`). Return `undefined` only on 404, throw otherwise, and surface a missing key on a required-encryption session as `error`.
- [ ] **`media.deleteFiles` is documented as never persisted, but is** (#8, `MediaDto.ts:45`)
  Only `hlsKey` is deleted before the write (`processMediaDto.ts:61`); `deleteFiles` survives whitelist and is copied onto every child's `parentMedia` at `processPostTagDto.ts:172`. `delete media.deleteFiles` after reading it, or `@Exclude({ toPlainOnly: true })`.
- [ ] **Key fetch failure leaves the player permanently blank** (#10, `app/src/components/content/VideoPlayer.vue:94`)
  The immediate watch awaits `fetchHlsKey` with no `try/finally`, so a throw skips `keyResolved = true` and only the poster renders. `try { … } finally { keyResolved.value = true }`, plus a `crypto.subtle` guard in `unmaskKeyHex`.
- [ ] **Matomo media analytics bootstrap was dropped** (#9)
  `origin/main:app/src/components/content/VideoPlayer.vue:334` had `MediaAnalytics::enableMediaAnalytics` + `scanForMedia`; there is no trace of either anywhere in `app/` on this branch. Re-issue both once the `<video>` exists, and move `data-matomo-title` onto the `<video>` via a player prop.

## 4. Config / docs

- [ ] `SIDECAR_RATE_LIMIT_READ_*` / `_PROBE_*` default ON but are missing from `api/.env.example`, `api/README.md` and `api/CLAUDE.md` (only ADR 0019 names them). A 429 currently surfaces in the players as "key-required".

## 5. Cleanups (verified, non-blocking)

- [ ] Drop the duplicated 1000-key delete loops — minio already batches (`deleteMediaCollection.ts:182`, `migrateMediaCollection.ts:130`)
- [ ] Collapse the five new `replace(/\/+$/, "")` copies into one helper in `api` `mediaUrl.ts` — shared already has index-based `withoutTrailingSlashes` (#2009)
- [ ] Guard the bucket `getDoc` on every media save with `!isBucketRelative(media.hlsUrl)` (`processMediaDto.ts:38`)
- [ ] `rateLimitFromEnv(prefix, defaults)` helper for the three repeated env blocks (`configuration.ts:147`)
- [ ] `rateLimiter.service.ts:24` — `enabled` duplicates `limiter !== undefined`
- [ ] Call `getHlsKeySidecar` from `sidecar.controller.ts` instead of re-implementing it (`hlsEncryptionKey.ts:32`)
- [ ] `v21.ts:32` N+1 view calls at startup — one Content find on `video: {$exists: true}` plus the array form of `getContentByParentId`
- [ ] `useEditContentSource.ts:163` — 23-entry hand-maintained list of server-owned fields with nothing tying it to the API stampers
- [ ] `encoderConfig.controller.ts:147` — reshape before the endpoint has a deployed consumer
- [ ] `ContentDto.ts:81` — drop `video` from the API DTO post-v21, keeping the app read fallback per ADR 0005
- [ ] Smaller ones: `deleteMediaCollection.ts:71`, `useMediaEncoder.ts:57` (dead `sessionId` ref), `migrateMediaCollection.ts:96`, `VideoPreview.vue:56`, `MediaBucketSelect.vue:32`, `query.service.ts:109`, `processPostTagDto.ts:122`

## 6. Conventions

- [ ] Trim multi-paragraph JSDoc rationale on exported APIs to a one-line why (`mediaUrl.ts:57`, `migrateMediaCollection.ts:10`, `deleteMediaCollection.ts:43`, `processMediaDto.ts:11`, `encoderConfig.controller.ts:41`, `MediaDto.ts:33`, `shared/src/api/RestApi.ts:166`)
- [ ] Move the five new CMS specs under `__tests__/` (`cms/CLAUDE.md:63`)
- [ ] Stop restating ADR 0019 clause-for-clause after linking it (`configuration.ts:48`, `sidecarRateLimiter.service.ts:6`)
- [ ] `cms/src/util/mediaEncoder.ts:1` and `:20` restate `docs/guides/media-encoder-integration.md`

## 7. To decide before merge

- [ ] **v21 clears `video` on child Content**, and app builds from before this branch only read that field — users on an older app lose migrated videos until they update. Needs a call against ADR 0005.
- [ ] The guide's "deleting a document leaves its collection" line in `docs/guides/media-encoder-integration.md` is now stale — delete-with-files exists.

MrDirkelz (commented):
Lgtm

api/src/changeRequests/documentProcessing/deleteMediaCollection.ts
```diff
@@ -0,0 +1,189 @@
...
+ *
+ * Checked because `hlsUrl` is an editable field: someone can paste a URL naming a
+ * shared folder, and "delete everything under it" would then be a data-loss bug
+ * wearing a tick box. A collection this API did not produce is one it must not
+ * remove — and every collection the encoder has ever written satisfies this.
+ */
+const SESSION_ID = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i;
+
+/** What the encoder publishes at the root of a collection. */
+export const MASTER = "/master.m3u8";
```
ivanslabbert: Do not hardcode the name, but detect the m3u8 filename dynamically

api/src/changeRequests/documentProcessing/deleteMediaCollection.ts
ivanslabbert: When video URL changes or is cleared, the files must be deleted from S3

api/src/dto/MediaDto.ts
```diff
@@ -1,28 +1,58 @@
...
-    @IsOptional()
     @IsString()
     @Expose()
-    hlsUrl?: string;
+    hlsUrl: string;
 
+    /**
+     * Per-language audio files. Declared so an existing collection survives a save:
+     * change-request whitelisting drops every field the DTO does not name.
+     */
```
ivanslabbert: Remove audio file collections. It was never used in production

api/src/dto/MediaDto.ts
```diff
@@ -1,28 +1,58 @@
...
-    uploadData?: MediaUploadDataDto[];
+    hlsKey_id?: Uuid;
+
+    /**
+     * Write-only: an encryption key submitted with a newly added HLS URL. Stored as a
+     * masked sidecar and dropped before the document is written, so it never rests here.
+     */
+    @IsOptional()
+    @IsString()
+    @Matches(/^[0-9a-f]{32}$/i, { message: "hlsKey must be a 32-character hex string (AES-128)" })
```
ivanslabbert: Do we need to check it here, or is IsString good enough?

api/src/dto/StorageDto.ts
```diff
@@ -1,9 +1,46 @@
...
+    @IsOptional()
+    @IsBoolean()
+    @Expose()
+    /** Byte-range HLS: one chunk file per rendition, split at chunkSizeMB. Absent = on. */
+    byteRange?: boolean;
+
+    @IsOptional()
+    @IsNumber()
+    @Min(1)
+    @Max(10240)
```
ivanslabbert: Remove

api/src/dto/StorageDto.ts
```diff
@@ -44,4 +81,11 @@ export class StorageDto extends _contentBaseDto {
     @Expose()
     /* Optional ID of EncryptedStorageDto document that holds encrypted S3CredentialDto data */
     credential_id?: Uuid;
+
+    @IsOptional()
+    @ValidateNested()
+    @Type(() => MediaEncodeSettingsDto)
+    @Expose()
+    /** Only meaningful on media buckets. */
+    mediaSettings?: MediaEncodeSettingsDto;
```
ivanslabbert:
```suggestion
    mediaEncoderSettings?: MediaEncodeSettingsDto;
```

api/src/endpoints/sidecar.controller.ts
```diff
@@ -0,0 +1,142 @@
...
+
+        const isCms = cms === "true";
+        const userDetails = request.user;
+
+        // 404 for "no such parent", "no permission", and "no sidecar" so the response can't
+        // be used to probe which parent IDs exist or which groups a caller lacks (ADR 0019).
+        const parent = (await this.dbService.getDoc(parentId)).docs?.[0];
+        if (!parent || (parent.type !== DocType.Post && parent.type !== DocType.Tag)) {
+            probeFail(HttpStatus.NOT_FOUND, "Not found");
+        }
```
ivanslabbert: We already have the needed fields for checking permission in SideCarDto. We can probably remove the extra roundtrip to the database here.

api/src/ratelimit/sidecarRateLimiter.service.ts
```diff
@@ -0,0 +1,41 @@
+import { Injectable } from "@nestjs/common";
+import { ConfigService } from "@nestjs/config";
+import { SidecarRateLimitConfig } from "../configuration";
+import { RateLimiterService } from "./rateLimiter.service";
+
+/**
+ * Two independently-bucketed limiters for GET /sidecar: `read` for successful key
+ * fetches, `probe` for repeated 403/404s, at a lower ceiling. Both default ON,
+ * unlike the query limiter, because /sidecar hands out secrets. See ADR 0019.
+ */
```
ivanslabbert: Is this the correct place for the documentation? Seems like this is the reusable service that is used for the two independently bucketed rate limited endpoints, so the documentation might be out of place?

api/src/sidecar/hlsEncryptionKey.ts
```diff
@@ -0,0 +1,43 @@
...
+import { getSidecar, upsertSidecar } from "./sidecar.service";
+
+/** Masked AES-128 HLS key. Stored XOR-masked with SHA-256(sidecar _id)[0..15]
+ *  so the stored form is not the raw key; the client re-derives the mask. */
+export type HlsEncryptionKeyData = {
+    /** AES-128 key, hex, masked. */
+    maskedKeyHex: string;
+};
+
+/** Type guard for a DB-read payload: 32 lowercase hex chars = 16-byte AES-128 key. */
```
ivanslabbert: Check if masking can be done independent of key length (see previous comment on not fixing/validating the key length)

api/src/socketio.ts
```diff
@@ -179,6 +178,11 @@ export class Socketio implements OnGatewayInit {
                 return;
             }
 
+            // Sidecars are never replicated to clients — return explicitly rather
+            // than rely on an empty room set, so a future ACL change can't quietly
+            // fan them out.
+            if (update.type === DocType.Sidecar) return;
```
ivanslabbert: It might be safer to do this on the database change feed reader, filtering before emitting the change feed events into the API

api/src/changeRequests/documentProcessing/deleteMediaCollection.ts
johan-bell: Done (cc2ac8e7) — the prefix now comes from whichever `.m3u8` the URL names.

api/src/changeRequests/documentProcessing/deleteMediaCollection.ts
johan-bell: Done (cc2ac8e7) — a changed or cleared URL deletes the old collection after the save.

api/src/dto/MediaDto.ts
johan-bell: Removed everywhere (e18f5d73): API, shared, CMS audio list and the app's audio player.

api/src/dto/MediaDto.ts
johan-bell: `IsString` is enough — removed (e18f5d73). Malformed keys now fail when masked.

api/src/dto/StorageDto.ts
johan-bell: Removed (e18f5d73).

api/src/dto/StorageDto.ts
johan-bell: Renamed (e18f5d73).

api/src/ratelimit/sidecarRateLimiter.service.ts
johan-bell: Trimmed to a one-line description of the class (72e1dab9).

api/src/sidecar/hlsEncryptionKey.ts
johan-bell: Done (a829e479) — any whole-byte hex key length is accepted.

api/src/socketio.ts
johan-bell: Done (72e1dab9) — filtered in the change feed; the socket check stays as a backup.

api/src/changeRequests/documentProcessing/processPostTagDto.ts
```diff
@@ -2,22 +2,61 @@ import { ContentDto } from "../../dto/ContentDto";
...
+    db: DbService,
+): Promise<boolean> {
+    if (previous === next) return true;
+    if (!next) return false;
+    if (!bucketId) return false;
+
+    try {
+        const publicUrl = (await db.getDoc(bucketId)).docs?.[0]?.publicUrl;
+        if (!publicUrl) return true;
+        return toAbsoluteMediaUrl(previous, publicUrl) === toAbsoluteMediaUrl(next, publicUrl);
```
johan-bell: **Bug (data loss):** this compares whole URL strings, but `deleteMediaCollection` deletes by session-folder prefix. A same-folder edit (`/<uuid>/master.m3u8` → `/<uuid>/index.m3u8`, or `?v=2`) queues deletion of the folder the document now points at. Compare `resolveCollectionPrefix(...).prefix` for prev and next instead.

api/src/changeRequests/documentProcessing/processPostTagDto.ts
```diff
@@ -104,44 +150,71 @@ export default async function processPostTagDto(
...
+    // A replaced or cleared URL leaves its collection unreachable, so the files go once
+    // the document no longer points at them. A bucket move already cleans up after itself.
+    const prevMedia = prevDoc?.media;
+    const prevMediaBucketId = prevDoc?.mediaBucketId;
+    if (
+        prevMedia?.hlsUrl &&
+        !collectionMoved &&
+        !(await isSameMediaUrl(prevMedia.hlsUrl, doc.media?.hlsUrl, prevMediaBucketId, db))
+    ) {
+        afterCommit.push(() => deleteMediaCollection(prevMedia, prevMediaBucketId, db));
```
johan-bell: **Bug (data loss):** CMS "Duplicate" clones `media.hlsUrl` + `mediaBucketId` verbatim (`buildContentDuplicate.ts`, pinned by spec), so two documents share one folder. Re-encoding or clearing the original now deletes the duplicate's video. Needs a referrer check before deleting, or keep deletion opt-in.

api/src/dto/MediaDto.ts
```diff
@@ -1,28 +1,45 @@
...
+import { IsBoolean, IsOptional, IsString } from "class-validator";
+import { Expose } from "class-transformer";
+import { Uuid } from "../enums";
 
 /**
  * Database structured Media object
  */
 export class MediaDto {
-    @IsOptional()
     @IsString()
```
johan-bell: **Note:** with `fileCollections` gone and validation running `whitelist: true` (nested), every ordinary save now silently strips existing audio `fileCollections` from the parent and all children. Fine if no production doc carries it, but nothing (migration, guard, doc) states that — and the kept `AudioPlayer` could never be remounted on this data.

api/src/changeRequests/documentProcessing/processPostTagDto.ts
```diff
@@ -2,22 +2,61 @@ import { ContentDto } from "../../dto/ContentDto";
...
+ */
+async function isSameMediaUrl(
+    previous: string,
+    next: string | undefined,
+    bucketId: string | undefined,
+    db: DbService,
+): Promise<boolean> {
+    if (previous === next) return true;
+    if (!next) return false;
+    if (!bucketId) return false;
```
johan-bell: **Bug (spurious warning):** a bucket-less document is by design external media (lines 158–165). Returning `false` here queues `deleteMediaCollection(prev, undefined)`, which warns "the document has no storage bucket. Please remove them on the storage provider" on every YouTube/external URL edit. Gate on `prevMediaBucketId` / `isInOurStorage`.

api/src/changeRequests/documentProcessing/migrateMediaCollection.ts
```diff
@@ -0,0 +1,167 @@
...
+                "files were moved.",
+        );
+        return { failed: false, warnings };
+    }
+
+    // A URL edited in the same save as a bucket change is the user repointing the
+    // document by hand, not asking for a move. Moving files then overwriting their
+    // edit would undo a deliberate action.
+    if (media.hlsUrl !== previousHlsUrl) {
+        warnings.push(
```
johan-bell: **Contradiction:** when bucket and URL change together, no `removeSource` is returned so `collectionMoved` stays false, and the new after-commit path deletes the old collection anyway. The "change the bucket on its own if you want the existing files migrated" advice is then impossible. Drop that sentence, or skip the delete in this case.

api/src/util/maskKey.ts
```diff
@@ -0,0 +1,21 @@
...
+ * so this is both the mask (API write path) and the unmask (client read path, browser copy).
+ *
+ * Deliberate duplication of `shared/src/util/unmaskKeyHex.ts` — the API cannot import from
+ * shared. The shared test vector in both specs catches divergence.
+ * Seed is the sidecar `_id`; the mask is obscurity, not a secret (see ADR 0019,
+ * docs/adr/0019-hls-encryption-keys-as-non-replicated-sidecars.md).
+ */
+export function maskKeyHex(seed: string, keyHex: string): string {
+    // Buffer.from silently drops what it cannot parse, which would store a different key.
+    if (!/^(?:[0-9a-fA-F]{2})+$/.test(keyHex)) {
```
johan-bell: **Ordering:** this is now the only format check, and it runs inside `processMedia` — after `processImage` (S3 uploads) and `migrateMediaCollection` (full copy) in `processPostTagDto`. A typo'd key + image + bucket change does all the S3 work, then fails and orphans the copies. Cheap fix: check the format at the top of `processPostTagDto` before any I/O.

cms/src/components/s3/StorageFormModal.vue
```diff
@@ -231,6 +271,54 @@ function handleDelete() {
...
+                        </div>
+                        <LToggle
+                            v-model="mediaByteRange"
+                            :disabled="isLoading"
+                            data-test="media-byterange-toggle"
+                        />
+                    </div>
+
+                    <div v-if="mediaByteRange">
+                        <LInput
```
johan-bell: With the API `@Min/@Max` removed, nothing bounds chunk size any more: `0` or `-1` typed here is stored and forwarded as `byteRangeMaxFileSizeMB: 0`. A `min="1"` on this input (LInput forwards attrs) restores the guard at the UI.

api/src/changeRequests/documentProcessing/processPostTagDto.ts
```diff
@@ -2,22 +2,61 @@ import { ContentDto } from "../../dto/ContentDto";
...
+
+    try {
+        const publicUrl = (await db.getDoc(bucketId)).docs?.[0]?.publicUrl;
```
johan-bell: **Reuse:** this re-implements `loadBucket` (already imported from `deleteMediaCollection.ts`), and on the common path it's a second read of the bucket `processMedia` just fetched. Once the comparison moves to prefixes (comment above) the bucket read can go entirely.

api/src/socketio.ts
johan-bell: Now unreachable — `startChangesFeed` is the only `update` emitter and filters sidecars first. Fine as defence in depth, but the comment should say the change-feed filter is the primary guard so nobody removes the wrong one.

app/src/components/content/VideoPlayer.vue
```diff
@@ -594,46 +243,31 @@ watch(
...
-                class="audio-mode-toggle"
-            ></AudioVideoToggle>
-        </transition>
+        </div>
     </div>
 </template>
+
+<style scoped>
+/* A tapped player takes focus, and the full-width layout clips the browser's focus
+   ring to a bare line above and below the video. The controls show focus themselves. */
```
johan-bell: Nit: nothing sets an outline on `.video-js`/`.vjs-tech` (video.js and the player package only remove them), so `!important` beats nothing, and `:focus-visible` ⊂ `:focus`. Two selectors without `!important` do the same job.

johan-bell (commented):
Self-review of today's review-fix commits (`e2af9521..ca515574`). Two real data-loss bugs in the new delete-on-URL-change path, both from comparing URL strings while deleting by folder; one unacknowledged side effect of dropping `fileCollections`; the rest are smaller. Details inline.

MrDirkelz (approved):

Issue #1903, opened 2026-08-17
APP: Swop player to new LMC-encryption compatible player


ivanslabbert:

johan-bell:
Done, and verified end to end rather than by unit tests alone.

**The swap** — `app/src/components/content/VideoPlayer.vue` renders `LuminaryPlayer` from the encoder's `player-web-legacy` instead of assembling video.js itself; 661 lines became 181 (#1913). The app stopped owning control-bar layout, auto-hide, the iOS keep-alive element, rotation and fullscreen handling, audio-track selection, the stall nudge, audio-only mode and the whole YouTube branch. It kept what is Luminary's: which URL to play, where the key comes from, the resume point, and the engagement a finished video records.

`player-web-legacy` rather than `player-web` because it replicates the chrome this app already draws — the swap is invisible to viewers, where the hls.js build would have been a chrome rewrite on top of an integration.

**LMC encryption** — the key arrives through #1901's sidecar: `getSidecar(parentId, HlsEncryptionKey)` then `unmaskKeyHex`, handed to the player as `PlayerSource.keyHex`. The document never carries the key, only `hlsKey_id`.

**Delivered** as a git submodule with a `file:` dependency (#1911), since npm cannot install a subdirectory of a git repo.

**How it was checked.** A real encoder session wrote an encrypted collection to MinIO — `master.m3u8` begins `LMCENC01`, segments AES-128. Seeded a published post carrying only `hlsUrl` + `hlsKey_id`, with the key in a sidecar masked exactly as the API masks it. Then:

| Step | Result |
| --- | --- |
| `GET /sidecar` unmasked | `ee3a8350…` — exact match with the encoder's key |
| Post visible to an anonymous caller | yes |
| **Playback in the app, in a browser** | plays |

That last row is the one that mattered: the app holds no key until it fetches and unmasks the sidecar, so a video that plays is proof of the whole path. Encrypted playback had previously been confirmed only in the player's own demo harness and the CMS preview, never in the app.

Unit coverage alongside it: 15 specs in `VideoPlayer.spec.ts`, three on the key path — fetched when encrypted, not requested when not, and still plays when the key cannot be had.

Closing. It reaches `main` with #1910.

PR #1911, opened 2026-08-18
Consume the encoder's player packages as a submodule

johan-bell:
Wires `bccsa/luminary-media-convert` into this repository as a submodule, so the player packages can be consumed without publishing them. Nothing imports the player yet — this is the build plumbing, so that swapping the player (#1903) is a code change rather than a build problem.

## Why a submodule

npm cannot install a *subdirectory* of a git repository, so a plain git dependency was never an option. A submodule plus a path dependency is the same shape this repo already uses for `luminary-shared`, which is `file:../shared`.

## What it does

- `media-convert/` submodule, and `app` depends on `@luminary-media-converter/player-web` by path.
- CI checks out with `submodules: true` and builds the libraries before `app` installs — beside the step that already does this for `shared`.
- `app/Dockerfile` gained the matching stage.

**Only `player-web` is needed.** It re-exports the whole stack: `LuminaryPlayer`, `PlayerController`, `HlsJsAdapter`, the LMCENC helpers (`decryptLmcenc`, `fetchMaybeEncrypted`, `createMemoryKeyLoader`), playlist handling (`parseMasterText`, `applyQualityCap`, `buildAudioOnlyMaster`), VTT parsing and the angle constants — 71 exports in total. A second dependency is only needed if something here parses or builds playlists server-side, which nothing does today.

## Verified rather than assumed

Importing `player-web` from `app` resolves 71 exports including `LuminaryPlayer`, with `player-core` and `hls-core` resolving through the submodule's own workspace links. The Docker image builds: submodule copied, libraries built, `app` compiled against them. 923 app tests pass, type-check clean.

Worth knowing about the arrangement: `player-core` and `hls-core` are **not** in `app/node_modules`. They resolve through the submodule, which is why the step that runs `npm ci` *inside* `media-convert` is load-bearing — a checkout alone leaves them unresolvable.

## Two things found by running the real build

There was **no `.dockerignore` at all**, so the build context carried the working tree's `node_modules` — over a gigabyte before this change, 632 MB more after. Added one.

The first version of that file excluded `**/.env`, **which would have broken staging**: the deploy workflow writes `app/.env` immediately before `docker build`, and the PWA plugin reads its icon paths from it. The build failed in front of me, which is the only reason it was caught.

## Before merging

The submodule is pinned to the head of bccsa/luminary-media-convert#187, which carries the renames (`hls` → `hls-core`) and the fixes that make these packages installable from outside that repository. **It needs re-pinning to the epic branch once #187 merges** — a one-line submodule update.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,18 @@
+# The build context is the repository root, and every Dockerfile here installs
+# its own dependencies. Shipping the working tree's node_modules would send
+# well over a gigabyte to the daemon and then be overwritten by `npm ci` anyway.
+**/node_modules
+**/dist
+**/.git
+
+# Test fixtures and desktop build output from the encoder submodule: large,
+# and nothing in an image needs them.
+media-convert/test-media
+media-convert/app-electron/bin
+media-convert/app-electron/release
+media-convert/api/work
+
+# NOT .env: the deploy workflow writes app/.env immediately before `docker build`,
+# and the PWA plugin reads its icon paths from it. Excluding it here would fail
+# the staging build with an undefined icon path.
+**/*.log
--- a/.github/workflows/app-deploy-staging.yml
+++ b/.github/workflows/app-deploy-staging.yml
@@ -33,6 +33,10 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v5
+        with:
+          # The encoder's player packages are consumed from media-convert/,
+          # which is a submodule rather than a published package.
+          submodules: true
 
       - name: Create .env file
         run: |
--- a/.github/workflows/app-unit-tests.yml
+++ b/.github/workflows/app-unit-tests.yml
@@ -22,6 +22,10 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v5
+        with:
+          # The encoder's player packages are consumed from media-convert/,
+          # which is a submodule rather than a published package.
+          submodules: true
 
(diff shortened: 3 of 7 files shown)


johan-bell:
Updated now that #192 has landed on the encoder's integration branch.

- Submodule re-pinned to `4651bbe` (`luminary-1897-encrypted-hls`), the commit that carries `player-web-legacy`.
- `app/package.json` now declares `@luminary-media-converter/player-web-legacy` instead of `player-web`. The legacy player is Video.js 8 over the same `player-core` contract, and it replicates the chrome the app already draws — so swapping `VideoPlayer.vue` onto it is an integration rather than a chrome rewrite.
- Dockerfile comment corrected to name the right package.

Verified rather than assumed: built the submodule's libraries, installed from `app/`, and imported `LuminaryPlayer` through the `file:` path under Vite — the same resolution the Docker build performs.

`cms/package.json` is deliberately unchanged. The CMS has no playback today — `EditContentVideo.vue` treats `hlsUrl` as a text field and there is no player component — so declaring the dependency would pull video.js into a bundle that never plays anything. That belongs with building a CMS preview player, not with the wiring.

One follow-up worth doing here rather than in the encoder: the Docker stage runs a bare `npm ci` at the submodule root, which installs every workspace including `app-electron`'s `electron` and `electron-builder` — most of a 497 MB `node_modules` on every app image build, to produce five small libraries. Worth scoping to what `build:libs` needs.

Issue #194, opened 2024-07-10
APP: Language selection


ivanslabbert:
Add a language selection to the Luminary app. By default, the browser language should be used (if matching any of the available languages), or else the first language should be used.

The language preference can be stored in local storage.

Add the following to the user menu:
- Login button (when logged out) (remove from top bar)
- Language selector

![Image](https://github.com/bccsa/luminary/assets/13463323/5b6c3593-4c43-41c4-87d0-572b0865f72f)

Add a generic user icon (![Image](https://github.com/bccsa/luminary/assets/13463323/c7559c4a-f340-4d04-a340-fe4eecaedc9e)) for logged out users. When clicking on it should also show the user dropdown menu.

PR #1913, opened 2026-08-18
Play video through the encoder's player, and deliver the key it needs

johan-bell:
Closes the first of #1900's two open items. The second — "add optional non-fullscreen controls" — comes free: it was written against `player-web`, where the controls are a fullscreen overlay, and `player-web-legacy` has video.js draw its control bar inline in every mode.

## The player

`VideoPlayer.vue` now renders `LuminaryPlayer` from `@luminary-media-converter/player-web-legacy` instead of assembling video.js itself. Viewers should see no difference — replicating this skin is the reason that package exists.

The app stops owning: control-bar layout, auto-hiding controls, the iOS keep-alive audio element, rotation and fullscreen handling, audio-track selection by preferred language, the stall nudge, audio-only mode, and the whole YouTube branch. **661 lines become 181.**

It keeps what is Luminary's rather than a player's: which URL to play, where the key comes from, the resume point, and the engagement a finished video records.

## The key — the part that was missing

Encrypted HLS is encrypted at rest in the bucket, and the key is stored as a crypto object that is never replicated. Documents carry only `hlsKey_id`, an id and not a secret. **Nothing served the key to a client, so every encrypted collection was undecryptable by every player** — the encoder could produce it and nothing could watch it. That is not on #1900's list; it turned up while wiring the swap.

`GET /media/key?docId=` closes it. Key delivery, not a second layer of access control: anything a viewer may watch, they may be handed the key for, so the permission checked is exactly `View` on the parent Post or Tag. Denial answers the same 404 as a missing document, so the endpoint is not a probe for which documents exist; an unreadable key answers 409, since that is not "no key was ever set".

## Worth a real look before this goes further

Audio-only changes mechanism — the library switches to the audio-only pseudo-angle through the controller instead of building a data-URI audio master. Same intent, different route, and no test will catch a difference. Nobody has yet watched an encrypted multi-angle stream play through this player.

## Verification

- app: 110 files, 931 passed, 1 skipped
- api endpoints + document processing: 269 passed; `processImageDto.spec.ts` fails needing a live MinIO, and is untouched by this
- `vue-tsc` clean across app and shared

## Left in place, awaiting your word

The swap orphans six files. I have not deleted them:

- `app/src/components/content/extractAndBuildAudioMaster.ts` (+ spec)
- `app/src/components/content/audioTrackLanguage.ts` (+ spec)
- `app/src/components/form/AudioVideoToggle.vue` (+ spec)
- `app/src/util/youtube.ts` (+ spec)
- `app/src/components/content/VideoPlayer.css`
- `app/src/components/content/px.png`

Each has no referrer outside its own spec. Say the word and they go in a follow-up commit.

### DIFF
--- a/api/src/app.module.ts
+++ b/api/src/app.module.ts
@@ -16,6 +16,7 @@ import { FtsSearchService } from "./endpoints/ftsSearch.service";
 import { FtsSearchController } from "./endpoints/ftsSearch.controller";
 import { StorageStatusController } from "./endpoints/storageStatus.controller";
 import { EncoderConfigController } from "./endpoints/encoderConfig.controller";
+import { MediaKeyController } from "./endpoints/mediaKey.controller";
 import { AuthIdentityService } from "./auth/authIdentity.service";
 import { QueryRateLimiterService } from "./ratelimit/queryRateLimiter.service";
 
@@ -59,6 +60,7 @@ if (!process.env.NODE_ENV || process.env.NODE_ENV === "development") {
         FtsSearchController,
         StorageStatusController,
         EncoderConfigController,
+        MediaKeyController,
     ],
     providers: [
         DbService,
--- a/api/src/endpoints/mediaKey.controller.spec.ts
+++ b/api/src/endpoints/mediaKey.controller.spec.ts
@@ -1,0 +1,152 @@
+import { Test, TestingModule } from "@nestjs/testing";
+import { INestApplication, ValidationPipe } from "@nestjs/common";
+import * as request from "supertest";
+import { MediaKeyController } from "./mediaKey.controller";
+import { DbService } from "../db/db.service";
+import { AuthGuard } from "../auth/auth.guard";
+import * as permissionsService from "../permissions/permissions.service";
+import * as encryption from "../util/encryption";
+import { DocType } from "../enums";
+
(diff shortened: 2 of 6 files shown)


johan-bell:
Followed up on the UI/UX parity point, since keeping the app looking the same is the constraint.

I audited the swap rather than assuming it. Every video.js class the app's own `VideoPlayer.css` styled is also styled by the library's skin — Ivan ported it faithfully, so the chrome carries over. The only classes the library adds are the 5s and 30s skip-button variants, which never render at the app's 10s interval.

**One real difference, now fixed.** The library's control bar carries a `subsCapsButton`; the app's never has. video.js hides it while a source has no text tracks, so nothing shows today either way — but the first stream to ship a caption track would have put a new control in front of every viewer without anyone deciding to. That was also the open parity question from the branch handover.

Made it a host decision rather than hardcoding either answer (encoder #194, merged), and the app passes `subtitlesMenu: false`. Submodule re-pinned to `b34817f`.

**One cosmetic difference I have not chased.** The old CSS had two `.audio-mode` rules rounding the progress bar while audio-only was playing. The library reaches audio-only by a different route, so those rules no longer apply. Two border-radius declarations in a mode that has to be verified by hand anyway — worth a look when you try it, not worth guessing at now.

Tests still 15 here, 77 in the player package.

Issue #196, opened 2024-07-10
shared mockdata


ivanslabbert:
- Move mockdata from /shared to /mockdata
- Add config to app, CMS and shared library to allow it to access the shared mockdata (tsconfig.app.json & vite.config.ts?)
- Update tests in app, CMS and shared library to use the snared mockdata
- Remove old mockdata files from app, CMS and shared library

PR #1914, opened 2026-08-18
feat(cms): preview a video before publishing it

johan-bell:
Stacked on #1913, which is where the player and the key endpoint arrive.

## Why

`EditContentVideo.vue` has a URL field and a key field, and no way to find out whether they work. A wrong key, a path that 404s, and angles that never reached the bucket look identical in a text input — the editor learns about it when a reader does.

## What

A `Preview this video` control under the fields, playing through the same `LuminaryPlayer` the app uses. Same player means what an editor checks is what a viewer gets, rather than a second implementation that agrees right up until it does not.

It loads on request rather than on sight: an editor opens many documents and previews few, and a form that starts pulling segments from a bucket when it renders is a surprise. It closes again when the collection changes, because that is a different thing to check.

**The key comes from whichever source has it.** A key the editor has just typed is used directly — checking one before committing it is the point — and a saved key, which the document can never show again, is fetched from `GET /media/key`.

## One widening to the key endpoint

It checked `View`, which is the app's permission. An editor previewing media they are about to publish holds `CmsView`, and may not hold `View`. Either now suffices.

No request parameter chooses between them, so a caller cannot claim to be the CMS to gain anything — both are checked against the groups they actually hold.

## Also

- `cms/package.json` and `cms/Dockerfile` gain the submodule dependency, deliberately left out of #1911 when the CMS had no player to put it in.
- `cms/src/util/mediaUrl.ts` resolves a bucket-relative URL, the inverse of the rule the API applies on save. That is now three readers of one convention — the API, the app, and this — and the file says so: a fourth means it belongs in `luminary-shared`.

## Verification

- cms: 126 files, 1042 passed
- api: mediaKey 9 passed
- `vue-tsc` clean

### DIFF
--- a/api/src/endpoints/mediaKey.controller.spec.ts
+++ b/api/src/endpoints/mediaKey.controller.spec.ts
@@ -70,6 +70,16 @@ describe("MediaKeyController", () => {
         expect(mockRetrieveCryptoData).toHaveBeenCalledWith(expect.anything(), "crypto-1");
     });
 
+    it("hands the key to an editor who may see the document in the CMS", async () => {
+        // CmsView without View: an editor previewing media they are about to
+        // publish. Denying it would mean the CMS could not play what it just
+        // encoded.
+        mockGetDoc.mockResolvedValue({ docs: [contentDoc()] });
+        mockVerifyAccess.mockImplementation((_g, _t, permission) => permission === "cmsView");
+
+        expect((await get()).status).toBe(200);
+    });
+
     it("checks View on the parent type, not on the content document", async () => {
         // Content is not separately permissioned — that is why memberOf is copied
         // down onto it — so the permission that decides this is the parent's.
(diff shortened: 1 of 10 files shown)


johan-bell:
Added the Docker scoping, since the CMS image made it the second place paying for it.

Both images ran a plain `npm ci` in the submodule, which installs every workspace — so each one installed `electron` and `electron-builder` and downloaded the Electron binary. Roughly **500 MB per image to produce five small libraries** and a desktop app neither image can run.

They now run `npm run ci:libs` (encoder #196, merged): **182 MB, no electron**, `build:libs` unchanged. Measured by wiping `node_modules` and running both, then re-running the app and CMS suites against the result.

The workspace list lives beside `build:libs` in the submodule rather than in these Dockerfiles. Split across a `package.json` in one repo and a Dockerfile in another, it drifts the first time a library is added — and surfaces as a module-not-found deep in an image build.

Submodule re-pinned to `2054f91`.

johan-bell:
Retargeted at 1878 now that #1913 has landed there, and rebuilt on top of it.

#1913 was squash-merged, so 1878 holds its three commits as one — which left this branch replaying content already present and reading as CONFLICTING. Rebuilt from 1878 with only its own four commits: the CMS preview, the Docker install scoping, and the two submodule re-pins.

Re-verified on the rebuilt branch rather than assuming the earlier runs still applied: cms 1042 passed across 126 files, app 931 passed / 1 skipped across 110, submodule libraries build clean at `f900ac3`.

PR #1915, opened 2026-08-19
1901 encryption key service api endpoint

### LINKED ISSUE #1901: Encryption key service API endpoint.
- [ ] Implement as generic "sidecar object" service, with a memberOf field and Data field. Data field should support any type (object, string, number, etc). Check how TypeScript enforcing can be implemented best.
- [ ] API endpoint name: sidecar
- [ ] Delete sidecar object when: 1. The Post/Tag is deleted; 2. The key field is deleted;
- [ ] Encryption keys are "sidecar" documents to a given post / tag - the Post / Tag carries the ID of the sidecar document
- [ ] Replicate Post/Tag's memberOf property to sidecarDto
- [ ] Implement key masking in the same was as luminary-media-convert does it. (stored masked?)
- [ ] Encryption keys does not need to be stored encrypted at rest (i.e. not needed to use cryptoDto)
- [ ] Ensure cryptoKey docs are never retrievable via the query API endpoint (should not be possible to bulk-extract keys)
- [ ] Sidecar API: Authenticated (JWT checking, future session token checking).
- [ ] Sidecar API: Supply post / tag ID (parentID of contentDto) to retreive key

MrDirkelz:


johan-bell:
Dirk — I've duplicated part of this, and your design is the better one. Flagging it before anything lands so we don't end up with two key mechanisms on 1878.

Yesterday and today I built `GET /media/key` (`api/src/endpoints/mediaKey.controller.ts`), which landed on 1878 in #1913. It reads the key back out of a `CryptoDto` via `retrieveCryptoData` and gates on `View`-or-`CmsView` against the parent's `memberOf`. ADR-0018 names that mechanism as exactly wrong for a client-delivered key, and I agree with your reasons — particularly the third row of your table: encrypting at rest turns a recoverable problem into an unrecoverable one and buys nothing against the only threat it would defend against, since the same attacker can read `ENCRYPTION_KEY` from the environment.

Concretely, yours has four protections mine lacks: refusing draft/scheduled/expired parents, exclusion from `/query`, `/fts`, sync and socket rooms so there is no bulk-extraction path, rate limiting, and masking. Mine returns the raw key with none of those.

**How I found it:** I went looking for existing work in the issues and in #1900's checklist, neither of which mentions key delivery, and did not check for open PRs from others against 1878 before building. That's my mistake.

**Where we collide** — ten files, the load-bearing ones being:

- `processMediaDto.ts` / `.spec.ts` — you change `processMedia(media, parent, db)` and swap `storeCryptoData` for `upsertHlsKeySidecar`; I changed the same function to normalise the media URL and to no-op when the submitted key equals the stored one (comparing via `retrieveCryptoData`, which your version removes).
- `MediaDto.ts`, `shared/src/types/dto.ts`, `app.module.ts`, `processPostTagDto.ts`, `EditContentVideo.vue`.

**What I'd suggest, if you agree:** your sidecar design stays and mine goes. That means deleting `mediaKey.controller.ts` and its spec, and repointing the two consumers I added — `app/src/components/content/VideoPlayer.vue` and `cms/src/components/content/VideoPreview.vue` — at `GET /sidecar`. Both call a single `getRest().getMediaKey(docId)` in `luminary-shared`, so it is one function to change plus the URL. That also removes your conflict, since the overlap disappears with my endpoint.

Two things of mine worth keeping regardless of whose endpoint wins, both in `processMedia`:

1. The URL normalisation (`toStoredMediaUrl`) from #1912.
2. Re-submitting an unchanged key is not a change. The CMS cannot tell — a saved key is only ever an id there — so it confirms a replacement it cannot rule out; the API can compare and keep the existing record rather than minting a second one. That comparison currently uses `retrieveCryptoData`; against a masked sidecar it becomes a cheaper comparison, since you can mask the submitted key and compare without unmasking anything.

Happy to do the migration and hand you a clean rebase, or to leave it with you — your branch, your design. Which would you prefer?

johan-bell:
Reviewed the branch, ran every touched suite. Findings, most severe first.

**1. Two tests fail as merged.** `Tests: 2 failed, 27 passed` on a clean run. Both are spec defects, not code defects.

- `api/src/endpoints/sidecar.controller.spec.ts:389` — `expect(rateLimiter.recordProbeStrike).not.toHaveBeenCalledWith("test-user")` arrives with the mock already called 8 times. Line 386 clears `recordReadStrike` only, the suite's `beforeEach` resets just `requestUser`, and jest config sets no `clearMocks` — so the eight preceding 403/404 tests leave their strikes behind. The controller is correct: `recordProbeStrike` is only reachable via `probeFail` (`sidecar.controller.ts:73–76`), never on the 200 path.
- `api/src/changeRequests/documentProcessing/processPostTagDto.sidecar.spec.ts:98` — `TypeError: Cannot read properties of undefined (reading 'filter')`. `processChangeRequest` only sets `res.warnings` when the array is non-empty (`processChangeRequest.ts:146–148`), and a clean delete produces none.

**2. Local dev scaffolding is committed.** `.worktree-ports.env`, `dev-api.sh`, `dev-app.sh`, `dev-cms.sh` at the repo root, containing `cd "/Users/dirk/orca/workspaces/luminary/1901-.../api"`. Anyone else running `./dev-api.sh` gets `cd: no such file or directory`.

**3. Sidecar-ID squatting blocks a victim's key permanently.** IDs are deterministic (`sidecar.service.ts:16–18`) and `_id`s are client-chosen. `validateChangeRequest.ts:69` rejects a doc whose *type* is `Sidecar`, but nothing rejects a **Post** whose `_id` begins `sidecar-`. So an editor with create rights in any group can pre-create a Post at `sidecar-<victimParentId>-hlsEncryptionKey`; the victim's later key write then hits `"Document type change not allowed"` (`db.service.ts:452–457`), `processMedia` rethrows, and the change request fails permanently with an error pointing nowhere near the cause. Deleting the victim parent also hard-deletes the squatting doc cross-group with no `DeleteCmd`, leaving stale copies on clients that synced it. No key leaks either way — a squatting doc fails `isHlsEncryptionKeyData` → 409. Cheapest fix: reject client-submitted `_id`s with the `sidecar-` prefix, or check `doc.type === DocType.Sidecar` in `getSidecar`/`upsertSidecar`.

**4. The 403/404 split leaks parent existence.** `sidecar.controller.ts:97–105`: absent parent → 404, existing parent without View → 403. An unentitled caller can distinguish real parent IDs from fabricated ones. The ADR's 404-for-both promise covers no-parent vs no-sidecar and is honoured, so this is narrower than the ADR implies but still a probe. Bounded by the limiter and by IDs being UUIDs; 404 on the permission failure closes it.

**Deployment note, not a code defect:** with `TRUST_PROXY` unset behind a reverse proxy, every anonymous viewer shares one `anon:<proxy-ip>` bucket of ~30 reads/min.

**Checked and sound** — read in full, not skimmed: access control (View against the parent's own freshly-fetched `memberOf`, `verifyAccess` argument order, parent type restricted to Post/Tag, draft/scheduled/expired refused via `isParentAvailable`, a forged `hlsKey_id` inert since the ID derives from `parentId`); every bulk-extraction exclusion the ADR claims — `/query` blocking both `type` and `docType` post-expansion, FTS failing closed, socket broadcast early-returning before room fan-out, `aclValidation` unable to grant sidecar permissions, no `DeleteCmd` emission; rate limiter keying, strike placement and bounded memory; the mask round-tripping with a shared test vector, and nothing in the code treating it as a secret; v20's idempotence and rerun-safety after partial failure; and `processMediaDto`'s unchanged-key no-op plus `finally { delete media.hlsKey }` keeping the raw key off the replicated parent on every path.

api/src/endpoints/sidecar.controller.spec.ts
```diff
@@ -0,0 +1,431 @@
...
+            const res = await get({ parentId: "post-sc-live", sidecarType: SidecarType.HlsEncryptionKey });
+            expect(res.status).toBe(429);
+            expect(res.headers["retry-after"]).toBe("1");
+        });
+
+        it("records a read strike on a successful fetch, keyed by the caller's identity", async () => {
+            rateLimiter.recordReadStrike.mockClear();
+            await get({ parentId: "post-sc-live", sidecarType: SidecarType.HlsEncryptionKey });
+            expect(rateLimiter.recordReadStrike).toHaveBeenCalledWith("test-user");
+            expect(rateLimiter.recordProbeStrike).not.toHaveBeenCalledWith("test-user");
```
johan-bell, line 422: Fails on a clean run: the mock arrives with 8 calls from the earlier 403/404 tests. Line 386 clears `recordReadStrike` only, and jest config sets no `clearMocks`. Clear `recordProbeStrike` too (or add `clearMocks` to the config). The controller is right — `recordProbeStrike` is only reachable via `probeFail`.

api/src/changeRequests/documentProcessing/processPostTagDto.sidecar.spec.ts
```diff
@@ -0,0 +1,244 @@
...
+            ["group-super-admins"],
+            db,
+        );
+
+        const del = postCr(id);
+        del.doc.deleteReq = 1;
+        const result = await processChangeRequest("test-user", del, ["group-super-admins"], db);
+
+        expect(result.result.ok).toBe(true);
+        expect(result.warnings.filter((w) => w.includes("sidecar"))).toEqual([]);
```
johan-bell: Throws `Cannot read properties of undefined (reading 'filter')`. `processChangeRequest` only sets `res.warnings` when the array is non-empty (`processChangeRequest.ts:146–148`), and a clean delete produces none. `(result.warnings ?? [])` fixes it.

api/src/changeRequests/validateChangeRequest.ts
```diff
@@ -63,6 +63,16 @@ export async function validateChangeRequest(
         };
     }
 
+    // Sidecars are server-side only; deny explicitly. DocType.Sidecar is a valid
+    // enum member so the type check above won't reject it, and class-validator's
+    // forbidUnknownValues could flip silently on a dependency bump.
+    if (changeRequest.doc.type === DocType.Sidecar) {
```
johan-bell, line 71: This rejects a doc whose *type* is `Sidecar`, but not a **Post** whose `_id` starts `sidecar-`. Since IDs are deterministic, an editor with create rights in any group can squat `sidecar-<victimParentId>-hlsEncryptionKey`; the victim's key write then fails with "Document type change not allowed" permanently, pointing nowhere near the cause. No key leaks. Rejecting a client-submitted `sidecar-` prefix here would close it.

api/src/endpoints/sidecar.controller.ts
```diff
@@ -0,0 +1,136 @@
...
+        }
+
+        const hasPermission = PermissionSystem.verifyAccess(
+            parent.memberOf,
+            parent.type,
+            AclPermission.View,
+            userDetails.groups,
+        );
+        if (!hasPermission) {
+            probeFail(HttpStatus.FORBIDDEN, "Forbidden");
```
johan-bell: 403 here vs 404 above lets an unentitled caller tell real parent IDs from fabricated ones. The ADR's 404-for-both covers no-parent vs no-sidecar and is honoured, so this is narrower than it reads — but 404 here too would close the probe entirely, at no cost to a legitimate caller.

dev-api.sh
johan-bell, line 1: Machine-specific: `cd "/Users/dirk/orca/workspaces/luminary/1901-.../api"`. Anyone else running this gets `cd: no such file or directory`. Same for `dev-app.sh`, `dev-cms.sh` and `.worktree-ports.env` — worth dropping from the PR and gitignoring.

johan-bell (commented):
Design is sound — I read the access control, every bulk-extraction exclusion, the limiter, the mask and the v20 upgrade in full, and they do what ADR-0018 says. Five inline notes: two specs that fail on a clean run, committed dev scripts, and two edge cases (sidecar-ID squatting, the 403/404 split). None touch the core.

Also relevant: my `GET /media/key` on 1878 duplicates this and should go in favour of it — details in my earlier comment.

MrDirkelz:
Fixing:
- @johan-bell: review summary — Design is sound — I read the access control, every bulk-extraction excl…
- @johan-bell: comment — Reviewed the branch, ran every touched suite. Findings, most severe fir…

Will be in the next commit.

johan-bell:
Your three failing checks: one is yours, two are mine.

**Yours — `Check ADR numbers`.** Two ADRs are numbered 0018: your `0018-hls-encryption-keys-as-non-replicated-sidecars.md` and `0018-web-ssg-and-isr-tier.md`, which landed on the base branch meanwhile. Renumbering yours to 0019 clears it.

**Mine — the App and CMS test jobs.** Both are the submodule integration, not anything in this PR:

- The app workflow asks for the submodule and cannot have it. `luminary-media-convert` is private and a workflow's `GITHUB_TOKEN` is scoped to this repository, so the clone answers `Repository not found` and checkout aborts in six seconds.
- The CMS workflow never asked for it — I added the dependency in #1914 without touching the workflow — so `@luminary-media-converter/player-web-legacy` fails to resolve and ten suites die on import, including the `EditContent` ones that have nothing to do with media.

Fix in #1918. It needs a `MEDIA_CONVERT_TOKEN` secret that does not exist yet, so it will not turn your checks green on its own — but nothing in your branch is causing them, and there is no point rebasing to chase them.

Also: everything in my review is addressed on `56937dce`, and the two specs are fixed — I have closed my #1917 as superseded. Verified your head at 51/51 on the three affected suites and 906 passed across the api suite, with only the two MinIO-dependent suites failing as they do on the base branch.

johan-bell:
The workflow fix is merged into the base (#1918), so it reaches this PR without you doing anything — CI runs against a merge of your head into the base, so your next run picks up the corrected workflows.

It will not turn your checks green yet, though. They now need `MEDIA_CONVERT_TOKEN` on the repository — a credential with read access to the private encoder repo — and that secret does not exist. Johan has it to create; nothing on your side is blocking it.

What is still yours here is the ADR number: `0018` is taken by `0018-web-ssg-and-isr-tier.md` on the base branch, so `Check ADR numbers` fails until yours becomes `0019`.

johan-bell:
Your two red checks are not yours, and they are already fixed on the base — they just cannot see it.

A rerun replays the original merge commit, so it keeps the base as it was at 11:52 and misses everything since. What landed since:

- `luminary-media-convert` is **public** now, so the submodule clones with the built-in token — no secret needed, and #1918's `MEDIA_CONVERT_TOKEN` is gone from the workflows.
- The workflows now build the submodule's libraries (`ci:libs` + `build:libs`). They ship only `dist/`, which is gitignored, so a fresh checkout had sources and no entry points — that is the `Failed to resolve entry for package` you are seeing.
- A `player-core` bug behind it: `serve()` threw above `load()`'s try block where there is no `createObjectURL`, so every fire-and-forget `load()` leaked an unhandled rejection under jsdom. 23 of them, failing the app run while all 901 tests passed.

**Merging the base into your branch is all it needs.** I simulated exactly that locally — your head merged with current base, submodule at `23288e8`, both suites run the way CI runs them:

| | Exit code | Result |
| --- | --- | --- |
| app | 0 | 106 files, 901 passed, 1 skipped |
| cms | 0 | 127 files, 1047 passed, 3 skipped |

I did not push that to your branch — you have been working on it all day and an unexpected merge commit is yours to invite. Hit "Update branch" (or merge the base yourself) and both go green.

PR #1917, opened 2026-08-19
test(api): fix two sidecar specs that fail on a clean run

johan-bell:
The two failing specs from my review of #1915, fixed. Into your branch rather than pushed onto it, since you have been pushing to it all morning and I did not want to land on top of your working copy — take it, cherry-pick it, or ignore it and fix them your own way.

Both are spec defects. The code they cover is correct in both cases.

**`sidecar.controller.spec.ts`** — the read-strike test asserts `recordProbeStrike` was *not* called, but nothing clears it. The suite sets no `clearMocks` and its `beforeEach` resets only `requestUser`, so the eight 403/404 cases above have already struck as `"test-user"` and the negative assertion fails on strikes they left behind. The three sibling tests below it clear both mocks; this one cleared only the read mock. `recordProbeStrike` is unreachable on the 200 path — only `probeFail` calls it — so the assertion is right and its setup was not.

**`processPostTagDto.sidecar.spec.ts`** — `result.warnings.filter(...)` throws `Cannot read properties of undefined`. `processChangeRequest.ts:174` assigns `res.warnings` only when the array is non-empty, and a clean delete produces none. Guarded with `?? []`, so the assertion still means "no sidecar warnings" whether the field is absent or present.

## What I could not do

**I have not run these.** Both suites need a live CouchDB on 5984 and Docker is down on my machine — every test in them errors with `ECONNREFUSED` regardless of the fix, so a green run here would have proved nothing anyway. What I did instead:

- `tsc --noEmit` over `api/` — both specs compile.
- Re-read the code each fix depends on: `processChangeRequest.ts:174` for the `warnings` premise, and the three sibling tests in the same describe block that already clear both mocks for the other.

So please run them before merging. If your CI reports more than these two failing, I have not seen that output — I found two on `3144f71d` and confirmed both still live on `ac305b97`.

Unrelated but worth knowing: `s3.service.spec.ts` and `processImageDto.spec.ts` fail identically on the base branch (39 tests, needing a live MinIO), so they are not yours.

### DIFF
--- a/api/src/changeRequests/documentProcessing/processPostTagDto.sidecar.spec.ts
+++ b/api/src/changeRequests/documentProcessing/processPostTagDto.sidecar.spec.ts
@@ -95,7 +95,9 @@ describe("processPostTagDto — sidecar lifecycle", () => {
         const result = await processChangeRequest("test-user", del, ["group-super-admins"], db);
 
         expect(result.result.ok).toBe(true);
-        expect(result.warnings.filter((w) => w.includes("sidecar"))).toEqual([]);
+        // `warnings` is omitted entirely when there are none — processChangeRequest
+        // only assigns it for a non-empty array — and a clean delete produces none.
+        expect((result.warnings ?? []).filter((w) => w.includes("sidecar"))).toEqual([]);
         expect((await db.getDoc(id)).docs).toHaveLength(0);
     });
 
--- a/api/src/endpoints/sidecar.controller.spec.ts
+++ b/api/src/endpoints/sidecar.controller.spec.ts
@@ -415,7 +415,12 @@ describe("SidecarController", () => {
         });
 
         it("records a read strike on a successful fetch, keyed by the caller's identity", async () => {
+            // Both, because the second assertion below is a negative one and the
+            // suite clears nothing between tests: the eight 403/404 cases above
+            // have already struck as "test-user", so an uncleared probe mock
+            // arrives pre-called and the assertion fails on the strike they left.
             rateLimiter.recordReadStrike.mockClear();
+            rateLimiter.recordProbeStrike.mockClear();
             await get({ parentId: "post-sc-live", sidecarType: SidecarType.HlsEncryptionKey });
             expect(rateLimiter.recordReadStrike).toHaveBeenCalledWith("test-user");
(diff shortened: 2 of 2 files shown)


johan-bell:
Now verified — I said on opening that I could not run these; a CouchDB is available again and I have.

Against your `ac305b97`, both suites in isolation:

| | Result |
| --- | --- |
| Without this change | `Tests: 2 failed, 30 passed, 32 total` |
| With it | `Tests: 32 passed, 32 total` |

Full api suite with it applied: **905 passed, 30 failed across 78 suites** — and the only two failing suites are `s3.service.spec.ts` and `processImageDto.spec.ts`, which fail identically on the base branch and need a live MinIO. Nothing else moved.

One correction to my own first run in the PR description: an earlier attempt reported both suites green *before* the fix, which was wrong — I had stashed some unrelated `.env.example` noise while my fix sat in the commit, so "before" was measured with the fix applied. Redone by checking the two spec files out at your head, which is where the 2-failed number above comes from.

johan-bell:
Superseded — you fixed both specs in 56937dce, and the  guard is byte-identical to what was here. Closing.

Verified your head: the three affected suites are **51 passed, 51 total**, and the full api suite is **906 passed** with only `s3.service.spec.ts` and `processImageDto.spec.ts` failing — both pre-existing on the base branch and needing a live MinIO.

All four review findings addressed, and the two I flagged as edge cases you closed properly rather than minimally: the `sidecar-` prefix guard stops the squat at validation with an error that names the cause, and moving the permission denial to 404 removes the existence probe entirely rather than narrowing it.

johan-bell:
Superseded — you fixed both specs in 56937dce, and the `warnings` guard is byte-identical to what was here. Closed.

Verified your head: the three affected suites are **51 passed, 51 total**, and the full api suite is **906 passed** with only `s3.service.spec.ts` and `processImageDto.spec.ts` failing — both pre-existing on the base branch, needing a live MinIO.

All four findings addressed, and the two I flagged as edge cases you closed properly rather than minimally: the `sidecar-` prefix guard stops the squat at validation with an error that names the cause, and moving the permission denial to 404 removes the existence probe outright rather than narrowing it.

PR #1922, opened 2026-08-19
test(cms): wait for the unmasked key instead of a fixed tick

johan-bell:
The one check still red on #1915 after the base update:

```
FAIL src/components/content/VideoPreview.spec.ts > fetches a saved key, which the document cannot show again
AssertionError: expected undefined to be '000102030405060708090a0b0c0d0e0f'
```

`mountAndOpen` waits one `setTimeout(0)`, but the key arrives **two** awaits deep — `getSidecar`, then `unmaskKeyHex`'s `crypto.subtle.digest`. One tick covers the first and not the second, so the assertion reads `storedKey` before the watcher has written it.

Being straight about the evidence: **it passes locally both with and without this change** — I ran it repeatedly. The race only loses on a slower runner, which is why it failed CI and not my machine. So the argument is structural rather than empirical: `waitForExpect` waits for the value instead of for a duration, and the outcome stops depending on host speed. Your next CI run is the actual proof.

The component is correct; only the assertion's timing was. Full CMS suite on this branch with the change: 127 files, 1047 passed, 3 skipped.

Into your branch rather than onto it, as before.


johan-bell:
Second commit, after reading the rest of your branch properly.

You already delete `mediaKey.controller.ts` and its spec, and repoint both consumers at `getSidecar` — so the migration I offered on #1915 is done, and better than my sketch: keying the lookup off `parentId` is right, since the sidecar id derives from the parent, and my endpoint only reached it by walking `parentMedia`.

What that leaves behind is mine: `VideoPreview.live.spec.ts` drove `GET /media/key` against a running API, with `CryptoDto` fixtures that no longer have a producer. It stays green only because it skips unless `LUMINARY_LIVE_API` is set — so nothing would have caught it, and the next person to set that variable loses an afternoon to a 404 on a deleted endpoint. Removed.

The wiring it covered — component through the shared client to a real API — is worth having again against `/sidecar`, but that is your branch's shape to write rather than a rewrite of mine to guess at.

Verified on the merged result: cms 126 files / 1047 passed (no skips now), app 106 files / 901 passed.

PR #1930, opened 2026-08-21
1929 cms api change the view auto assign rule to cmsview

### LINKED ISSUE #1929: CMS, API: change the view auto-assign rule to cmsView
When a group's ACL entry for a doc type gains its first permission, the CMS auto-adds \`AclPermission.View\` alongside it (\`cms/src/components/groups/permissions.ts\`, \`validateAclEntry\`) — and \`EditAclByGroup.vue\` does the same on manual entry creation. Conversely, clearing \`View\` clears all other permissions on that entry.

App users only ever consume \`View\`; every other permission (\`Edit\`, \`Translate\`, \`Publish\`, \`Assign\`, etc.) is CMS-only. Auto-assigning \`View\` when any other permission is set therefore grants app-facing visibility as a side effect of a CMS-only permission change, when \`AclPermission.CmsView\` is what's actually implied.

**Scope:**
- Update the auto-assign/auto-clear rule in \`cms/src/components/groups/permissions.ts\` (\`validateAclEntry\`) and \`cms/src/components/groups/EditAclByGroup.vue\` to use \`AclPermission.CmsView\` instead of \`AclPermission.View\`.
- Add an API schema upgrade script (\`api/src/db/schemaUpgrade\`, next version after \`v19\`) that walks existing \`Group\` docs and applies the same rule retroactively: for every ACL entry that has any permission set but lacks \`cmsView\`, add \`cmsView\`. Wire it into \`db.upgrade.ts\`.
- Check \`shared/src/permissions/permissions.ts\` and its spec for any assumption that \`View\` is present whenever another permission is (per the existing GitHub #160 comments there — View and CmsView already resolve independently, so this should be additive, but worth confirming nothing quietly relied on the old CMS-side behavior).

Confirm with the reporting team member whether \`View\` should still be auto-added the very first time *any* CMS permission is granted (so a brand-new group gets both flipped together), or whether \`View\` should now be a fully independent, manually-set toggle.

MrDirkelz:


cms/src/components/groups/EditAclByGroup.vue
```diff
@@ -80,7 +80,8 @@ const toggleAclEntry = (aclEntry: any) => {
     if (aclEntry.permission.length > 0) {
         aclEntry.permission = [];
     } else {
-        aclEntry.permission.push(AclPermission.View);
+        // Matches the auto-assign rule in validateAclEntry
+        aclEntry.permission.push(AclPermission.CmsView);
```
ivanslabbert: Duplicate logic? `permissions.ts` is supposed to already set the CmsView permission

cms/src/components/groups/permissions.ts
```diff
@@ -109,18 +109,19 @@ export const hasChangedPermission = computed(() => {
...
-        aclEntry.permission.push(AclPermission.View);
+        aclEntry.permission.push(AclPermission.CmsView);
     }
 
-    // Remove all other permissions if the view permission is removed
-    if (!aclEntry.permission.includes(AclPermission.View)) {
-        aclEntry.permission = [];
+    // CMS-only permissions can't be exercised without CmsView; app-facing View stands on its own
+    if (!aclEntry.permission.includes(AclPermission.CmsView)) {
+        aclEntry.permission = aclEntry.permission.filter((p) => p === AclPermission.View);
```
ivanslabbert: Remove all other permissions if CmsView and View permissions are both removed. Also implement in API

cms/src/components/groups/permissions.ts
```diff
@@ -109,18 +109,19 @@ export const hasChangedPermission = computed(() => {
...
+    // assigned deliberately, so a CMS-only permission change can't grant app visibility by accident.
     if (
         aclEntry.permission.length &&
-        !aclEntry.permission.includes(AclPermission.View) &&
+        !aclEntry.permission.includes(AclPermission.CmsView) &&
         prevAclEntry.permission.length === 0
     ) {
-        aclEntry.permission.push(AclPermission.View);
+        aclEntry.permission.push(AclPermission.CmsView);
     }
```
ivanslabbert: Automatically set view if any other permission is set.

api/src/endpoints/query.service.ts
```diff
@@ -131,10 +131,10 @@ export class QueryService {
...
 
         // CMS-scoped requests (cms:true) are permission-gated by CmsView; app/public requests by
-        // View. A CmsView ACL entry always also carries View (see validateAcl), so CMS users keep
-        // their app-side published access too. When the caller holds no CmsView on any requested
-        // group the resulting empty groups fall through to the same 403 guards as a missing View —
-        // so requesting cms:true without CmsView is Forbidden, not a silent published-only result.
+        // View. The two are granted independently, so CMS access does not imply app-side published
+        // access. When the caller holds no CmsView on any requested group the resulting empty groups
+        // fall through to the same 403 guards as a missing View — so requesting cms:true without
+        // CmsView is Forbidden, not a silent published-only result.
```
ivanslabbert: Simplify comment to not include info on logic not implemented here

api/src/changeRequests/aclValidation.ts
ivanslabbert: See comment in CMS permissions.ts

ivanslabbert (approved):

PR #1934, opened 2026-08-21
1860 app inappbrowsercheck not working with new telegram

### LINKED ISSUE #1860: APP: InAppBrowserCheck not working with new telegram

MrDirkelz:


ivanslabbert (commented):
There is a npm package / github repo that detects in-app browsers, with Telegram support. Check if it has been updated to detect Telegram correctly (after Telegram removed the API that was previously used by the library to detect the Telegram in-app browser).

johan-bell (changes requested):
The move to inapp-spy introduces a regression for installed users:

- `inAppBrowser.ts:41` — inapp-spy's generic pattern `(iPhone|iPod|iPad)(?!.*Safari/)` matches iOS home-screen PWA user agents (they omit `Safari/`), so installed iOS users would get the "open in your browser" interstitial on every cold launch. Please gate on the existing `isInstalledStandalone()` from `app/src/globalConfig.ts`.
- `inAppBrowser.ts:15` — the comment saying inapp-spy has no Telegram detection isn't accurate: its `getDetectClientSide` already checks `TelegramWebviewProxy`; only `__tg__webview_set` is genuinely additional.
- `inAppBrowser.ts:27` — the hand-rolled bridge check covers fewer properties (`TelegramWebview`, `TelegramWebviewProxyProto` missing) than the library it supplements.
- Test gap: an iOS standalone user-agent case in `playwright-tests/app/flows/in-app-browser.spec.ts` would have caught the PWA regression.

ivanslabbert (approved):

johan-bell (approved):

PR #1937, opened 2026-08-24
fix(app): close the per-route JSDOM in the SSG prerender (#1936)

### LINKED ISSUE #1936: APP: vite-ssg leaks a JSDOM per prerendered route, OOMing the SSG build
## Summary

`vite-ssg` constructs a `new JSDOM(renderedHTML)` for **every** prerendered route and never
closes it. Each unclosed instance retains ~13 MB, so a full-site prerender (~2000 routes)
accumulates far more than any configurable heap, and `build:web` dies with
`FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory`.

This currently blocks the SSG/ISR initial build from ever completing.

## Evidence

In `node_modules/vite-ssg/dist/node.mjs` (v0.23.8, the render loop):

```js
const jsdom = new JSDOM(renderedHTML);
renderPreloadLinks(jsdom.window.document, ctx.modules || new Set(), ssrManifest);
if (head)
  await renderDOMHead(head, { document: jsdom.window.document });
const html = jsdom.serialize();
// ← no jsdom.window.close()
```

Grepping the whole file for `.close()` returns nothing. Each `new JSDOM()` builds a full DOM
tree from a 135–155 KiB page *and* allocates a Node `vm.Context` with its own globals and
prototypes; without an explicit close those are never reclaimed.

Measured in isolation — 300 instances of a representative ~200 KiB page:

| | `heapUsed` after 300 instances |
|---|---|
| Without `close()` | **3900 MB** (~13 MB retained each) |
| With `close()` | **160 MB** (~0.5 MB each) |

A **24× reduction** from one line.

The GC log from the failing build corroborates retention rather than GC pressure — scavenges
reclaim only a few MB against a ~2020 MB live set:

```
Scavenge 2026.0 (2078.5) -> 2019.0 (2078.4) MB
Scavenge 2027.4 (2079.6) -> 2024.0 (2097.4) MB
FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory
```

## Not fixed upstream

Confirmed the identical code (no `.close()`) in the **latest** published `vite-ssg`, 28.3.0 —
so there is no version upgrade to take instead. Worth raising upstream separately; that
shouldn't gate unblocking our build.

## Proposed fix

Add `jsdom.window.close()` immediately after `jsdom.serialize()`, via `patch-package`
(the repo has no dependency-patching tooling yet, so this introduces it).

Closing after `serialize()` is safe: `renderDOMHead` and `renderPreloadLinks` both run before
it, and `onPageRendered` afterwards only manipulates the already-serialised HTML string —
nothing touches the DOM once it's closed.

## Verification

`npm run build:web` completes a full-site prerender without OOM, with heap staying flat across
routes instead of climbing.

MrDirkelz:
Part of #1936 (removes the JSDOM contributor; the unhead and cache-seed contributors remain — see the update below)

## The bug

`vite-ssg` constructs a `new JSDOM(renderedHTML)` for **every** prerendered route and never
closes it (`node_modules/vite-ssg/dist/node.mjs`, the render loop):

```js
const jsdom = new JSDOM(renderedHTML);
renderPreloadLinks(jsdom.window.document, ctx.modules || new Set(), ssrManifest);
if (head)
  await renderDOMHead(head, { document: jsdom.window.document });
const html = jsdom.serialize();
// ← never closed
```

Grepping the whole file for `.close()` returns nothing. Each instance holds a full DOM tree
built from a 135–155 KiB page, plus its own Node `vm.Context` with its own globals and
prototypes — none of it reclaimed without an explicit close.

Over a full-site prerender this exhausts any heap the build can be given, and `build:web` dies:

```
FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory
```

## Measurement

300 instances of a representative ~200 KiB page, with and without the close:

| | `heapUsed` after 300 instances |
|---|---|
| Without `close()` | **3900 MB** (~13 MB retained each) |
| With `close()` | **160 MB** (~0.5 MB each) |

**24× reduction** from one line.

The GC log from the failing build confirms retention rather than GC pressure — scavenges
reclaim only single-digit MB against a ~2020 MB live set:

```
Scavenge 2026.0 (2078.5) -> 2019.0 (2078.4) MB
Scavenge 2027.4 (2079.6) -> 2024.0 (2097.4) MB
```

> **Update (measured):** this removes the JSDOM contributor *completely* — a closed per-route JSDOM
> retains 2.75 MB/page vs 2.74 MB/page with JSDOM bypassed entirely, i.e. nothing. But it is **not**
> the whole leak: `@unhead/dom` accounts for ~3.5 MB/page and the per-page response-cache seed for
> ~2.75 MB/page. Land this on its merits; it does not on its own make a full-site build complete.
> See the comment below for the bisect table and method.

## The fix

`jsdom.window.close()` immediately after `jsdom.serialize()`, applied via `patch-package`.

Safe at that point: `renderPreloadLinks` and `renderDOMHead` both run *before* the serialize,
and `onPageRendered` afterwards only manipulates the already-serialised HTML string — nothing
touches the DOM once it's closed.

## Why a patch rather than an upgrade

The identical unclosed-JSDOM code is present in the **latest** published `vite-ssg` (28.3.0),
verified directly — there's no version to upgrade to that fixes this. Worth raising upstream
separately, but that shouldn't gate unblocking our build. `patches/README.md` records when the
patch can be dropped.

This introduces `patch-package`; the repo had no dependency-patching tooling before.

## Verification

- Patch applies to a pristine dependency: removed `node_modules/vite-ssg`, reinstalled clean
  (confirmed `jsdom.window.close` absent), ran `npm run postinstall` → applied `✔`, and the
  resulting file passes `node --check`.
- Lockfile changes are limited to `patch-package` and its own dependency tree.

**Not yet verified:** a full `build:web` run to completion against a production-sized dataset.
That needs the deployment environment and is the real end-to-end confirmation — worth doing
before this is relied on.

### DIFF
--- a/app/patches/vite-ssg+0.23.8.patch
+++ b/app/patches/vite-ssg+0.23.8.patch
@@ -1,0 +1,12 @@
+diff --git a/node_modules/vite-ssg/dist/node.mjs b/node_modules/vite-ssg/dist/node.mjs
+index 6faa9cd..a8fc2d2 100644
+--- a/node_modules/vite-ssg/dist/node.mjs
++++ b/node_modules/vite-ssg/dist/node.mjs
+@@ -1043,6 +1043,7 @@ async function build(ssgOptions = {}, viteConfig = {}) {
+         if (head)
+           await renderDOMHead(head, { document: jsdom.window.document });
+         const html = jsdom.serialize();
++        jsdom.window.close();
+         let transformed = await onPageRendered?.(route, html, appCtx) || html;
+         if (critters)
+           transformed = await critters.process(transformed);
--- a/app/package-lock.json
+++ b/app/package-lock.json
@@ -55,6 +55,7 @@
                 "fake-indexeddb": "^6.2.5",
                 "jsdom": "^23.2.0",
                 "npm-run-all2": "^6.2.6",
+                "patch-package": "^8.0.1",
                 "postcss": "^8.5.4",
                 "prettier": "^3.0.3",
                 "prettier-plugin-tailwindcss": "^0.5.11",
@@ -4195,6 +4196,13 @@
                 "node": ">=10.0.0"
             }
         },
+        "node_modules/@yarnpkg/lockfile": {
+            "version": "1.1.0",
+            "resolved": "https://registry.npmjs.org/@yarnpkg/lockfile/-/lockfile-1.1.0.tgz",
+            "integrity": "sha512-GpSwvyXOcOOlV70vbnzjj4fW5xW/FdUF6nQEt1ENy7m4ZCczi1+/buVUPAqmGfqznsORNFzUMjctTIp8a9tuCQ==",
+            "dev": true,
+            "license": "BSD-2-Clause"
+        },
         "node_modules/abbrev": {
             "version": "2.0.0",
             "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-2.0.0.tgz",
@@ -4856,6 +4864,22 @@
             },
             "engines": {
                 "node": ">= 6"
+            }
+        },
+        "node_modules/ci-info": {
+            "version": "3.9.0",
+            "resolved": "https://registry.npmjs.org/ci-info/-/ci-info-3.9.0.tgz",
+            "integrity": "sha512-NIxF55hv4nSqQswkAeiOi1r83xy8JldOFDTWiug55KBu9Jnblncd2U6ViHmYgHf01TPZS77NJBhBMKdWj9HQMQ==",
+            "dev": true,
+            "funding": [
+                {
+                    "type": "github",
+                    "url": "https://github.com/sponsors/sibiraj-s"
+                }
+            ],
+            "license": "MIT",
+            "engines": {
+                "node": ">=8"
             }
         },
         "node_modules/clean-css": {
@@ -6215,6 +6239,16 @@
             },
             "funding": {
                 "url": "https://github.com/sponsors/sindresorhus"
+            }
+        },
+        "node_modules/find-yarn-workspace-root": {
+            "version": "2.0.0",
+            "resolved": "https://registry.npmjs.org/find-yarn-workspace-root/-/find-yarn-workspace-root-2.0.0.tgz",
+            "integrity": "sha512-1IMnbjt4KzsQfnhnzNd8wUEgXZ44IzZaZmnLYx7D5FZlaHt2gW20Cri8Q+E/t5tIj4+epTBub+2Zxu/vNILzqQ==",
+            "dev": true,
+            "license": "Apache-2.0",
+            "dependencies": {
+                "micromatch": "^4.0.2"
             }
         },
         "node_modules/flat-cache": {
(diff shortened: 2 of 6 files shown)


MrDirkelz:
## Update: measured — this fixes the JSDOM component completely, but it is not the whole leak

Since opening this I built a local reproduction (mock `/query` API + synthetic corpus, no server needed) and instrumented the render loop to sample `heapUsed` after two forced GC passes. That turns the earlier isolated micro-benchmark into a measurement of the real loop.

### The JSDOM fix does exactly what this PR claims

| variant | MB retained per page |
|---|---|
| JSDOM bypassed entirely | 2.74 |
| JSDOM created **and closed** (this PR) | 2.75 |

Indistinguishable — with `window.close()`, a per-route JSDOM contributes **no** measurable retention. The patch is correct and worth landing on its own.

### But two other contributors remain

Bisecting the render loop by disabling one call at a time (growth measured pages 10 → 40):

| variant | MB/page |
|---|---|
| baseline | 6.24 |
| skip `renderPreloadLinks` | 6.21 |
| skip `renderDOMHead` (@unhead/dom) | **2.75** |
| skip both | 2.75 |

- **`@unhead/dom`'s `renderDOMHead` — ~3.5 MB/page (56%).** `@unhead/dom` stores live DOM element references (`state.elMap[id] = $el`), so a retained head instance keeps its JSDOM element graph alive. This is consistent with the heap-snapshot diff, which showed `SymbolTreeNode` (+2101/page), `AttrImpl` (+1556/page) and `HTMLLinkElementImpl` (+506/page) accumulating.
- **Per-page response-cache seed — ~2.75 MB/page (44%).** A second snapshot diff with unhead disabled shows ~500 content-doc copies retained per render (doc body strings, author strings, `(object properties)` arrays).
- `renderPreloadLinks` contributes nothing.

### Caveat on the absolute numbers

These come from synthetic data whose feed queries match the whole corpus, producing ~970 KB pages against ~150 KB in production. The **mechanisms** hold; the magnitudes are inflated. Production measurement puts the real figure nearer ~1.2 MB/page — which still exhausts a 2 GB heap well before the full route set.

### What this means for reviewers

Land this on its merits — it removes one of three contributors, and on the constrained host it took the build from failing at ~92s to ~192s with full mark-compacts being reached. **It does not on its own make the full-site build complete**, and the issue should stay open for the unhead and cache-seed components.

The deployment side is being addressed independently by batching the render across processes, which bounds accumulation regardless of which library is responsible.

MrDirkelz:
## Added to this branch: `SSG_FORCE_REDIRECTS`

Second commit (`2de74cfa`) adds a small, separate capability needed by the deployment-side fix.

**Why.** Redirect artifacts are gated `if (!IS_SCOPED)`, so a scoped rebuild never writes
`ssg-redirect-index.json` — correct today, because a partial index would overwrite the complete
one, and some run is always unscoped.

That assumption breaks under the batched build being added deployment-side: to keep the per-page
leak bounded, the site is rendered in several **scoped** passes and there is no full build at all.
Nothing would then generate redirects, and `ssg-redirect-index.json` is in the deployer's required
artifacts. Redirects aren't optional on this site and are expected to grow.

**What.** `SSG_FORCE_REDIRECTS=1` writes redirect artifacts from a scoped build. The deployment
driver sets it on one enumeration pass only. Safe regardless of scope — redirect enumeration reads
its own docs and does not depend on which routes rendered.

**Verified** against a local mock API serving two redirect docs:

| | `ssg-redirect-index.json` |
|---|---|
| scoped, no flag | **missing** |
| scoped, `SSG_FORCE_REDIRECTS=1` | `{"redirect-1":{"slug":"old-path","status":301},"redirect-2":{"slug":"legacy-page","status":302}}` |

Static redirect files (`old-path.html`, `legacy-page.html`) are emitted too. `npm run type-check`
passes. Default behaviour is unchanged when the flag is unset.

MrDirkelz:
## Correction to my earlier comment on the redirect commit

My previous comment justified the redirect change with "a partial index would overwrite the
complete one." **That is wrong** — `writeRedirectFiles` calls `fetchRedirects(apiUrl)`, a full
enumeration independent of `SSG_ONLY_ROUTES`, so its output is never partial. Apologies; the
commit has been amended (`ec47bff5`) and the flag renamed.

**The actual reason scoped builds skip redirect artifacts:** they are re-read in full and
rewritten wholesale, so exactly one pass should own them. Between builds the ISR watcher maintains
`ssg-redirect-index.json` itself — it loads, mutates it from polled redirect docs, and saves it.
A scoped rebuild re-reading the API could therefore revert a change the watcher has already
applied locally but the API has not caught up with. The gate protects that.

**Why a flag is still needed:** the gate assumes some pass in a run is unscoped. A driver that
renders the site in several scoped passes (to bound the per-page memory retention this PR's other
commit only partly addresses) never runs one, so nothing would emit redirects at all.

`SSG_FORCE_REDIRECTS` → **`SSG_EMIT_REDIRECTS`**, because it is not an override of a safety
measure: it designates which pass owns redirect generation when no unscoped pass exists. Watcher
behaviour and default behaviour are both unchanged.

app/patches/vite-ssg+0.23.8.patch
ivanslabbert: When well tested, submit a PR on the vite-ssg project if possible

ivanslabbert (approved):
Add a follow-up ticket to remove the patch once fixed in the upstream vite-ssg

PR #1938, opened 2026-08-24
APP: player stalls after ~5s when no audio track matches the app language

johan-bell:
**Bug:** on a video with multiple audio tracks where none matches the app language (e.g. English), playback stops after ~5 seconds.

**Cause:** `VideoPlayer.vue`'s `setAudioTrackLanguage` looped over the tracks and set `track.enabled = false` for every non-matching one. When *no* track matched, that disabled **all** audio tracks — the player then ran out its buffer (~5s) with no audio track selected and stalled.

**Fix:** find the matching track first; if there's no match, **keep the currently-enabled (default) track** and return, instead of disabling everything. Only when a track matches do we switch to it.

- Extracted the selection into a pure, tested helper `selectAudioTrackIndex(trackLanguages, languageCode)` in `audioTrackLanguage.ts` — returns the match index, or `-1` meaning "keep the current track".
- Added regression tests (match → index; no match / empty → `-1`).

Verified: type-check clean, lint clean, 21 tests pass across `audioTrackLanguage.spec.ts` + `VideoPlayer.spec.ts`. Not device-verified — worth a quick check on a real multi-audio video with a non-matching app language.

### DIFF
--- a/app/src/components/content/VideoPlayer.vue
+++ b/app/src/components/content/VideoPlayer.vue
@@ -5,7 +5,7 @@ import "videojs-mobile-ui";
 import type Player from "video.js/dist/types/player";
 import { type ContentDto } from "luminary-shared";
 import px from "./px.png";
-import { matchTrackLanguage } from "./audioTrackLanguage";
+import { selectAudioTrackIndex } from "./audioTrackLanguage";
 import LImage from "../images/LImage.vue";
 import { appLanguagesPreferredAsRef, queryParams } from "@/globalConfig";
 import { getMediaProgress, removeMediaProgress, setMediaProgress } from "@/contentProgress";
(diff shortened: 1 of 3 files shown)


app/src/components/content/VideoPlayer.vue
```diff
@@ -91,20 +91,20 @@ function setAudioTrackLanguage(languageCode: string | null) {
...
+        // the buffer drains.
+        console.warn(`No matching audio track found for language: ${languageCode}`);
+        return;
     }
 
-    if (!trackFound) {
-        console.warn(`No matching audio track found for language: ${languageCode}`);
+    for (let i = 0; i < audioTracks.length; i++) {
+        audioTracks[i].enabled = i === matchIndex;
     }
```
MrDirkelz:
```suggestion
    audioTracks.forEach((track, i) => track.enabled = i === matchIndex);
```

app/src/components/content/VideoPlayer.vue
```diff
@@ -91,20 +91,20 @@ function setAudioTrackLanguage(languageCode: string | null) {
         return;
     }
 
-    let trackFound = false;
-    for (let i = 0; i < audioTracks.length; i++) {
-        const track = audioTracks[i];
+    const trackLanguages: (string | null)[] = [];
+    for (let i = 0; i < audioTracks.length; i++) trackLanguages.push(audioTracks[i].language);
```
MrDirkelz: const trackLanguages = audioTracks.map(track => track.language);

MrDirkelz (commented):
Just 2 suggestions

ivanslabbert (approved):

PR #1939, opened 2026-08-24
Disable saving of reading position for posts <= 1min reading time

MrDirkelz:


ivanslabbert (approved):
If the functionality is practically testable, add tests

PR #1945, opened 2026-08-26
Fix seed data that the document pipeline would otherwise derive

johan-bell:
## Problem

A freshly seeded database is visibly broken in the app:

1. **No blog posts on the home page.** The "Newest" collection renders only tags.
2. **Article bodies render as raw JSON.** Opening any post shows `{"type":"doc","content":[{"type":"paragraph"...` instead of prose.

## Common cause

Seeding writes raw JSON straight to CouchDB and deliberately bypasses `process*Dto`, so anything the document pipeline normally derives has to already be present in the seed files. Two such fields were missing or stale.

It compounds with schema versioning: `initSchemaVersion` stamps a fresh database at the current baseline, so back-migrations that would repair the data never run.

---

## Fix 1 — `parentPublishDateVisible`

Denormalized onto content docs by `processContentDto.ts` from the parent post's `publishDateVisible`. Four content docs were missing it, and the seed data disagreed with itself across translations of the same post:

| content doc | parent's `publishDateVisible` | seed had |
|---|---|---|
| `content-blog1-eng` / `-fra` | `true` | missing |
| `content-blog2-eng` | `true` | missing |
| `content-blog2-fra` | `true` | `true` |
| `content-page2-eng` | `true` | missing |
| `content-page2-fra` | `true` | `true` |

`HomePageNewest.vue` filters on `{ parentPublishDateVisible: true }`, so those docs were dropped. Because the French translations carried the field, switching the app to French made Blog 2 appear — which is what pointed at the seed data rather than the query.

Each value is now read from its parent post.

## Fix 2 — content text stored as HTML

`SingleContent.vue` does `const text = computed(() => content.value?.text ?? "")` and feeds it to `v-html`, so `text` must already be HTML.

The TipTap-JSON to HTML conversion exists in exactly one place — schema upgrade **v12** — which runs only at schema version 11. A fresh database never reaches that path:

```
Initialized database schema version to 17 (fresh database)
Skipping schema upgrade v12: current version is 17, expected 11
```

So all 10 content docs with body text rendered their raw TipTap JSON. The 8 empty tag/category docs were unaffected, which is why topic pages looked fine and only articles were broken.

Converted using v12's own `isTipTapJson` guard and `generateHTML(parsed, [StarterKit, Link])`, so the seed output matches what an upgraded database holds.

One deviation, deliberate: the server-side serializer emits `xmlns="http://www.w3.org/1999/xhtml"` on every block element. That is redundant in HTML and is not what authored content looks like — the CMS `RichTextEditor` uses `content-format="html"` — so it is stripped.

## Verification

- Re-seeded a local CouchDB and loaded the app as an unauthenticated guest. Before: "Newest" showed only a Topic A tile, and `/blog1-eng` rendered raw JSON. After: Blog 1 appears in "Newest" and the article renders as prose.
- Checked every converted doc by stripping tags from the HTML and comparing against the text nodes of the original TipTap JSON: **10/10 preserve their text exactly, 0 mismatches.**
- No code changes in this PR — seed data only.

## Note for reviewers — out of scope here

Checking every seed content doc against its parent turned up 8 further mismatches in the opposite direction: the four tag docs (`tag-topicA`, `tag-topicB`, `tag-category1`, `tag-category2`) have `publishDateVisible: false`, but all eight of their content docs carry `parentPublishDateVisible: true`.

By the same denormalization rule those should be `false` — but flipping them would remove the topic tiles from "Newest", changing what the seeded demo displays. That is a product decision rather than a mechanical fix, so it is deliberately not part of this PR.


MrDirkelz (approved):
Nice

PR #1947, opened 2026-08-26
Shared: scope the RestApi localChanges live query

### LINKED ISSUE #1946: Shared: useDexieLiveQuery called outside a Vue effect scope in the RestApi constructor leaks the localChanges subscription
## Problem

`useDexieLiveQuery` is called directly in the `RestApi` constructor:

```ts
// shared/src/api/RestApi.ts
constructor() {
    this.http = new HttpReq(config.apiUrl || "");

    const localChanges = useDexieLiveQuery(
        () => db.localChanges.toArray() as unknown as Promise<LocalChangeDto[]>,
        { initialValue: [] as unknown as LocalChangeDto[] },
    );
    syncLocalChanges(localChanges);
}
```

`getRest()` constructs this singleton from `luminary.ts` during init — outside any
component `setup()` or `effectScope`. `getCurrentScope()` is therefore `null`, and
the composable's own guard fires:

> [useDexieLiveQuery] called with no active Vue effect scope — the Dexie subscription
> can never be cleaned up and will leak forever.

## Impact

- The warning is logged on every app and CMS boot.
- `tryOnScopeDispose` is a no-op without a scope, so the `localChanges` subscription
  is never unsubscribed.
- `getRest({ reset: true })` compounds it: the replaced instance's subscription stays
  live, so each reset strands another one.

## Expected

The subscription should be owned by a disposable scope, matching the two call sites
that already do this correctly:

- `shared/src/fts/ftsLiveSync.ts` — `effectScope(true)` torn down with the composable
- `shared/src/util/useHasLocalChange/useHasLocalChange.ts` — detached scope, documented
  as a deliberate process-wide singleton

## Suggested fix

Wrap the live query in its own `effectScope(true)` held on the instance, expose a
`dispose()`, and call it from `getRest()` before replacing the singleton so resets
tear down the previous subscription.

## Notes

Found while running the Zitadel auth PoC locally against the CMS. Not specific to that
work — it affects every `app/` and `cms/` boot.

MrDirkelz:
Closes #1946

## Problem

`getRest()` constructs the `RestApi` singleton from `luminary.ts` during init — outside any component `setup()` or `effectScope`. `useDexieLiveQuery` was called bare in the constructor, so `getCurrentScope()` was `null` and the composable's own guard fired on every app and CMS boot:

> [useDexieLiveQuery] called with no active Vue effect scope — the Dexie subscription can never be cleaned up and will leak forever.

Without a scope `tryOnScopeDispose` is a no-op, so the `localChanges` subscription was never unsubscribed. `getRest({ reset: true })` compounded it: the replaced instance's subscription stayed live, stranding another one per reset.

## Change

Own the subscription in an `effectScope(true)` held on the instance, expose `dispose()`, and call it from `getRest()` before replacing the singleton.

This matches the two call sites that already got it right — `ftsLiveSync.ts` (scope torn down with the composable) and `useHasLocalChange.ts` (detached scope, documented as a deliberate process-wide singleton). Those two were audited and left alone.

## Testing

- `shared` type-checks clean
- 344 tests pass in `shared/src/api`

No behavioural change to request handling — only subscription ownership.


johan-bell (approved):
Looks good — scope creation, single subscription, and teardown on reset all check out, and it matches the ftsLiveSync/useHasLocalChange precedents. Two non-blocking nits: getRest() disposes the old instance before constructing the new one (a throwing constructor would leave a disposed singleton behind — only reachable in tests), and a spy-based test asserting dispose() fires on getRest({ reset: true }) would lock the fix in.

ivanslabbert (approved):

Issue #1868, opened 2026-08-01
Shared, API, APP, CMS: Resolve remaining Dependabot alerts (breaking major upgrades)


johan-bell:
Umbrella to clear the **remaining** Dependabot alerts across all four packages after the non-breaking fixes land in #1849.

Everything left needs a **breaking major upgrade**, so each must be migrated + verified against the relevant suite — this cannot be done as a single `npm audit fix --force` PR (it would install NestJS 11 + vitest 4 + vite 8 + eslint 10 + jest 30 at once and leave the build red).

### Residual alerts by required upgrade (post-#1849)

**api**
- **NestJS 10 → 11** — tracked separately in #1866 (production `@fastify/middie` critical + fastify/multer/fast-uri highs). Highest priority.
- **jest 27 → 30** (+ `ts-jest`) — cluster of test-tooling highs (`@jest/*`, `babel-jest`, `glob`, …).
- **sharp** high — needs image-processing verification.

**app / cms / shared** (dev/build tooling — the two criticals are `@vitest/coverage-v8` / `vitest`)
- **vitest 3 → 4** (`@vitest/coverage-v8`, `vite-node`) — critical.
- **vite 7 → 8** (`esbuild`, `vite`).
- **eslint 9 → 10** (`@eslint/eslintrc`, `@humanwhocodes/config-array`, `brace-expansion`, …).
- **@vue/eslint-config-typescript → 14** (the coupled `@typescript-eslint/*` 8.x line).
- **vue-tsc**, **@vue/test-utils**, **eslint-plugin-vue**, **@vitejs/plugin-vue**, **uuid**, **vite-plugin-dts** (shared), **doiuse** (app).

### Approach
Split into per-tool migration PRs, each verified independently:
1. NestJS 11 (#1866) — api integration suite (CouchDB/MinIO).
2. jest 27→30 (api).
3. vitest 3→4, vite 7→8, eslint 9→10, vue-tsc / @vue/test-utils / eslint-config-typescript (app + cms + shared) — verify type-check + lint + unit tests + build per package.

### Refs
- Plan: #1800 · Non-breaking half: #1849 · NestJS slice: #1866

Issue #1800, opened 2026-07-09
Security: Dependabot vulnerability audit + remediation plan (180 alerts)


johan-bell:
## Dependency vulnerability audit (Dependabot, all 4 packages)

Full sweep of the [Dependabot alerts](https://github.com/bccsa/luminary/security/dependabot) across `api/`, `app/`, `cms/`, `shared/`, with a remediation split by risk. A first PR landing the safe, non-breaking half is linked below; the remainder needs scoped major-version upgrades that can't be verified without the integration test infra.

### Snapshot

180 open Dependabot alerts at time of audit: **8 critical, 84 high, 81 medium, 7 low**. The count is inflated by the same advisory appearing across up to four lockfiles and by multiple CVEs on one package; the distinct-package picture is much smaller.

**Only one critical is production-facing.** Everything else critical is dev/build/test toolchain — real, but not shipped to users.

| Critical | Scope | Fix requires |
|---|---|---|
| `@fastify/middie` (auth bypass in child plugin scopes) — **api, runtime** | **production** | middie 8.x→9.x, i.e. **NestJS 10→11** (major) |
| `vitest` + `@vitest/coverage-v8` (UI server arbitrary file read/exec) — app, cms, shared | dev/test | **vitest 3→4** (major) |
| `handlebars` (JS injection via AST confusion) — api | dev (build) | semver-compatible |
| `shell-quote` — app, cms | dev (build) | semver-compatible |
| `form-data` (unsafe boundary RNG) — shared | dev | semver-compatible |

### What the linked PR fixes (safe, non-breaking)

Ran a non-`--force` `npm audit fix` in each of the four packages — **lockfile-only, no `package.json` version-range changes**, so no declared dependency was altered. Result:

| Package | Before | After | Cleared |
|---|---|---|---|
| api | 49 | 23 | 26 |
| app | 39 | 15 | 24 |
| cms | 39 | 15 | 24 |
| shared | 42 | 22 | 20 |

~94 alert-instances cleared. This covers the semver-compatible transitive bumps: `axios` → 1.16.x, `ws` → 8.21, `socket.io-parser` → 4.2.6, `form-data`, `brace-expansion`, `postcss`, `uuid`, `qs`, `glob`, `music-metadata`, `lodash`/`lodash-es` → 4.17.23, and the dev-side `handlebars`/`shell-quote` criticals, among others.

[...]

PR #1950, opened 2026-08-27
fix(api): migrate NestJS 10 → 11 to clear critical @fastify/middie auth-bypass

### LINKED ISSUE #1866: API: Migrate NestJS 10 → 11
Migrate the `api/` backend from NestJS 10 to NestJS 11.

This is the highest-priority security follow-up from #1800 and the breaking half deliberately left out of the non-breaking audit-fix PR #1849. It's the **only production critical** in the Dependabot sweep.

### Why
The NestJS 10 line pins a cluster of vulnerable Fastify-stack packages that can't be patched without the major bump:

- **`@fastify/middie`** — auth bypass in child plugin scopes (**production critical**). Fixed by middie 8.x→9.x, which requires NestJS 10→11.
- The associated highs pinned by the same line: **`@nestjs/platform-fastify`**, **`fastify`**, **`fast-uri`**, **`multer`**.

### Scope
- Bump `@nestjs/*` (core, common, platform-fastify, platform-express, schematics, cli, testing) to 11.x and reconcile the Fastify/middie/multer stack.
- Address any NestJS 11 breaking changes (Fastify v5 under the hood, middleware/adapter API changes, etc.).
- Verify against the `api` integration suite (CouchDB / MinIO) — this can't be validated on unit tests alone, which is why it was excluded from the non-breaking PR.

### Refs
- Parent plan: #1800
- Non-breaking half (already landed / in review): #1849

ChrisTouo:
## Why

`@nestjs/platform-fastify@10` pins `@fastify/middie@8.3.3`, which has a **critical** auth-bypass
in child plugin scopes ([GHSA-72c6-fx6q-fr5w](https://github.com/advisories/GHSA-72c6-fx6q-fr5w)).
The fix only exists in middie 9.x → needs Fastify 5 → needs NestJS 11. No way to patch it in place.

This API registers no Express-style middleware, and `platform-fastify@11` doesn't depend on middie
at all — so the package leaves the tree entirely:

```
$ npm ls @fastify/middie
└── (empty)     # was: @nestjs/platform-fastify@10.4.22 → @fastify/middie@8.3.3
```

## What changed

Dependencies only — **no source changes**. All `@nestjs/*` → 11.2.3, plus the majors Fastify 5
forces (`@fastify/multipart` 8→10, `@fastify/compress` 7→9) and two packages whose peer ranges
cap at Nest 10 (`@nestjs/jwt`, `nest-winston`). Fastify 4→5 transitively.

Nothing in the code needed touching: only `@Get`/`@Post` are exposed (both CORS-safelisted), the
exception filter only *reads* `reply.sent` rather than assigning it, and the `FastifyRequest`
augmentation still resolves. Express 5 arrives only via a devDependency used by two specs —
production is Fastify-only.

## Verification

Full suite against real CouchDB + MinIO, before and after:

| | Before | After |
|---|---|---|
| Tests | 817/817 | **817/817** (per-suite diff identical) |
| `npm audit` | 22 (1 critical) | **1** (`sharp`, tracked separately) |

Also clean: `tsc --noEmit`, `eslint`, `nest build`, and a real boot with all routes mapped and all
four Socket.io handlers registered — including the deprecated `joinSocketGroups` alias that
deployed clients rely on (ADR 0005).

## Follow-ups

`sharp` needs its own major (#1868). `socketio.ts` is at 35% coverage, so the auth-failure path
and the multipart recovery path need a manual pass. E2E only runs against deployed environments,
so it can't gate this pre-merge.

Refs #1866, #1800


ivanslabbert (approved):
Check why rxjs dependencies are added? We removed code dependent on rxjs (shared library), but it could be a downstream dependency in the api.

ChrisTouo:
> Check why rxjs dependencies are added? We removed code dependent on rxjs (shared library), but it could be a downstream dependency in the api.

Unlike shared/, rxjs can't be dropped from api/ — it's a required peer dep of @nestjs/core and @nestjs/common (^7.1.0), since Nest is built on Observables. But none of our own api/src code imports it directly, so it's purely a framework requirement.

PR #1951, opened 2026-08-27
refactor(cms): remove the auth bypass

### LINKED ISSUE #1949: CMS: Fix auth provider sync

MrDirkelz:
VITE_AUTH_BYPASS set only the Authorization header, never x-auth-provider-id and never the socket auth, so the API resolved those sessions as anonymous. It rendered a super-admin-shaped UI over a guest API session, which is why it could not be extended past one role.

The Playwright suite it was written for no longer exists in this package — no config, no e2e directory, no @playwright/test dependency, and the test:e2e scripts the README documented are absent. Persona sign-in in playwright-tests covers this properly now.

Two behaviours it suppressed are restored: conditionalAuthGuard actually guards, and registerAuthFailureHandler attaches its connect_error listener. Several call sites simplify, since the remaining isAuthPluginInstalled check was doing the real work.


app/src/auth.ts
```diff
@@ -69,7 +69,15 @@ function consumeForceReauthOnNextLogin(): boolean {
...
-    return /^https?:\/\//.test(domain) ? domain.replace(/\/$/, "") : `https://${domain}`;
+    const withScheme =
+        domain.startsWith("http://") || domain.startsWith("https://")
+            ? domain
+            : `https://${domain}`;
+    // A trailing slash survives into `${authority}/.well-known/openid-configuration`,
+    // which some providers 404 on.
+    let end = withScheme.length;
+    while (end > 0 && withScheme[end - 1] === "/") end -= 1;
+    return withScheme.slice(0, end);
```
ivanslabbert: Is there a simpler way to do this?

app/src/main.ts
```diff
@@ -84,50 +78,10 @@ async function Startup() {
...
-            clearAuthCache();
-            socket.setAuth("", null);
-            if (lastProvider) {
-                await loginWithProvider(lastProvider, { prompt: "login" });
-            } else {
-                openProviderModal();
-            }
-        },
-    );
+    registerAuthFailureHandler();
```
ivanslabbert: Do we need an `await` here?

cms/test-coverage.md
```diff
@@ -103,7 +103,7 @@ These files are excluded in `vitest.config.ts` and do not count toward coverage
 ## Files That Cannot Easily Reach 100%
 
 ### `src/auth.ts` (~50% coverage, up from 24%)
```
ivanslabbert: Measure test coverage and update figure

ivanslabbert (approved):

app/src/auth.ts
MrDirkelz: I checked the alternatives. The indexed scan deliberately removes any number of trailing slashes in linear time without reintroducing the CodeQL polynomial-regex finding. URL.origin is shorter but would silently strip issuer paths such as /realms/acme, while repeatedly slicing would be quadratic. I think the current implementation is the safest option.

cms/test-coverage.md
MrDirkelz: Validated with npx vitest run --coverage src/auth.spec.ts: src/auth.ts is now at 100% for statements, branches, functions, and lines across 56 tests. The summary is also stale (latest CI: 124 test files / 1,042 tests versus 102 / 674 in the document), so this needs a measured documentation update. I am leaving this thread open until that is done.

app/src/main.ts
MrDirkelz: No. registerAuthFailureHandler() is synchronous and returns void; it only attaches the connect_error listener. Awaiting it would not change startup ordering. The asynchronous setupAuth() call immediately below is already awaited.

PR #1953, opened 2026-08-27
APP: video won't play when no audio track matches the app language

johan-bell:
Follow-up to #1938. Two related fixes for videos whose audio tracks don't include the app language (e.g. English).

## 1. Video wouldn't start at all

On such content, clicking play did nothing — playback never started.

**Cause:** #1938 fixed the *disable-every-track* stall by keeping the current track when nothing matches. But if **no track was enabled** at that point (the case for this content), "keep the current track" keeps nothing enabled — and with no enabled audio rendition, HLS never starts.

**Fix:** always enable **exactly one** track. When no track matches the app language, fall back to a default — the track already enabled (the stream default), else the first — via a small tested helper `fallbackTrackIndex`.

## 2. Playback started, then stalled mid-video

With (1) in place the video starts, but stalls a few seconds in. The console shows VHS flagging **audio** segment downloading as excessive and thrashing through every video rendition (480→720→1080→360→240), plus a repeating "Player stalled, attempting to refresh buffer".

**Cause (app-side amplifier):** the `waiting` stall-recovery seeked `currentTime + 0.001` on *every* waiting event and re-applied the audio track. Seeking aborts in-flight HLS segment requests, which is exactly the condition VHS reports as "excessive segment downloading" — so the recovery fed the starvation it was trying to fix, turning a transient hiccup into a hard stall.

**Fix:** throttle the buffer-refresh seek to at most once per 6s, and drop the audio-track re-selection from the `waiting` handler (it's unrelated to buffering and re-touching the track list resets the audio loader).

> Note: the *root* cause of the stall on the `cameroon-test-video` asset is likely media-side — VHS keeps flagging **audio** (not video) segments, so the demuxed audio rendition itself is the bottleneck (slow/heavy segments or poor CDN caching). This PR stops the app from amplifying that, but a genuinely slow audio rendition should also be checked in the convert pipeline.

## Verification

Type-check clean, lint clean, 23 tests pass across `audioTrackLanguage.spec.ts` + `VideoPlayer.spec.ts`. Not device-verified — worth a check on the same staging content that wouldn't start/stalled.

### DIFF
--- a/app/src/components/content/VideoPlayer.vue
+++ b/app/src/components/content/VideoPlayer.vue
@@ -274,23 +274,26 @@ onMounted(async () => {
             }
         });
 
-        // Handle the "waiting" event, which occurs when the player is buffering
+        // Recover from a wedged buffer, but at most once every few seconds. Seeking aborts
+        // in-flight HLS segment requests, so nudging on every "waiting" event feeds the very
+        // buffer starvation it tries to fix — VHS then reports "excessive segment downloading"
+        // and thrashes through renditions. Re-selecting the audio track here is dropped too: it
+        // is unrelated to buffering and re-touching the track list resets the audio loader.
+        let lastStallNudge = 0;
         player.on("waiting", () => {
             const currentTime = player?.currentTime() || 0; // Get the current playback time
             setTimeout(() => {
-                // Check if the player is still stalled after 2 seconds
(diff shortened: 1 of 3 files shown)


ivanslabbert:
Already fixed in #1953

PR #1957, opened 2026-08-28
Graph button alignment and graph fullscreen toggle fix

### LINKED ISSUE #1948: CMS: Groups Overview Graph action buttons not aligned

brucefrankstad26-beep:


johan-bell (changes requested):
The toolbar alignment part is fine, but the LModal change regresses other dialogs:

- `LModal.vue` isFullscreen: `(isMobileScreen && stickToEdges) || (!isMobileScreen && stickToEdges)` reduces to just `stickToEdges` — the mobile check is dead code.
- Because of that, every desktop dialog passing `stickToEdges` (Edit Group, Create/Edit User, Redirect, S3 Bucket, Auto Group Mapping, Auth Provider) now renders 100vw × 100dvh with square corners instead of a centred dialog. Please give the graph its own `fullscreen` prop instead of repurposing `stickToEdges`.
- The backdrop only drops its `p-2` when `stickToEdges && isMobileScreen`, so a desktop stick-to-edges modal without `no-padding` overflows the viewport.
- `text-zinc-800` and `max-h-[100dvh]` added to the shared base classes affect every CMS modal — please keep those out of this fix.
- Nit: `isFullscreen` returns a class string but reads as a boolean; a `computed` would fit better.

brucefrankstad26-beep:
When I apply those changes suggested to the modal, then it takes things back to the bug we were trying to fix in the first place; but I did remove the condition from the function that checks for mobile screens only, and removed the text-zinc-800 and max-h-[100dvh], and changed the returned value as a computed value instead, those fixes keep it working properly, but the first two, take it back to the bug as. it was

johan-bell:
Check marging on desktop 
image

brucefrankstad26-beep:
Screenshot 2026-09-04 at 09 52 27


Margin fixed

johan-bell (approved):
LGTM

